PR #3, opened 2026-04-22
chore: fix all ESLint + tsc --noEmit errors

epeicher:
## Summary

CI's `js` lane was failing because trunk carried 294 ESLint errors and 5 TypeScript errors. This branch cleans both to zero.

- **ca6acc0** — ESLint: `eslint --fix` handled 284 mechanical issues (indent, curly, jsdoc alignment). The remaining 10 are hand-fixed: drop an unused `DesktopConfig` import, swap three `document.activeElement` calls for `this._el.ownerDocument.activeElement` (satisfies `@wordpress/no-global-active-element`), convert a ternary-as-statement to `if/else`, flatten a nested ternary, rename shadowed `registry` and type-parameter `body`, disable `no-alert` on the one intentional `window.confirm()` with an inline reason, and add the missing parent `@param options` JSDoc line.
- **f1c436e** — TypeScript: drop the unused `_unsubCommands` field (panel is page-lifetime, subscription dies with the page); rename unused `cmd` parameter to `_cmd`; replace a call to the nonexistent `this._clearResults()` with inline clear+hide; make `x` / `y` optional on `NativeWindowDef` (server-declared native windows never set them, and `createRegisterWindow` already nullish-coalesces with `?? 0` — the type was out of sync with runtime).

No runtime behavior changes — both commits are type-/style-only.

## Test plan

- [ ] `js` job on CI: ESLint, `tsc --noEmit`, and Vitest all green
- [ ] `npm run lint` locally: 0 errors
- [ ] `npx tsc --noEmit` locally: 0 errors
- [ ] Vitest still passes (~360 tests)

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -100,6 +100,19 @@ module.exports = {
 		// trust the reader knows multiplication binds tighter than
 		// addition.
 		'no-mixed-operators': 'off',
+		// The WP/ESLint indent rule was tuned for JSX — it has special
+		// handling for JSXElement / JSXExpressionContainer nodes so
+		// ternaries indented deeper inside `prop={ cond ? x : y }` read
+		// naturally. We use tagged-template DSLs (`html\`...\``) instead,
+		// and the rule has no equivalent awareness there: ternaries
+		// inside `${...}` get flattened against the enclosing block and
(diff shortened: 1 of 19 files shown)


AllTerrainDeveloper (approved):
LGTM! Works locally as expected too 👍

PR #2, opened 2026-04-22
ci: make `npm run test:php` actually work

epeicher:
## Summary

A fresh clone had no way to run PHP tests — `vendor/bin/phpunit` didn't exist, and CI had to curl the WP test installer every run. This wires up a Composer-based toolchain so local dev and CI share one binary and one setup flow.

- `composer.json` + `composer.lock` pin `phpunit/phpunit ^9.6` and `yoast/phpunit-polyfills ^2.0`
- `bin/install-wp-tests.sh` is the canonical wp-cli scaffold installer, pinned so CI no longer curls it
- `package.json`: `test:php` → `vendor/bin/phpunit …`; new `test:php:install` wraps the SVN installer with local defaults
- `tests/phpunit/bootstrap.php` loads the Composer autoload first (polyfills) and points the "missing test suite" error at `test:php:install`
- `.github/workflows/ci.yml` drops the `phpunit:9` tool pin, adds `composer install`, calls the committed installer
- `DEVELOPMENT.md` documents the one-time setup (svn, MySQL, `composer install`, `npm run test:php:install`)

## Test plan

- [ ] PHPUnit job on CI passes on both PHP 8.1 and 8.2 matrix legs
- [ ] Composer install step completes cleanly on CI
- [ ] `bin/install-wp-tests.sh` populates `/tmp/wordpress-tests-lib`
- [ ] Local run after `composer install` + `npm run test:php:install` + `npm run test:php` succeeds

PR #4, opened 2026-04-23
fix: resolve 3 PHPUnit failures + 1 error on trunk test code

epeicher:
Depends on #2 for the working `vendor/bin/phpunit` + WP test-suite install. Once the CI infra from #2 started actually running PHPUnit, four real issues became visible in trunk's code/tests. Each fix below is targeted and explains *why* rather than re-stating *what*.

### 1. `WpDesktopBuildDockItems::test_placement_distinguishes_core_from_plugin_menus` — ERROR

`Undefined array key "menu-index-php"`.

The test helper `make_menu_row` built synthetic hooknames as `'menu-' . sanitize_key( $slug )`. `sanitize_key` strips `.` entirely, so `index.php` → `indexphp`, producing dock id `menu-indexphp`. The assertion expects `menu-index-php` (the real WP hookname style that normalizes dots to dashes).

**Fix:** pre-replace `.` → `-` in the slug before passing it to `sanitize_key` inside the helper.

### 2+3. `Security::test_data_uri_rejected` and `WpDesktopBuildDockItems::test_icon_data_svg_rejected` — FAILURES

Both expect every `data:` URI (including `data:image/svg+xml;base64,…`) to fall back to `dashicons-admin-generic`. `wpdm_sanitize_dock_icon` carried a special-case branch that *accepted* base64-encoded SVG data URIs, directly contradicting its own `@since 0.11.0 Rejects data: URIs outright` docstring.

**Fix:** drop the branch so every `data:` URI falls through to the fallback, matching the documented Phase-11 intent.

### 4. `WpDesktopSession::test_save_session_accepts_equal_timestamp` — FAILURE

`Failed asserting that false is true.` on the second save.

`wpdm_sanitize_session` always stamped `$clean['updated'] = time()` before the save logic ran, overwriting whatever `updated` the client sent. That meant the stale-write guard compared `incoming` against a server-time `stored`, so equal-timestamp writes looked strictly-older and got rejected.

**Fix:** preserve the client's `updated` when valid, fall back to `time()` only when missing or invalid. `test_save_session_rejects_stale_write` and `test_save_session_accepts_missing_timestamp` continue to pass under the same logic.

## Test plan

- [ ] CI: `Tests: 331, Assertions: 698, Errors: 0, Failures: 0` on both PHP 8.1 and 8.2 matrix legs
- [ ] `js` lane: still green

### DIFF
--- a/includes/helpers.php
+++ b/includes/helpers.php
@@ -485,22 +485,6 @@ function wpdm_sanitize_dock_icon( $icon ) {
 		return preg_replace( '/[^a-z0-9_-]/', '', $icon );
 	}
 
-	// SVG data URI — the JS dock renders these as CSS background-image,
-	// which does NOT execute script in any currently shipping browser
-	// (Chrome/Firefox/Safari all treat SVG-in-CSS-background as an
-	// image-only context since ~2017). Strict validation on the base64
-	// payload prevents anything other than proper base64 from reaching
-	// the DOM; any other data:* scheme (javascript:, text/html, etc.)
-	// is rejected outright by the prefix check.
-	$svg_prefix = 'data:image/svg+xml;base64,';
-	if ( 0 === stripos( $icon, $svg_prefix ) ) {
-		$payload = substr( $icon, strlen( $svg_prefix ) );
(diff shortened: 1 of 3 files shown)


epeicher:
Closed in favour of #7

PR #7, opened 2026-04-23
ci: wire PHP tests through wp-env + fix 4 pre-existing failures

epeicher:
## Summary

- `npm run test:php` now actually runs. The toolchain moves to [`@wordpress/env`](https://github.com/WordPress/gutenberg/tree/trunk/packages/env) — the WP core test library, MariaDB, and PHP all live inside a container, so the only local prereq is Docker + `npm install`.
- PHPUnit stays pinned to `^9.6`. Tried bumping to 10; WP 6.9.4's test library still calls `PHPUnit\Util\Test::parseTestMethodAnnotations()` (removed in PHPUnit 10) and the polyfills don't bridge direct WP-core calls. Blocked upstream.
- Four pre-existing test failures fixed — all were latent on trunk and only surfaced once the harness was wired up.

## What's in here

**Commit 1 — `ci: wire PHP tests through wp-env`**
- `.wp-env.json` mounts the repo at `wp-content/plugins/wp-desktop-mode`
- `composer.json` with `phpunit/phpunit ^9.6` + `yoast/phpunit-polyfills ^2.0`
- `bootstrap.php` autoloads composer and defaults `WP_TESTS_DIR` to `/wordpress-phpunit`
- `phpunit.xml.dist` schema realigned to 9.5 (attributes were already 9.x)
- New scripts: `env:start`, `env:stop`, `env:destroy`, `test:php:install`, `test:php`
- Stops tracking the `.phpunit.result.cache` state file

**Commit 2 — `fix: resolve 4 pre-existing PHPUnit failures`**
- `includes/helpers.php` — remove the stale SVG data-URI accept branch in `wpdm_sanitize_dock_icon`. The docstring (`@since 0.11.0`) already declared every `data:` URI rejected; code and doc disagreed.  
  ⚠️ **Behavior change**: any plugin shipping an SVG data-URI dock icon now falls back to `dashicons-admin-generic`.
- `includes/session.php` — `wpdm_sanitize_session` now preserves the client's `updated` timestamp when valid, so two saves inside the same second no longer lose the second write to the stale-write guard.
- `tests/phpunit/tests/wpDesktopBuildDockItems.php` — normalize `.` to `-` in the synthetic hookname helper (`sanitize_key('index.php')` strips the dot entirely, producing `indexphp` instead of `index-php`).

## Test plan

- [ ] `npm install`
- [ ] `npm run env:start` (first run pulls WP + MariaDB images, ~2–5 min)
- [ ] `npm run test:php:install` (composer install inside the `tests-cli` container)
- [ ] `npm run test:php` → expect `OK (339 tests, 723 assertions)`
- [ ] Verify the SVG data-URI behavior change against any plugin that relies on it (we don't ship one in-tree)

## Notes

- CI workflow (`.github/workflows/ci.yml`) is **not** updated in this PR — its `php` lane will still fail until switched to wp-env. Worth a follow-up once this lands.
- Supersedes the abandoned `fix/php-tests` branch (PR #2) and `fix/php-tests-assertions` branch (PR #4). Both can be closed.


includes/helpers.php
```diff
@@ -485,22 +485,6 @@ function wpdm_sanitize_dock_icon( $icon ) {
...
-	// is rejected outright by the prefix check.
-	$svg_prefix = 'data:image/svg+xml;base64,';
-	if ( 0 === stripos( $icon, $svg_prefix ) ) {
-		$payload = substr( $icon, strlen( $svg_prefix ) );
-		if ( preg_match( '/^[A-Za-z0-9+\/=]+$/', $payload ) ) {
-			return $icon;
-		}
-		return $fallback;
-	}
-
```
epeicher, line 503: Flagged as stale code

PR #20, opened 2026-04-27
Fix/Real time dock updates

AllTerrainDeveloper:
# Live menu refresh: fix three blockers, add the missing tests

Activating or deactivating a plugin from inside the chromeless
`plugins.php` window did not update the dock, the bottom centered
taskbar, the wallpaper desktop icons, or any plugin's branded menu
icon — users had to F5 to see the change. Three bugs all sat on the
same path; fixing them in one PR because the test scaffolding is shared.

### 1. JS syntax error killed the entire chromeless bridge

`includes/render.php` emits the chromeless bridge as one inline
`<script>` block (single heredoc, lines 513–1896). At line 1216 the
opening of a `__wpdLog( '...' ,` call had been removed but its
trailing arguments + closing paren were left orphaned:

```js
__wpdLastRawCommands = merged;
    merged.length,
    loadersList.length,
    Array.isArray( resultsBucket.statics ) ? resultsBucket.statics.length : 0
);
__wpdSchedulePost();
```

JS parses each `<script>` tag all-or-nothing, so a syntax error
anywhere in the block stops the whole bridge from running — including
the `wp-desktop-plugins-changed` `postMessage` at line 736 that drives
live refresh. That's why nothing updated live: the iframe never told
the parent shell anything had changed.

**Fix:** restored the original `__wpdLog( '[wpd-cmd:iframe] react-harvester: …', …)`
call shape so the bridge parses again. `__wpdLog` is a no-op when
`WP_DEBUG !== true`, so the call is purely a debug breadcrumb — but
it has to be syntactically complete.

### 2. `applyPayload` didn't re-render `desktopIcons` on live refresh

PHP's `wpdm_build_menu_payload()` always emits `desktopIcons`, and
the boot path in `src/desktop.ts` rendered them from
`config.desktopIcons`. But the live-refresh applier inside
`bindMenuRefresh()` was missing that key entirely — so a
plugin-registered wallpaper icon (e.g. Jorvy) only appeared after F5,
and likewise stayed stuck on the wallpaper after deactivation.
`serverTitleBarButtonScripts` had a fragile cast-around-the-type
shape of the same problem.

**Fix:** extracted `applyPayload` into its own pure factory at
`src/menu-refresh-apply.ts` (`createApplyPayload`). The new
`MenuRefreshPayload` interface lists every key the bridge can carry,
so adding a key without wiring it now fails type-check loudly. Wired
both `desktopIcons` and `serverTitleBarButtonScripts` properly. A
shared `renderIcons` callback drives both the boot path and the
live-refresh path through the same renderer (avoids drift).

### 3. SVG menu icons collapsed to the gear fallback

`wpdm_sanitize_dock_icon()` (`includes/helpers.php`) had been
rejecting **all** `data:` URIs since 0.11.0, replacing them with
`dashicons-admin-generic`. Yoast, WooCommerce, Jetpack, Elementor —
basically every modern WP plugin — registers `$menu[$i][6]` as
`data:image/svg+xml;base64,…`, so all of them landed on the gear.
The TS dock already had a renderer for SVG data URIs
(`src/dock.ts:451-457`, CSS `background-image`), but it never fired
because the value was nuked upstream.

The original 0.11.0 rationale — "SVG can carry inline scripts that
execute as a CSS background" — was true ~10 years ago but isn't now:
modern browsers sandbox scripts inside SVGs loaded via
`background-image` exactly like they sandbox `` SVGs.

**Fix:** allow two strict shapes through the sanitizer, anchored
whole-string so smuggled quotes / whitespace / newlines bounce:

```
^data:image/svg\+xml;base64,[A-Za-z0-9+/=]+$
^data:image/svg\+xml,[A-Za-z0-9._~!$&'()*+,;=:@/?%-]+$
```

Other `data:` schemes (`data:text/html`, `data:application/javascript`,
…) and the existing `javascript:` / `vbscript:` / `file:` rejections
are untouched.

## Files changed

- `includes/render.php` — restored the orphaned `__wpdLog` call.
- `includes/helpers.php` — `wpdm_sanitize_dock_icon` now passes
  well-formed `data:image/svg+xml` URIs through; updated docblock with
  rationale.
- `src/menu-refresh-apply.ts` (new) — `createApplyPayload` factory
  with the full payload-key contract typed.
- `src/desktop.ts` — uses the new factory; passes a `renderIcons`
  callback so boot and live-refresh share the same renderer.

## Tests

- `tests/vitest/menu-refresh-apply.test.ts` (new, 12 tests) —
  unit + real-`Dock` end-to-end for the live-refresh applier. Two
  regression-guards specifically pin `desktopIcons`
  activation/deactivation; four end-to-end tests pin a Yoast-style
  menu plugin appearing on / disappearing from the bottom taskbar.
- `tests/vitest/native-windows-sync.test.ts` (new, 10 tests) —
  the native-window (calc / home / code) live add/remove contract
  with real `Dock` instances: activation, deactivation, last-tile,
  reactivation, dock-vs-taskbar placement, idempotency, the
  hidden-rail-then-tile-arrives race the live-refresh path can hit.
- `tests/phpunit/tests/security.php` — split the old "all data URIs
  rejected" test into `test_svg_data_uri_allowed` (well-formed shapes
  pass through, including mixed-case scheme) and
  `test_data_uri_rejected_when_not_svg_or_malformed` (`text/html`,
  `application/javascript`, `;utf8,…`, smuggled quote, embedded
  newline — all fall back).
- `tests/phpunit/tests/wpDesktopBuildDockItems.php` — flipped
  `test_icon_data_svg_rejected` → `test_icon_well_formed_svg_data_uri_passes_through`.
- `tests/phpunit/tests/icons.php` — split the desktop-icon test into
  "malformed falls back" + a new "well-formed SVG passes through" so
  `wp_register_desktop_icon()` mirrors the dock policy.

JS suite: **464/464 passing** (22 new). Lint + `tsc --noEmit` clean.
PHP syntax clean. PHPUnit run locally outside this branch.

## Test plan

- [ ] Reload the WP admin tab (no build needed for the PHP fixes).
- [ ] Open the chromeless `plugins.php` window.
- [ ] Activate a plugin with a top-level menu (Yoast, Woo, …) →
      tile appears on the bottom centered taskbar with the plugin's
      branded SVG, not the gear.
- [ ] Deactivate the same plugin → tile disappears live.
- [ ] Activate / deactivate a native-window plugin (Calculator, Home
      Assistant, Code editor) → its icon adds / removes from the
      bottom bar live.
- [ ] After `npm run build`: activate / deactivate a plugin that
      registers a `wp_register_desktop_icon()` → wallpaper icon
      appears / disappears live.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22landingPage%22%3A%22%2Fwp-admin%2F%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fdesktop-mode%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-20-54a93a875f26e83ab3c214d934bb3246e3d26c7a.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


epeicher (approved):
LGTM!

PR #31, opened 2026-04-27
feat(dock): unify dock + taskbar into one rail with user-configurable placement

shaunandrews:
## Summary

The previous architecture split the admin menu across two rails — a left-edge dock for core WordPress menus and a bottom-edge pill taskbar for plugin-contributed top-level menus. The split was an implementation artifact that leaked into the UX: it hid the taskbar entirely on installs without plugin menus, and trained users to think "core vs. plugin" instead of "apps."

Now there is **one dock**. Default placement is **bottom** (macOS-style floating pill); users can pick **left** or **right** in **OS Settings → Appearance → Dock placement**. Core admin menus are ordered ahead of plugin-contributed menus and separated by a hairline group divider so "default apps" still read as a distinct cluster from "installed apps." Every existing affordance (tooltips, instance rail, multi-window chip, notification badge, OS Settings system tile) flips correctly across all three placements.

## Notable changes

- PHP payload collapses to a single `dockItems` array. The `wp_desktop_dock_placement` filter narrows from `'dock' | 'taskbar' | 'hidden'` to `'dock' | 'hidden'`. Each item carries an `isCore` flag the JS renderer uses to insert the group separator at the core→plugin transition.
- The `Dock` class gains a `'right'` orientation and a `setOrientation()` method so a placement change in OS Settings flips the tooltip anchor without re-instantiating.
- CSS keys placement off the shell root's `data-wp-desktop-dock-placement` attribute. The previous `--horizontal` / `--vertical` modifier classes are removed; tests are updated to verify the new tooltip-anchor contract instead.
- The PHP icon sanitizer now accepts `data:image/svg+xml;base64,…` URIs so plugins like Jetpack that ship inline SVG art render with their real icon. SVG-as-CSS-background is inert per spec, so this doesn't open a script-execution surface.
- The wallpaper-surfaces collector reads the placement attribute and emits a placement-aware dock edge (face: top / left / right) so collision-aware wallpapers respond to the dock wherever it lives.

## Breaking changes (pre-1.0)

- \`wp_desktop_dock_placement\` filter values: \`'taskbar'\` removed.
- \`wpDesktopConfig.taskbarItems\` removed; merged into \`dockItems\`.
- \`#wp-desktop-taskbar\` DOM element removed.
- \`DockOrientation\` type narrows to \`'left' | 'right' | 'bottom'\`.
- \`WallpaperSurface.kind\`: \`'taskbar'\` removed.

[...]

PR #39, opened 2026-04-29
fix(dock): improve icon visibility and fix separator height

darksns:
- Force monochrome white palette on plugin SVG icons that ship with hardcoded fill attributes, so they match the dashicons look
- Add hover opacity transition on dock icons (0.7 → 1)
- Fix horizontal separator height: was 60% (collapsing against the taskbar pill), now fixed 28px to match the dock tile glyph area
- Bump dock border alpha from 0.08 to 0.2 for better visibility


AllTerrainDeveloper:
Thank you for your first contribution!
| BEFORE  | AFTER |
| ------------- | ------------- |
| image | image |

There seems to be kind of a regression in the taskbar, a new "separator" appeared. Would you mind to refine it and we ship it!

Again, thank you!

darksns:
Hi, I see!
I'll work on it so it looks good in both situations. In the screenshot, I imagine it's a clean WordPress installation. In my installation, where there are also third-party plugin icons to the left of the separator, the missing separator was noticeable and there was a visible empty space.
I'll be back shortly with an update.
Thank you
Il 30 apr 2026, 19:38 +0200, Daniel López Sánchez (prismiwi2015) ***@***.***> ha scritto:

darksns:
Hi! Thanks for the review and the screenshots — they made the regression easy to spot.

Pushed cfc786e which hides the separator when it's the first child of the dock. The TS code creates the system/pinned separator before its first cluster item, so when the dock has no menu items above (clean WP install) the separator becomes the first child and now display: none kicks in.

Verified both scenarios on Chrome, Firefox and Safari:
- Clean install (no plugin menus in taskbar): separator hidden, no stray hairline
- Plugin-rich install: separator visible between menu items and system cluster, as before
CSS-only fix, no TS changes needed.

BEFORE
CleanShot 2026-04-30 at 23 09 23@2x

AFTER
CleanShot 2026-04-30 at 23 10 00@2x

darksns:
Pushed `5f69cb4`: re-applies the `height: 28px` fix on the new `data-wp-desktop-dock-placement="bottom"` selector, since the old `--horizontal` modifier was removed by PR #31. Same intent as before, just on the unified dock selector.

Also rebased onto current `trunk` to clear the conflict — the final diff vs upstream is just the SVG icon normalization (`filter: brightness(0) invert(1)` on `.wp-desktop-dock__item-svg`), the `:first-child` hide-when-empty rule, and this 28px separator height. Tested on Chrome, Firefox and Safari, both with and without third-party top-level admin menus.

Ready for another look!

AllTerrainDeveloper (approved):
Looks great now!

PR #52, opened 2026-04-30
feat(window): add Reload button to title-bar controls

darksns:
## What

Adds a **Reload button** to the title-bar of iframe-backed windows. Mirrors the Safari / Chrome refresh affordance users already expect — re-fetches the active iframe without leaving the desktop shell.

## Why

Right now, to reload an admin page open in a window, the user has to detach it to a classic tab (or close + reopen). The reload button closes that gap, especially valuable for editor workflows where a long-running save fails and the user wants to retry without losing their window layout.

## How

- **`src/window-chrome/controls/built-ins.ts`** — registers `core/reload` next to the other built-ins, between detach (order 40) and close (order 50). Filtered for native windows via `match: ( win ) => ! win.config.native`.

- **`src/window/index.ts`** — new `public reload()` method. Resolves target surface (primary iframe or external sub-tab) and its URL **before** any observable side effect, so a missing iframe returns silently instead of leaving the loading overlay stuck. Same-origin iframes use `location.reload()`; cross-origin external tabs fall back to `src` re-assignment. Integrates with the loading API from #50 by calling `markContentLoading()` — the iframe's `wp-desktop-ready` postMessage clears the spinner automatically once the new page hydrates.

- **`Window._spinReloadButton()`** — private helper that drives the one-shot click-feedback rotation. Mirrors the `shake()` pattern (remove class, flush reflow, add class, listen for `animationend`). Keeps DOM manipulation inside `Window` so `built-ins.ts` stays a pure registration list of one-line `onClick` handlers.

- **`src/ui/components/wpd-window-button/wpd-window-button.ts`** — `reload` icon entry in the inline-SVG map. Filled style chosen deliberately for stronger visual weight as a primary action; a wrapper `<g transform>` re-targets a 512×512 source path into the shared 12×12 viewBox.

- **`src/hooks.ts`** — `HOOKS.WINDOW_RELOADED = 'wp-desktop.window.reloaded'`, fired with `{ windowId, url }` so plugins can invalidate caches, force saves, or sync companion surfaces. Native windows do not fire it.

- **`assets/css/window-chrome.css`** — single 360° rotation on click with a steeper-than-default ease-out curve (0.6s, custom cubic-bezier explained inline) so the gesture feels acknowledged but visually distinct from a continuous loading spinner. While the body is in the `--loading` state from `markContentLoading()`, the button is dimmed and `pointer-events: none` — `:has()`-driven, no JS state mirroring. Reduced-motion media query opts out of the rotation but preserves the dim state for accessibility.

## Diff size

```
src/hooks.ts                                                |  +11
src/ui/components/wpd-window-button/wpd-window-button.ts    |  +22 -7
src/window-chrome/controls/built-ins.ts                     |  +18 -3
src/window/index.ts                                         | +107
assets/css/window-chrome.css                                |  +59
─────────────────────────────────────────────────────────────────────
                                                              +210 lines
```

## Edge cases handled

- **Native windows**: `match` predicate excludes them; `Window.reload()` itself also has a `config.native` early-return as defense in depth.
- **Missing iframe / external tab entry**: target resolution happens before `markContentLoading()` to avoid stuck spinners.
- **Double-click during in-flight reload**: body's `wp-desktop-window__body--loading` class is checked at entry; `pointer-events: none` on the button via CSS provides a second line of defense.
- **Cross-origin external tabs**: `try`/`catch` around `contentWindow.location.reload()` falls back to `src` re-assignment.
- **Custom chrome layers** that don't render the standard reload button: `_spinReloadButton()` silently no-ops if the button isn't found.
- **Reduced motion preference**: rotation animation suppressed; dim/disable state preserved.

## Public hook example

```ts
import { addAction, HOOKS } from '@wp-desktop-mode/hooks';

addAction( HOOKS.WINDOW_RELOADED, 'my-cache-plugin', ( payload ) => {
    myCache.invalidate( payload.url );
} );
```

## Testing

- [x] Chrome, Firefox, Safari
- [x] Same-origin admin pages (post editor, settings, etc.)
- [x] Cross-origin external sub-tabs (where applicable)
- [x] Native windows excluded as expected
- [x] Light + dark admin schemes
- [x] `prefers-reduced-motion: reduce` honoured

## Notes on style consistency

- The inline reload icon is filled (`fill: currentColor`) while the other built-in icons are stroke-only. This is an **intentional design choice** — the reload action has a primary-action visual weight that the stroke style undersells. The `<g transform>` keeps the path data in its 512×512 authoring viewBox for easy round-trips with vector tools. Happy to swap to a stroke variant if you prefer the homogeneous look.
- The cubic-bezier value `( 0.05, 0.7, 0.1, 1 )` is novel in this codebase (other transitions use `ease-out` or `cubic-bezier( 0.2, 0, 0.2, 1 )`). The choice is documented inline; happy to align with an existing token if you'd rather.

### DIFF
--- a/assets/css/window-chrome.css
+++ b/assets/css/window-chrome.css
@@ -627,3 +627,62 @@ wpd-tab-chip + wpd-tab-chip {
 	height: 18px;
 	display: block;
 }
+
+/* ---------------------------------------------------------------
+ * Reload button feedback.
+ *
+ * Click feedback is decoupled from the loading state:
+ *
+ *   1. On click, the icon performs a single 360° rotation with a
+ *      fast-start / slow-end ease curve — confirms the user's
+ *      gesture independently of how long the page takes to load.
+ *      The `--spinning` class is added by the click handler and
+ *      removed on `animationend` so a subsequent click restarts
+ *      the animation cleanly.
+ *
+ *   2. While the window body is in the `--loading` state (set by
+ *      `markContentLoading()` and cleared on the iframe's
+ *      `wp-desktop-ready` postMessage), the button is dimmed and
+ *      non-interactive so a second click can't desync the
+ *      chromeless bridge's loading handshake.
+ *
+ * `:has()` is used instead of mirroring the `--loading` modifier
+ * onto the window root because upstream's loading API only
+ * decorates the body element. Browsers without :has() (very old)
+ * fall back to a static dimmed button — feature, not bug.
+ * --------------------------------------------------------------- */
+
+.wp-desktop-window:has( .wp-desktop-window__body--loading )
+	.wp-desktop-window__btn--reload {
+	pointer-events: none;
+	opacity: 0.55;
+}
+
+/*
+ * Click feedback animation. The custom easing — `cubic-bezier( 0.05,
+ * 0.7, 0.1, 1 )` — is a steeper-than-default ease-out: ~70% of the
+ * rotation lands in the first 30% of the duration, then it decelerates.
+ * Reads as "snap, then settle" rather than the linear glide of a
+ * loading-spinner, reinforcing that this is a one-shot acknowledgement
+ * of the gesture rather than progress through a long task. The `0.6s`
+ * duration is short enough to never block the user but long enough
+ * that the deceleration is legible.
+ */
+.wp-desktop-window__btn--reload.wp-desktop-window__btn--spinning {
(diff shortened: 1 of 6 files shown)


AllTerrainDeveloper (approved):
Lovely! Looks great!

https://github.com/user-attachments/assets/b7064785-b510-44d3-9a88-4e35259ca5a5

PR #71, opened 2026-05-03
feat(code-editor): extract to standalone desktop-mode-code-editor extension

epeicher:
## Summary

Moves the Code editor out of the desktop-mode plugin into a new sibling extension (`extensions/desktop-mode-code-editor/`) so it can be installed and activated independently of desktop-mode — same pattern as the existing `desktop-mode-cron-manager` and `desktop-mode-phpmyadmin` extensions, but the first true *extraction* (those two were green-field).

Stacked on `add/marketplace-extensions`.

## What moves

- **PHP** — `includes/code-editor/{bootstrap,window,filesystem,php-indexer,php-workspace-indexer,rest}.php` → `extensions/desktop-mode-code-editor/includes/`. All renamed: `wpdc_*` → `desktop_mode_code_editor_*`, `WPDC_*` → `DESKTOP_MODE_CODE_EDITOR_*`, textdomain `desktop-mode-code-editor`, REST namespace `desktop-mode-code-editor/v1`, hook prefix `desktop_mode_code_editor_*`.
- **TS sources** — `src/code-editor/` → `extensions/desktop-mode-code-editor/src/`. Self-contained (no shell imports), so they build standalone.
- **Built assets** — `assets/{js,css}/code-editor*` → `extensions/desktop-mode-code-editor/assets/`.
- **Monaco vendor** (~14 MB) — `assets/vendor/monaco-editor/` → `extensions/desktop-mode-code-editor/assets/vendor/monaco-editor/` (gitignored, fetched by `bin/fetch-monaco.sh` during packaging).

## What stays the same (public API)

`wpdc-editor` window id, `wpDesktopCodeEditorConfig` global, `data-wpdc-editor-*` selectors, and `wpdc-editor*` CSS classes — they're public per `docs/examples/code-editor-open.md`, and renaming would break consumers that deep-link via `wp.desktop.openWindow('wpdc-editor')`.

## Why the config delivery changed

The in-tree code shipped its config via `wp_localize_script( 'wp-desktop-code-editor', 'wpDesktopCodeEditorConfig', … )`. That works on the eager-load path and silently fails on the lazy-load path the native-window sync takes (`loadVendorScript()` bypasses `wp_print_scripts`). Switched to the admin-ajax-served-bundle pattern used by `desktop-mode-cron-manager`: the bundle URL is `admin-ajax.php?action=desktop_mode_code_editor_bundle`, the response body sets `window.wpDesktopCodeEditorConfig = { … }` and then streams the prebuilt JS — no hook lifecycle to depend on, no inline-script tag to lose.

## Per-extension build

`extensions/desktop-mode-code-editor/{package.json, vite.config.js, tsconfig.json}` plus `bin/fetch-monaco.sh`. `npm install && npm run build` produces `assets/js/code-editor.{js,min.js}` and copies Monaco into `assets/vendor/`. `bin/fetch-monaco.sh` is auto-discovered by the existing `bin/package-extensions.sh` (it already runs `bin/fetch-*.sh` per extension before staging), so packaging "just works" on a fresh clone.

**Open question**: should `bin/package-extensions.sh` also auto-build TS bundles? Today it only splices `assets/vendor/*` back in after the `git ls-files` archive; build outputs (`assets/js/*.js`) are committed, matching the cron-manager convention. Happy to extend the splice in a follow-up if you want the bundle built on package.

## Removed from desktop-mode

- `includes/code-editor/`, `src/code-editor/`, `assets/{js,css}/code-editor*`, `assets/vendor/monaco-editor/`
- `require_once` in `desktop-mode.php`, registrations in `includes/assets.php`
- `code-editor` Vite target, `build:code-editor` + `vendor:monaco` npm scripts, `monaco-editor` + `@monaco-editor/loader` deps
- obsolete `/assets/js/code-editor*` lines in root `.gitignore`
- `tests/phpunit/tests/wpdc*.php` — referenced the renamed functions; re-establishing PHPUnit coverage inside the extension is a deliberate follow-up (neither cron-manager nor phpmyadmin ship tests today, so wiring up a phpunit setup inside an extension would be a new pattern)

## Docs updated

- `extensions.json` — new entry
- `docs/javascript-reference.md` — note next to `wpdc-editor` deep-link example that the extension must be active
- `docs/examples/code-editor-open.md` — header callout + renamed function reference

## Test plan

- [x] `npm run build` at root passes (no `code-editor` target)
- [x] `npm run lint` passes
- [x] `tsc --noEmit` at root passes
- [x] `npm run test:js` passes (717/717)
- [x] `npm install && npm run typecheck && npm run build` inside the extension passes
- [x] `bin/package-extensions.sh /tmp/out` produces a 6.4 MB zip with the correct vendor layout
- [ ] Manual smoke: install the produced zip on a fresh dev site, activate it alongside desktop-mode, open the Code window, edit + save a file, confirm the workspace symbol index lights up in completion
- [ ] Manual smoke: deactivate the extension and confirm `wp.desktop.openWindow('wpdc-editor')` returns falsy and the deep-link example is a no-op (graceful degradation)

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22landingPage%22%3A%22%2Fwp-admin%2F%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fdesktop-mode%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-71-1c211b7ecd96f655c11969bab989f8b94fe9bcee.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
Bye Code :(

PR #74, opened 2026-05-04
chore(sync): mirror plugin tree into a sibling wordpress-develop checkout

epeicher:
## Summary
- Adds `bin/sync-to-wp-develop.sh` + `bin/sync-to-wp-develop.includes` to mirror the working tree into `~/github/wordpress-develop/src/wp-content/plugins/desktop-mode` (override with `WPDM_SYNC_DEST`).
- One-shot mode: `bin/sync-to-wp-develop.sh` runs `npm run build`, builds each extension that has `node_modules/`, then rsyncs the include set.
- Watch mode: `bin/sync-to-wp-develop.sh --watch` adds an fswatch loop over the source inputs (TS under `src/`, CSS, PHP, per-extension `src/`, etc.) — built bundles under `assets/js/` and `assets/vendor/` are rsynced but **not** watched, so the build can't loop on its own output.

## Why this is on top of `add/code-editor-extension`
The includes manifest references `extensions.json`, which only exists on this branch (added by `8f162e5 feat(marketplace): add Extensions tab + release pipeline`). Targeting `trunk` would land a broken manifest.

## Notable behaviours
- rsync of `extensions/` excludes `node_modules/`, `src/`, `package.json`, `package-lock.json`, `tsconfig.json`, `vite.config.js`, `.gitignore` — each extension owns its own build pipeline; only runtime artefacts ship to wordpress-develop.
- fswatch excludes `/node_modules/` and `/extensions/<name>/assets/(js|vendor)/` so dependency churn and per-extension build output never trigger a rebuild.
- Per-extension `npm run build` is dispatched from `build_extensions()` because the parent `npm run build` doesn't recurse into `extensions/`. Failures are tolerated so one broken extension can't take down the watcher.
- The watch loop wraps `build_and_sync` in `|| echo …` so a single failed build doesn't kill the long-running watcher.

## Test plan
- [ ] `bin/sync-to-wp-develop.sh` populates the destination with the include set and nothing else.
- [ ] Editing a TS file under `src/` triggers a rebuild + rsync in `--watch` mode.
- [ ] Editing a TS file under `extensions/desktop-mode-code-editor/src/` triggers the extension's own `npm run build` and ships the new `extensions/desktop-mode-code-editor/assets/js/code-editor*.js` to the destination.
- [ ] Running `npm install` inside an extension does not trigger an infinite rebuild loop (node_modules excluded from fswatch).
- [ ] Destination's `extensions/desktop-mode-code-editor/` does not contain `node_modules/`, `src/`, `vite.config.js`, `tsconfig.json`, `package.json`, or `package-lock.json` after a sync.


epeicher:
The sync scripts will be added as part of a fresh PR, closing this as not required.

PR #102, opened 2026-05-06
Render admin-bar discovery toggle for fresh admins

### LINKED ISSUE #98: First-time admins have no on-screen path to discover Desktop Mode
## Summary

After installing and activating the plugin, a fresh admin sees nothing in the WordPress UI: no admin notice, no dashboard widget, no admin-bar entry. The only entry point is the `/desktop-mode/` portal URL, which the user has to either already know or read in the plugin docs.

The admin-bar entry that *would* be the discovery affordance is gated on the very thing it's supposed to enable — classic chicken-and-egg.

## Evidence

- No `admin_notices`, `wp_dashboard_setup`, or `wp_add_dashboard_widget` callbacks anywhere in the codebase.
- `add_action( 'admin_bar_menu', 'desktop_mode_admin_bar_toggle', 190 )` registers the toggle at `includes/admin-bar.php:279`.
- `desktop_mode_admin_bar_toggle()` early-returns at line 38 unless `desktop_mode_is_enabled()` is true.
- `desktop_mode_is_enabled()` (`includes/helpers.php:17-23`) reads user meta `desktop_mode_mode`, which is unset by default → admin-bar entry is hidden.
- The only entry point is the `/desktop-mode/` portal — `includes/portal.php:25, 96, 108-119`.

## Impact

First-time UX: a fresh admin enables the plugin and sees no change to their admin. Adoption suffers and the plugin appears broken. Confirmed in both automated test runs.

## Suggested fix

Pick one (or both):

1. **Render the admin-bar entry unconditionally** for users who hold the relevant capability. When desktop mode is off, show "Switch to Desktop Mode" (the label is already in place at `includes/admin-bar.php:39-41`). This is the lowest-cost fix.
2. **One-time admin notice on activation**, dismissible, pointing to `/desktop-mode/`. Use the plugin's activation hook to set a transient and `admin_notices` to render.

A dashboard widget is also an option but probably overkill given (1).

## Severity

Medium. UX/discovery only — feature still works for users who know the URL — but it materially affects first impressions and adoption.

epeicher:
## Summary

Fresh admins had no on-screen discovery path for Desktop Mode after activating the plugin. Investigation showed the admin-bar toggle was actually rendering for all logged-in users, but with no capability gate, so subscribers and roles denied by `desktop_mode_mode_enabled` saw a non-functional toggle.

This PR adds the right capability gate and an active-user carve-out so policy flips can't trap users in the shell.

### Capability gate

`current_user_can( 'read' ) && apply_filters( 'desktop_mode_mode_enabled', true, $user_id )` — exactly mirroring the AJAX `save-desktop-mode` endpoint at `includes/ajax.php:23,37` and the `/desktop-mode/` portal at `includes/portal.php:74`.

**Carve-out:** a user who already has Desktop Mode active sees the toggle even if the filter would deny them, so a mid-session policy flip can't trap them in the shell.

### Files

- `includes/admin-bar.php` — gate added to both `desktop_mode_admin_bar_toggle()` and `desktop_mode_enqueue_toggle_assets()` (so we don't ship CSS/JS to a user who has no node).
- `tests/phpunit/tests/adminBarDesktopToggle.php` — 4 new tests:
  - `test_toggle_renders_for_fresh_admin_without_meta` (pins the discovery affordance for #98)
  - `test_toggle_hidden_when_mode_enabled_filter_returns_false`
  - `test_toggle_renders_for_active_user_even_when_filter_denies` (the carve-out)
  - `test_toggle_hidden_for_user_without_read_cap`
  - Plus `remove_all_filters( 'desktop_mode_mode_enabled' )` in `tear_down`.
- `docs/hooks-reference.md` — `desktop_mode_mode_enabled` entry now notes the admin-bar render-side gate and active-user carve-out.
- `docs/examples/gate-by-role.md` — replaced an inaccurate claim (that the filter affects `desktop_mode_is_enabled()` directly) with the actual effects (AJAX refusal + admin-bar hide + active-user carve-out).

## Verified

- `npm run build` clean (5 Vite targets)
- `npm run lint` clean
- `tsc --noEmit` clean
- `npm run test:js`: 803/803 pass
- `php -l` on edited PHP files: no syntax errors

## Open question (separate follow-up)

The doc previously claimed `desktop_mode_is_enabled()` returns false when the filter returns false. It doesn't, the helper only reads user meta. I corrected the doc to match code; if the doc was the intended contract, the helper should consult the filter, but that's a behavior change affecting every render-time gate. Filed as a follow-up.

Closes #98

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22landingPage%22%3A%22%2Fwp-admin%2F%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fdesktop-mode%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-102-6713a0f18042db913a77e9149f1e4e72dd009de6.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/docs/examples/gate-by-role.md
+++ b/docs/examples/gate-by-role.md
@@ -27,7 +27,9 @@ add_filter( 'desktop_mode_mode_enabled', function ( $enabled, $user_id ) {
 Returning `false` has two effects:
 
 1. The **AJAX save** endpoint refuses to flip the user meta to `'1'` (it returns `desktop_mode_disabled`).
-2. Any existing `'1'` value is ignored — `desktop_mode_is_enabled()` returns `false` for this user.
+2. The **admin-bar "Switch to Desktop Mode"** discovery toggle is hidden for users whose desktop mode is currently off — they don't see a button that would do nothing.
+
+Users who *already* have desktop mode enabled (`desktop_mode_mode` user-meta is `'1'`) keep seeing the toggle so they can always switch back to classic admin, even if the filter flips to `false` mid-session.
 
 ## Disable the portal auto-enable too
 
--- a/docs/hooks-reference.md
+++ b/docs/hooks-reference.md
@@ -365,7 +365,7 @@ do_action( 'desktop_mode_prepare_window', string $page, array $args );
 
 ### `desktop_mode_mode_enabled` — Stable
 
-Gates whether desktop mode can be activated (or stay active) for a given user. The AJAX save endpoint consults this after the nonce check.
+Gates whether desktop mode can be activated (or stay active) for a given user. The AJAX save endpoint consults this after the nonce check; the admin-bar discovery toggle consults it before rendering so users who can't enable the feature don't see a button that does nothing.
 
 ```php
 apply_filters( 'desktop_mode_mode_enabled', bool $enabled, int $user_id );
@@ -382,7 +382,12 @@ add_filter( 'desktop_mode_mode_enabled', function ( $enabled, $user_id ) {
 }, 10, 2 );
 ```
 
-A `false` return means the user cannot toggle the mode on — the AJAX endpoint returns `desktop_mode_disabled`.
+A `false` return has these effects:
+
+- The AJAX `save-desktop-mode` endpoint refuses to flip the user's preference and returns `desktop_mode_disabled`.
+- The "Switch to Desktop Mode" admin-bar entry is hidden for users whose desktop mode is currently off.
+
+Users who already have desktop mode active (`desktop_mode_mode = '1'`) still see the admin-bar entry so they can switch back to classic admin — otherwise a filter flipped to `false` mid-session would trap them in the shell with no exit affordance.
 
 ---
 
--- a/includes/admin-bar.php
+++ b/includes/admin-bar.php
@@ -25,6 +25,34 @@ function desktop_mode_admin_bar_toggle( $wp_admin_bar ) {
 		return;
 	}
 
+	// Discovery affordance: render the toggle for any user who could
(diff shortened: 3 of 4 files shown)


epeicher:
Closing this. After re-reading `includes/admin-bar.php` on trunk, the premise in #98 doesn't hold up: there is no `desktop_mode_is_enabled()` early-return in `desktop_mode_admin_bar_toggle()`. The function only short-circuits on `! is_admin() || ! is_user_logged_in()`, then falls through and adds the node. `$is_active` is just an assignment (line 38) used to flip the label between "Switch to Desktop Mode" and "Switch to Classic Admin" — fresh admins with no `desktop_mode_mode` meta land on the former and see a working discovery affordance.

So the original bug doesn't reproduce, and this PR is fixing something else: it adds a `current_user_can('read') && desktop_mode_mode_enabled` gate so subscribers / filter-denied roles don't see a non-functional toggle. That's a legitimate concern but a different one from "fresh admins can't discover the feature." I'll re-file it as its own issue with the right framing rather than smuggle it in under #98.

Closing #98 too.

PR #125, opened 2026-05-08
fix(admin-bar): make Fullscreen click reliable in browser fullscreen

gavande1:
## Summary

The Fullscreen toggle on this branch sat in `top-secondary` at `top: 0` of the screen, the same band the OS and browser reserve for menu-bar / window-controls / Esc-toast reveal during browser fullscreen. Pointer events landing in that band were intercepted by the OS chrome layer before reaching the page, so the *exit* path was unreliable. Three small belts make the click land:

- **`draggable="false"` on the `<a class="ab-item">`** — WP renders admin-bar items as `<a href="#">`, which HTML5 marks implicitly draggable. Tiny pointer jitter on click was committing to a native link-drag (the floating "Exit fullscreen / http://…#" preview the user saw), so the click handler never ran.
- **`-webkit-user-drag: none; user-drag: none; user-select: none;`** on `.ab-item` and its descendants — WebKit doesn't always honour the `draggable` attribute alone, and disabling text-selection drag prevents the label from absorbing clicks via a selection gesture.
- **`body.desktop-mode-browser-fs` class on `fullscreenchange`** — hides "Switch to Classic Admin" and "Report a bug" while in fullscreen. Both sit closest to the right-edge reveal hot zone; hiding them shifts the Fullscreen toggle further from the corner so its own pointer events reach the page. The class is toggled inside the existing `paintFullscreen()` so OS / Esc exits stay in sync.

The two hidden buttons are surfaced again the moment fullscreen exits — they're not removed, just suppressed for the duration.

## Test plan

- [ ] Reload an admin page with desktop mode active, click the **Fullscreen** button → page enters browser fullscreen, "Switch to Classic Admin" and "Report a bug" disappear from the bar.
- [ ] Click the (now centred-ish) **Exit fullscreen** button → page exits fullscreen, both hidden buttons return.
- [ ] Repeat the exit click with deliberate pointer jitter; verify no native drag preview appears and the click consistently lands.
- [ ] Press **Esc** from inside fullscreen → both hidden buttons reappear (validates the body-class is driven by `fullscreenchange`, not the click handler).
- [ ] Right-click the Fullscreen button → standard browser context menu, no broken interactions.

https://github.com/user-attachments/assets/271e054d-7dbc-4b93-8989-557cebef89f4

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22landingPage%22%3A%22%2Fdesktop-mode%2F%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fdesktop-mode%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-125-ca75ab74a2acf5710fc263c4c8ffa49947329b62.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper (approved):
Checked in the playground and works amazingly! Thanks for the rest of the changes too!

PR #133, opened 2026-05-09
fix(content-graph): address #132 review comments

### LINKED ISSUE #132: Content graph improvements
Hey @epeicher 

I believe we could do some minor improvements here

Image

What about:
- Center icons in the bubbles (See tags, categories, comments, revisions, etc)
- Different orbits for different types? Sort it dynamically, the less elements, the closer orbit, the more elements of type, the further orbit
- The lines that connects satellites with the node, should be BEHIND the node.
- The satellites should have a SELECT state, I don't know when they are selected (a part from the right pane)
- Navigate back button, while it works, seems to abrupt to me. We might add some back animation (iOS style)
- Hove tooltip is QUITE OFF from the hovered node
- What about adding a white bacground to some labels? (or all?) 
Image
- What about FADE IN, FADE OUT animation for the text appearing on zoom level?

epeicher:
## Summary

Batches the eight review items from #132 into a single PR. Addresses the issue end-to-end.

- **Centre dashicons in satellite bubbles** — the font's bbox reserves descender space the icons never use, so the bare `0.5/0.5` anchor parked the visible glyph in the disc's upper half. Anchor lowered to `0.5/0.62` for dashicons plus a per-icon nudge map (`KIND_ICON_NUDGE`) for outliers (`admin-comments`). Disc bumped 12 → 14, glyph 13 → 18.
- **Spokes behind the focused node** — new `spokeLayer` container between `edgeLayer` and `nodeLayer`; the focused node's disc now occludes the spoke origins instead of being painted across by them.
- **Selected satellite state** — each ref gets a synthetic key (`user:42`, `term:category:7`, …) and the matching bubble picks up a thicker stroke + soft halo while its dossier is open. Selection auto-clears on back-to-post navigation via a new `onViewChange` panel callback.
- **iOS-style panel transition** — replaced the abrupt content swap with a sliding page stack: forward = new-from-right + outgoing parallax-left, back = new-from-left parallax + outgoing-to-right. 240ms `cubic-bezier(0.32, 0.72, 0, 1)`, honours `prefers-reduced-motion`.
- **Tooltip pinned to cursor** — Pixi v8's `event.global` is already canvas-local; subtracting the host's viewport rect was producing the drift the reviewer flagged. Dropped.
- **Label backing for readability** — each node label is now a `Container` with a translucent rounded-rect graphic + the text, drawn in `labelLayer` so the whole thing fades together.
- **Smooth label fade on zoom** — replaced the binary `scale > 0.85` cutoff with `smoothstep(0.55, 0.95)` over `world.scale.x`.

### "Different orbits per type" — tried, reverted

Implemented the per-kind concentric rings as suggested in the issue. With the typical 1-2 satellites per kind the result was an off-balance cluster (each kind landed on its own ring at increasing distances, and the staggered start angles bunched everyone on one side). Kept the single-ring layout with even angular distribution — much closer to the reference image attached in the issue. Refs come out of `flattenDetail` already banded by kind, so the ring still reads as colour-grouped as you walk around it.

Closes #132

## Test plan

- [ ] Open Content Graph, focus a post with 4-12 satellites — verify ring is tight and evenly distributed
- [ ] Hover a satellite — tooltip sits next to the cursor (not far away)
- [ ] Click a satellite — bubble shows the selected halo, panel slides forward; click breadcrumb back — panel slides back, halo clears
- [ ] Drag a node — only neighbours move (existing behaviour preserved)
- [ ] Pinch-zoom in/out — labels fade smoothly between ~0.55 and ~0.95 zoom
- [ ] Verify spoke lines start visually from BEHIND the focused node disc
- [ ] Toggle `prefers-reduced-motion` in OS settings — panel transitions become instant

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22landingPage%22%3A%22%2Fdesktop-mode%2F%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fdesktop-mode%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-133-327acbcbae064261dfa76fa25de15b345c095982.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper (commented):
A few comments

image

Not great :)
- Seems the icon centering is not addressed.

Awesome!
- The text transition is superb
- The select state is good!
- The tooltip positioning is great!
- The "Back" smooth transition is nice!
- The post title font could be bigger, matching the rest of labels.

Also
- What about removing the alpha so the lines are truly hidden?
- That "bug"/glitch when a node is slightly moving and we click on it and its not opening is a bit annoying... I believe it has to be with the DRAG state. I don't think we need to drag anything in this view. Removing the drag features would solve it immediately I believe?
-
image

epeicher:
Thanks for your review @AllTerrainDeveloper! I have addressed all the comments:

- Seems the icon centering is not addressed 
- What about removing the alpha so the lines are truly hidden?
CleanShot 2026-05-11 at 12 32 22@2x

- That "bug"/glitch when a node is slightly moving and we click on it and its not opening is a bit annoying... I believe it has to be with the DRAG state. I don't think we need to drag anything in this view. Removing the drag features would solve it immediately I believe? ✅ 

- One more thing, I realised that a "double" loading is happening. That's we can use the hook that we have to. indicate that we have finished load everythign (that way, the BIG W loading icon stays until the initial render is completed: 
  - @AllTerrainDeveloper  I believe this is fixed, but please let me know if that is the case as I am not 100 % sure

This is ready for another review

AllTerrainDeveloper:
Looking!

AllTerrainDeveloper (approved):
Let's ship it! I found an unrelated not-nice UX that we can address on a new issue I will open now

PR #190, opened 2026-05-12
Faster Desktop Mode, main bundle cut by 59 %

AllTerrainDeveloper:
## What changed, in one paragraph

Desktop Mode used to ship one giant JavaScript file with **everything** in
it — every window control, every settings panel, every overlay, the
AI Copilot, the wallpaper animations — even for users who would never
open them. We split it into a small core plus a handful of feature
bundles that only download the moment they're actually needed. Click-to-
react steady-state stays instant because the most-used pieces preload
silently after first paint.

## The headline number

| Main bundle (what every user downloads first) | Before | After    | Saved   |
| --------------------------------------------- | ------ | -------- | ------- |
| Minified                                       | 745 kB | **307 kB** | −59 %  |
| Gzipped (what travels the wire)                | 208 kB | **88 kB**  | −58 %  |

> **More than half of the code is gone from first paint.**

## Estimated time-to-interactive (Desktop Mode opens)

Approximate, on a mid-range laptop with a typical home connection:

| Connection         | Before | After  | Faster by |
| ------------------ | ------ | ------ | --------- |
| Fast Wi-Fi (50 Mbps) | ~0.9 s | ~0.4 s | ~0.5 s    |
| Average 4G          | ~2.4 s | ~1.0 s | ~1.4 s    |
| Slow 3G             | ~9.0 s | ~3.8 s | ~5.2 s    |

ASCII at a glance (Average 4G):

```
Before  ████████████████████████  2.4 s
After   ██████████                1.0 s
```

For a user who never opens a window, never opens settings, and never
talks to the AI, that's the whole story — they pay less than half what
they used to.

## What got lazy-loaded

Each of these now downloads only when the user actually triggers it:

| Feature                              | Roughly | Loads when…                            |
| ------------------------------------ | ------- | --------------------------------------- |
| Window chrome (the window itself)    | ~17 kB gz | First window opens                      |
| Overlays (toasts, dialogs, menus)    | ~16 kB gz | First overlay shows                     |
| OS Settings panel                    | ~15 kB gz | User opens Settings                     |
| AI Copilot                           | ~12 kB gz | User opens the AI panel                 |
| About scene (PixiJS particles)       | ~4 kB gz  | User opens Settings → About             |
| Animated-logo wallpaper              | ~3 kB gz  | User has it selected                    |

The two pieces a user is most likely to touch — windows and overlays —
**preload silently after the desktop has rendered**, so by the time
anyone clicks anything they're already in memory. The lazy bundles only
feel "lazy" on the very first paint.

## What does *not* change

- Existing plugins keep working.
- No visible difference in how the desktop looks or behaves.
- One small breaking API change (`windowManager.open()` is now async) —
  documented in [`docs/migration-0.8.4-async-windowmanager.md`](docs/migration-0.8.4-async-windowmanager.md),
  one-line update per call site.

## Bonus: a 1-in-10 visual bug fixed along the way

While auditing the lazy-loaded code we tracked down a long-standing
race in the **Settings → About** screen where the animated scene would
occasionally render at a single pixel. It is no longer possible — the
scene now waits until its container actually has a size before booting
up.

## Tests

All 1 125 unit tests still green. PHP test suite green. No public API
removed.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22landingPage%22%3A%22%2Fdesktop-mode%2F%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fdesktop-mode%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-190-2c727984e015f6eb5913887955c10e28b8c846e4.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
@epeicher I'd like to do a huddle and heavy test this PR together, please :)

PR #193, opened 2026-05-13
Fix: "Edit Post" from the front-end admin bar opens nothing

### LINKED ISSUE #170: 在前端点击编辑页面，打开的是桌面而非编辑文章页面
如题

```
Plugin version:    0.8.2
WordPress version: unknown
User agent:        Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/123.0.0.0 Safari/537.36
Viewport:          1920x919
Platform:          Win32
Current URL:       https://kmmer.com/wp-admin/post.php?post=1&action=edit
```

</details>

AllTerrainDeveloper:
### What's broken

You're reading a post on the front end with Desktop Mode turned on. You click **Edit Post** in the admin bar (or any other deep link into wp-admin — a bookmark to Plugins, a notification email, anything).

Instead of the desktop opening a window with Gutenberg, you get *just* the desktop — your saved windows are restored, the URL bar shows the post-edit link as a teaser, but the page you actually asked for never appears.

## Why it happens

Desktop Mode treats `/desktop-mode/` as the canonical address. To keep that promise, any plain `/wp-admin/...` URL is bounced through the portal first, then forwarded back into the admin with a `desktop_mode_portal=1` flag. The shell sees that flag and knows "this request came through the portal."

The problem: the boot logic used the same flag to mean two different things at once:

1. "The portal stamped this URL" (a technicality — happens on every entry).
2. "Don't disturb the user's restored windows by auto-opening anything."

Those overlap most of the time — but they're not the same thing. When you click *Edit Post*, the portal still forwards you, *but it forwards you to the page you asked for*. The shell couldn't tell the difference between "user typed `/desktop-mode/` and the portal picked a default landing page" and "user clicked a specific link and the portal preserved it." Anyone with a saved session got the former behaviour for every navigation, so the click was silently swallowed.

## The fix

Split the one signal into two:

- **`fromPortal`** — same as before. "This request went through the portal." Used for stamping the address bar back to `/desktop-mode/`.
- **`fromPortalIntent`** — new. "The portal forwarded here *because* the user supplied a specific URL." Set only when a `?target=` survived the redirect chain (admin-bar clicks, bookmarks, notification links, etc.).

The boot flow now opens the requested page whenever intent is present, regardless of whether there's a saved session. Bare `/desktop-mode/` visits keep the old behaviour — your stack is restored exactly as you left it, nothing is forced on top.

## Knock-on effects

- Bookmarks to specific admin pages work again. Activate, edit, settings deep links — anything you'd reach via a URL — opens in a window even if you already had other windows saved.
- Email notifications that link into wp-admin work again for the same reason.
- A plain visit to `/desktop-mode/` still does what it did: restore your last session, leave it alone.
- Older builds of the shell don't break — the new flag is optional, and when it's missing the previous behaviour kicks in.

## Where to look in the code

- `includes/portal.php` — adds the new flag to the redirect when a `target` was supplied.
- `includes/render/assets.php` — strips both portal flags from `currentPage`, surfaces `fromPortalIntent` in the shell config.
- `src/boot/auto-open.ts` — small extracted helper that decides whether to auto-open. Lives on its own so the five-case decision matrix can be tested directly.
- `src/desktop.ts` — calls the helper; the comment block above the call documents the matrix in plain English.

## Tests

- **PHPUnit**: bare-visit redirects omit the intent flag; target-driven redirects carry it; invalid targets fall back without the intent flag; an already-tagged target doesn't double-stamp. The shell-config filter exposes the new boolean in both `true` and `false` shapes.
- **Vitest**: the eight rows of the auto-open decision matrix are pinned, including the regression case (portal + intent + session ⇒ open).
- Full suites green: 666 PHPUnit, 1133 Vitest.

Closes #170

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22landingPage%22%3A%22%2Fdesktop-mode%2F%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fdesktop-mode%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-193-80867bf4ddc55c5225c21d3e4be4848aa0a80b79.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -80,16 +80,17 @@ Key server-side entry points:
 
 ## Browser flow
 
-1. `/wp-admin/` loads → portal redirect sends the user to `/desktop-mode/`.
-2. `/desktop-mode/` serves a real admin page (Dashboard by default) with the shell wrapped around it.
-3. The shell's Vite-built TypeScript bundle (`desktop.js` in dev, `desktop.min.js` in prod) initializes:
+1. `/wp-admin/...` loads → the portal redirect on `admin_init` bounces the request through `/desktop-mode/?target=<original-url>`.
+2. `/desktop-mode/` (with or without `target`) forwards back into wp-admin tagged with `desktop_mode_portal=1`. When the redirect resolved from a `target` (user-supplied intent — admin-bar link, bookmark, etc.) the URL also carries `desktop_mode_portal_intent=1`. Both flags are stripped from `currentPage` before the shell sees it; the booleans `fromPortal` / `fromPortalIntent` ride along in the shell config so the boot flow can distinguish "portal stamped this URL" from "user actually asked for it."
+3. The landing page renders with the shell wrapped around it (Dashboard by default; the user's saved focused window, the default-window preference, or the `target` URL otherwise).
+4. The shell's Vite-built TypeScript bundle (`desktop.js` in dev, `desktop.min.js` in prod) initializes:
    - Creates the `WindowManager`.
    - Creates the **layout dispatcher** which owns the dock(s) for the active `desktopLayout` (see [Desktop layout modes](#desktop-layout-modes)).
-   - Either restores the saved session (if one exists) **or** opens the current page in a new window.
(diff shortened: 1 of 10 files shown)


demironmanx:
出色的

Issue #178, opened 2026-05-12
Not Updating Edit Pane


chalamministries:
if I have an Edit window open, and I go to edit another post, it just flashes the currently opened Edit window. It doesn't open a new one or change the existing one. I have to close the edit window to edit another post.

## Steps to reproduce

Open Posts
Edit Post
Do not close Edit window
Open Pages (or go back to the Posts window)
Edit another post/page
Notice it does not open a new window or change existing Edit window

<details><summary>Environment</summary>

```
Plugin version:    0.8.2
WordPress version: unknown
User agent:        Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/148.0.0.0 Safari/537.36
Viewport:          1920x1080
Platform:          MacIntel
Current URL:       https://thepiagency.com/wp-admin/index.php?desktop_mode_portal=1
```

</details>

AllTerrainDeveloper:
I can't replicate. MAYBE it's due the upcoming new fixes and patches in the version to be released (0.8.3) that is already solving your issues. I'll keep this open until we release 0.8.3 and you confirm us it's fixed.

https://github.com/user-attachments/assets/f1c8f9dc-c411-4f87-aec5-b313b22947fc

AllTerrainDeveloper:
0.8.3 released :)

chalamministries:
I get a 500 request failed when in Desktop mode to upload any plugin. If I go back to Classic Mode, I can update or install fine

AllTerrainDeveloper:
@epeicher please, you are the master in the plugins scope :) I would appreciate it if you can check the 500 error while uploading a plugin :)  (tomorrow)

epeicher:
Hi @chalamministries, thanks for your report! I have merged #197, which fixes some issues uploading plugins by using the new Desktop native Plugins window. I am closing this task now, but feel free to re-open or create any other issues if you still find issues in the latest versions.

PR #197, opened 2026-05-13
fix(plugins-window): support re-uploading existing plugins, add post-install Activate panel

epeicher:
## Summary

Fixes two issues reported on issue #178 with the native Plugins window's .zip upload flow:

1. **Re-uploading a plugin returned a 500** — the AJAX handler hit Core's `Plugin_Upgrader::install()` without `overwrite_package`, so a second upload of the same plugin (e.g. updating from .zip) failed with `folder_exists`.
2. **No post-install feedback** — the dialog auto-closed after 1.2 s with no indication that the install succeeded, the Installed tab kept showing stale state until Refresh was pressed, and there was no in-place way to activate the freshly-installed plugin.

Mirrors WP Core's classic `update.php?action=upload-plugin` UX inside the native window, without leaving the dialog.

## What changed

**Server (`includes/plugins-window/ajax.php`)**
- Accepts `overwrite=1` and forwards `overwrite_package=true` to the upgrader.
- When the install fails with `folder_exists` and the client didn't ask for overwrite, returns a structured **409** with `code: folder_exists` so the client can prompt.
- On success, returns the just-installed plugin's `plugin_name` + `plugin_version` so the client doesn't need a follow-up round-trip.

**Dialog (`src/plugins-window/upload-dialog.ts`)**
- Catches the 409, shows a `<wpd-confirm-dialog>` "Replace existing plugin?" (red Replace), retries with `overwrite=1` on confirm.
- Replaces the auto-close with an in-dialog success panel: **plugin name + version**, **Activate Plugin** (primary), **Close** (ghost).
- Clicking Activate calls Core REST `PUT /wp/v2/plugins/{plugin}` and swaps the panel to a final "Plugin activated." state with a single **Done** button. User dismisses on their own schedule; no auto-close at any step.
- Publishes `desktop-mode.plugin.changed` (`action: 'install'` then `'activate'`) on the shared cross-view topic so the Installed + Browse tabs repaint without the user hitting Refresh.

**REST client (`src/plugins-window/rest.ts`)**
- `uploadPluginZip( file, { overwrite } )` — passes the flag through as multipart `overwrite=1`.
- Enriched `UploadPluginResult` (`plugin_name`, `plugin_version`).
- `readJsonOrThrow` now unwraps `wp_send_json_error` envelopes before calling `extractAjaxError`, so the `code` and `status` fields land on the thrown error. This was a latent bug for any non-2xx response with a `{ success: false, data: { code, message } }` envelope — only this PR's 409 path exercises it today, but it's a strict improvement for every other admin-ajax call routed through the same helper.

**Window-chrome regression guard (same file)**
While iterating I introduced a regression: importing from `../ui/components/wpd-confirm-dialog/wpd-confirm-dialog` (and `../toast` momentarily) inlined `defineComponent('wpd-confirm-dialog', …)` into `desktop.min.js`. That tag is the canary `src/shell-overlays/loader.ts` uses to detect "bundle already loaded," so the loader short-circuited and `wpd-window-button` never registered — every window's Minimize / Maximize / Fullscreen / Close icons disappeared. The fix is to import from `../wpd-confirm` (the main-bundle-safe shim that lazy-loads the bundle before constructing the element). An inline comment now explains the trap so this doesn't regress.

## Tests

- New vitest cases for `uploadPluginZip` covering the `overwrite=1` passthrough and the surfaced `code: folder_exists` / `status: 409` on the rejected promise.
- Existing test updated to assert the enriched success payload shape.
- `npm run lint`, `tsc --noEmit`, and the full `npm run test:js` suite (1135 tests) pass.
- End-to-end manual verification via Chrome DevTools MCP against the wordpress-develop checkout:
  - First upload → 200, in-dialog "Plugin installed successfully. / `<name> <version>`" → click **Activate Plugin** → REST PUT 200 → "Plugin activated." → click **Done** → dialog closes, plugin shows as Active on the Installed tab without pressing Refresh.
  - Same plugin uploaded again → server returns 409 with `code: folder_exists` → "Replace existing plugin?" prompt → Replace re-uploads with `overwrite=1` → 200 with Core's "Removing the current plugin… Plugin installed successfully." messages → success panel reappears.

## Test plan

- [x] Pull, `npm install && npm run build`
- [x] Upload a fresh plugin .zip from the Plugins window → see the in-dialog success panel
- [x] Click **Activate Plugin** → see the "Plugin activated." confirmation state, then **Done**
- [x] Open the Installed tab → freshly-installed plugin is present and active (no Refresh required)
- [x] Re-upload the same plugin → "Replace existing plugin?" prompt appears; confirm → success panel reappears
- [x] Cancel the prompt → dialog returns to the pre-upload state
- [x] Verify Minimize / Maximize / Fullscreen / Close icons still render on every window

Refs: #178

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22landingPage%22%3A%22%2Fdesktop-mode%2F%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fdesktop-mode%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-197-a0519942607bb7e8c07a6558341ee26e33f7e28c.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/assets/css/plugins-window.css
+++ b/assets/css/plugins-window.css
@@ -918,6 +918,19 @@
 	margin-block-start: 14px;
 }
 
+.desktop-mode-plugins__upload-success-heading {
+	margin: 4px 0 6px;
+	font-size: 1rem;
+	color: rgb( 25, 120, 65 );
+}
+
+.desktop-mode-plugins__upload-success-detail {
+	margin: 0 0 6px;
+	color: var( --wpd-fg, #333 );
+	font-family: ui-monospace, "SFMono-Regular", Menlo, Consolas, monospace;
+	font-size: 0.9em;
+}
+
 /* ─── Window-level .zip drop overlay ────────────────────────────── */
 
 .desktop-mode-plugins__window-drop {
--- a/includes/plugins-window/ajax.php
+++ b/includes/plugins-window/ajax.php
@@ -690,9 +690,55 @@ function desktop_mode_plugins_window_ajax_upload() {
 		return;
 	}
 
+	// phpcs:ignore WordPress.Security.NonceVerification.Missing -- nonce verified above via desktop_mode_plugins_window_ajax_guard().
+	$overwrite = ! empty( $_POST['overwrite'] );
+
 	$skin     = new WP_Ajax_Upgrader_Skin();
 	$upgrader = new Plugin_Upgrader( $skin );
-	$result   = $upgrader->install( $tmp_name );
+	$result   = $upgrader->install(
+		$tmp_name,
+		array( 'overwrite_package' => $overwrite )
+	);
+
+	// Treat "destination folder exists" specially when the caller did
+	// NOT explicitly ask to overwrite — return a 409 with enough
+	// context for the client to prompt the user and re-submit with
+	// `overwrite=1`. WP_Ajax_Upgrader_Skin parks the error on
+	// `$skin->result`; older paths and `Plugin_Upgrader::install()`
+	// itself can also surface it via `$result` directly or by
+	// returning `false`, so check all three. Mirrors Core's classic
+	// `update.php?action=upload-plugin` confirm flow without the
+	// full upload-and-rerun-from-disk dance.
+	$folder_exists = false;
+	if ( ! $overwrite ) {
+		if ( is_wp_error( $skin->result ) && 'folder_exists' === $skin->result->get_error_code() ) {
+			$folder_exists = true;
+		} elseif ( is_wp_error( $result ) && 'folder_exists' === $result->get_error_code() ) {
+			$folder_exists = true;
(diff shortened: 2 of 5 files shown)


AllTerrainDeveloper:
Lovely!

PR #229, opened 2026-05-16
Polish four framework surfaces for plugin authors

AllTerrainDeveloper:
Four small, additive APIs requested by the ODD plugin author. Each one removes a workaround that third-party code was carrying.

## What changed

- **`wp.desktop.widgets.redock( id )`** — stable namespace; `WidgetLayer.redock()` promoted from private, so `wp.desktop.widgetLayer.redock( id )` also works. Idempotent + null-safe.
- **`HOOKS.DESKTOP_ICONS_RENDERED` payload** now ships `{ ids, container, tiles }`. `tiles` is a frozen `id → tile <button>` map, mirroring `DOCK_AFTER_RENDER.tileElements`. `ids` retained for back-compat — existing subscribers keep working.
- **`HOOKS.WINDOW_GEOMETRY`** *(new filter)* — last call before `WindowConfig` is baked. Signature: `( geometry, ctx ) => geometry`, with `ctx = { windowId, baseId, hasSavedGeometry, callerPinned, desktopRect }`. The booleans are the only useful distinctions a filter actually needs: `hasSavedGeometry` flags a restored user layout (bail to respect their choice); `callerPinned` flags caller-passed dims (usually `true` for native windows because their registry passes default `width`/`height`; the filter is free to override). Lets plugins place / size their own windows declaratively instead of compensating after `open()` lands. Filter return is re-clamped to `minWidth`/`minHeight`.
- No new universal "surface rendered" hook — wallpaper, widget, and dock already pass containers; icons were the only gap and the payload change above closes it. Will revisit if a cross-cutting need crystallizes.

## Files

- `src/widgets/layer.ts` — `private redock()` → `public redock()`.
- `src/api/facade.ts`, `src/desktop.ts` — add `wp.desktop.widgets` namespace + reserved key.
- `src/desktop-icons.ts` — build a `tiles` map during render; include `container` + `tiles` in the payload.
- `src/hooks.ts` — new `WINDOW_GEOMETRY` constant; updated `DESKTOP_ICONS_RENDERED` JSDoc.
- `src/window-manager/index.ts` — `applyFilters( HOOKS.WINDOW_GEOMETRY, … )` after geometry resolution; exports `ResolvedWindowGeometry`, `WindowGeometryContext`.

## Docs

- `docs/javascript-reference.md` — `wp.desktop.widgets.redock` section, `WINDOW_GEOMETRY` row in the window-hook table + a full recipe section.
- `docs/examples/register-icon.md` — decorator recipe using `tiles` from the icons payload.

## Test plan

- [x] `./node_modules/.bin/tsc --noEmit` — clean.
- [x] `npm run lint` — clean.
- [x] `npm run test:js` — 1327/1327 (11 new tests across `widgets.test.ts`, `desktop-icons-render.test.ts`, `window-manager-hooks.test.ts`).
- [x] `npm run build` — all 17 bundles compile; `WINDOW_GEOMETRY` constant, `applyFilters` call site, `hasSavedGeometry` + `callerPinned` fields all confirmed in `assets/js/desktop.min.js`.
- [ ] Manual: open Heartbeat widget, liberate it, call `wp.desktop.widgets.redock( 'desktop-mode/heartbeat' )` from console — card snaps back into the column. (Note the namespaced id — the redock is idempotent + silent on unknown ids, so a bare `'heartbeat'` no-ops without error.)
- [ ] Manual *(optional — filter is exhaustively unit-tested)*: drop the following snippet into the browser console BEFORE opening Posts (or after closing it first), then click Posts in the dock. It should land in the bottom-right corner. Drag it somewhere else, close, reopen — it stays where YOU put it (filter bails on `hasSavedGeometry: true`).

  ```js
  wp.desktop.hooks.addFilter(
      wp.desktop.HOOKS.WINDOW_GEOMETRY,
      'manual-qa/bottom-right',
      ( g, ctx ) => {
          // Bail if the user has previously dragged/resized — respect their layout.
          if ( ctx.hasSavedGeometry ) return g;
          const W = 520, H = 360;
          return {
              ...g,
              width: W, height: H,
              x: ctx.desktopRect.width  - W - 20,
              y: ctx.desktopRect.height - H - 20,
          };
      }
  );
  ```

  Want to scope to a single window (e.g. just Posts)? Add an early bail:

  ```js
  if ( ctx.baseId !== 'desktop-mode-posts' ) return g;
  ```

### Defensive coverage for `WINDOW_GEOMETRY`

The filter is wrapped in try/catch + a defensive coalesce so a buggy plugin can't break window opens. Tests pin:

- Default-resolved geometry → can be fully overridden.
- Caller-pinned geometry (`callerPinned: true`) → filter still runs and can override (regression test for the native-window case).
- `hasSavedGeometry: true` set when localStorage has a saved geometry for the window — plugins use this to respect the user's saved layout.
- Filter return below `minWidth`/`minHeight` → re-clamped.
- Partial return (e.g. `{ width }` only) → missing fields fall through to pre-filter values.
- `NaN` / `Infinity` returns → falls through.
- Throwing filter → window still opens, error reported on `HOOKS.SHELL_ERROR`.
- Non-object / `undefined` return → falls through.
- Native windows → filter fires identically.
- Multiple subscribers → chain in priority order.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22landingPage%22%3A%22%2Fdesktop-mode%2F%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fdesktop-mode%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-229-5ddbf9259a46e46a2b94edc372e501068802c798.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/docs/examples/register-icon.md
+++ b/docs/examples/register-icon.md
@@ -174,6 +174,35 @@ If you want to see the error path in action, comment out the `'title'` argument 
 
 The `WP_Error` contract means you find typos at plugin-load time, not at first-click time.
 
+## Decorating the rendered grid
+
+When you need to enhance the wallpaper icons themselves — a cursor adornment, a status dot, a drag handle — subscribe to `HOOKS.DESKTOP_ICONS_RENDERED`. Since 0.25.0 the payload hands you the rendered container *and* a map of `id → tile element`, so your decorator doesn't have to query the DOM (and doesn't have to re-query on every live menu refresh — the hook fires exactly when the grid is rebuilt):
+
+```js
+wp.desktop.hooks.addAction(
+    wp.desktop.HOOKS.DESKTOP_ICONS_RENDERED,
+    'my-plugin/icon-status-dot',
+    ( payload ) => {
+        const { ids, container, tiles } = payload;
+        if ( ! ids.includes( 'jorvy' ) ) {
+            return;
+        }
+        const tile = tiles.get( 'jorvy' );
+        if ( ! tile ) {
+            return;
+        }
+        const dot = document.createElement( 'span' );
+        dot.className = 'my-plugin__status-dot';
+        tile.appendChild( dot );
+        // `container` is the <div class="desktop-mode-icons"> grid root —
+        // use it if your decoration spans multiple tiles (a connector
+        // line, a hover halo) instead of decorating a single tile.
+    }
+);
+```
+
+The hook is suppressed when the rendered DOM is unchanged (fingerprint short-circuit), so decorators run exactly when the grid actually rebuilds — not on every live menu refresh.
+
 ## See also
 
 - [`desktop_mode_register_window()`](../hooks-reference.md#registration-functions) — full argument reference and error-code table.
--- a/docs/javascript-reference.md
+++ b/docs/javascript-reference.md
@@ -2846,6 +2846,19 @@ wp.desktop.registerWidget( {
 
 User-placed geometry (position + size of liberated widgets) persists per-user in `localStorage` under `desktop-mode-widgets-geometry`. Removing a widget clears its stored geometry so a re-add starts docked in the column again.
 
+##### `wp.desktop.widgets.redock( id )` — Stable since 0.25.0
+
+Programmatically un-float a liberated widget back into the right-side column. Idempotent — already-docked widgets and unknown ids silently no-op. Mirrors what the user gets by clicking the re-dock affordance in the floating widget's chrome header.
+
+```js
+// "Reset widget positions" command for a power-user palette.
+for ( const id of wp.desktop.widgetLayer?.getEnabledIds() ?? [] ) {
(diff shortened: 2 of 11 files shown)


AllTerrainDeveloper:
@claude do your awesome review here

AllTerrainDeveloper:
Manual testing passed :)

PR #253, opened 2026-05-20
Fix plugins window stale-nonce 'Cookie check failed' on long-running sessions (#250)

### LINKED ISSUE #250: Plugins table sometimes won't load
Sometimes when I open Plugins, the table is blank and a message appears "Could not load plugins: Cookie check failed." If I do a hard refresh it then works. https://www.dropbox.com/scl/fi/zk78xk728lb91gtukkmst/CleanShot-2026-05-19-at-12.36.30.jpg?rlkey=jm7iqqcrd6ihx7hxcn0on3tsk&dl=0

```
Plugin version:    0.8.5
WordPress version: unknown
User agent:        Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/146.0.0.0 Safari/537.36
Viewport:          2670x1498
Platform:          MacIntel
Current URL:       https://bricks.infinitemonkeys.ca/wp-admin/edit.php?post_type=bricks_template&desktop_mode_portal=1&desktop_mode_portal_intent=1
```

</details>

epeicher:
## Summary

- Refresh shell-cached nonces on every Heartbeat tick so they don't go stale once a tab crosses WordPress's 24h `nonce_life` boundary.
- Server: new `heartbeat_received` filter ships `{ action: nonce }` under `desktop_mode_nonces`. Filter `desktop_mode_nonce_refresh_actions` lets other windows / plugins extend the set.
- Client: subscribes via the existing heartbeat bus, rewrites `desktopModeConfig.restNonce`, every per-window blob's `restNonce`, and the Plugins window's `ajaxNonce` / `updatesNonce` in place. New `registerNonceTarget()` API for plugin authors.

Fixes #250.

## Root cause

`includes/plugins-window/window.php:134` bakes `wp_create_nonce('wp_rest')` into the per-window config blob at page render. The desktop shell is a long-running SPA — once the tab is open past `nonce_life` (24h), the cached `restNonce` is rejected by `rest_cookie_check_errors()` and `/wp/v2/plugins` returns `rest_cookie_invalid_nonce`, surfaced as "Could not load plugins: Cookie check failed". A hard refresh re-mints the nonce, which is exactly what the bug report describes. The same shape exists in every window that stashes a `restNonce` (Posts, Pages, Users, Comments, …); the `wp_rest` target in this fix refreshes all of them generically.

## Why heartbeat refresh

`wp_create_nonce()` returns a stable value within a 12h tick window, so refreshing on every tick (≤60s in the background) means the cached value always reflects the current tick — well before the 24h hard expiry catches it. Same mechanism Gutenberg/Core already use for their own nonces.

## Test plan

- [x] `npm run lint` — clean.
- [x] `tsc --noEmit` — clean.
- [x] `npm run build` — clean.
- [x] `npm run test:js` — 1464/1464 pass (9 new tests in `tests/vitest/nonce-refresh.test.ts`).
- [ ] `npm run test:php -- --filter=Tests_DesktopMode_NonceRefresh` — 7 new tests; could not run locally (port collision with another worktree's wp-env). CI will exercise.
- [ ] Browser sanity: in DevTools console, set `window.desktopModeWindowConfig['desktop-mode-plugins'].restNonce = 'broken'`, try opening Plugins → fails with "Cookie check failed". Wait ~15s for one Heartbeat tick. Reopening Plugins succeeds without a page reload.
- [ ] Soak test: leave the shell tab open >24h, open Plugins — table loads. (Slow, optional.)

## Docs updated

- `docs/hooks-reference.md` — new `desktop_mode_nonce_refresh_actions` filter.
- `docs/javascript-reference.md` — `registerNonceTarget()` surface.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22landingPage%22%3A%22%2Fdesktop-mode%2F%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fdesktop-mode%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-253-b97626412ea77665709c66229b6a0e2fbc28b64c.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/desktop-mode.php
+++ b/desktop-mode.php
@@ -43,6 +43,7 @@ require_once DESKTOP_MODE_DIR . 'includes/assets.php';
 require_once DESKTOP_MODE_DIR . 'includes/admin-bar.php';
 require_once DESKTOP_MODE_DIR . 'includes/session.php';
 require_once DESKTOP_MODE_DIR . 'includes/presence.php';
+require_once DESKTOP_MODE_DIR . 'includes/nonce-refresh.php';
 require_once DESKTOP_MODE_DIR . 'includes/os-settings.php';
 require_once DESKTOP_MODE_DIR . 'includes/seen-intros.php';
 require_once DESKTOP_MODE_DIR . 'includes/welcome-dialog.php';
--- a/docs/hooks-reference.md
+++ b/docs/hooks-reference.md
@@ -2545,6 +2545,41 @@ See [`docs/pwa.md`](./pwa.md) for the full architecture and
 
 ---
 
+## Nonce refresh (since 0.8.7)
+
+The desktop shell is a long-running SPA whose nonces would
+otherwise go stale past WordPress's `nonce_life` (24 h). To
+prevent this, `includes/nonce-refresh.php` registers a
+`heartbeat_received` filter that ships fresh values for a fixed
+set of nonce actions on every Heartbeat tick — the client
+overwrites the cached values in `window.desktopModeConfig` and
+the per-window blobs in place.
+
+### `desktop_mode_nonce_refresh_actions` — Stable *(filter, since 0.8.7)*
+
+Filter the list of nonce-action strings the server refreshes on
+every Heartbeat tick.
+
+```php
+add_filter( 'desktop_mode_nonce_refresh_actions', function ( $actions ) {
+    $actions[] = 'my-plugin/admin-ajax';
+    return $actions;
+} );
+```
+
+- **Param** `string[] $actions` — default set: `[ 'wp_rest',
+  'desktop-mode-plugins', 'updates' ]`.
+- **Return** `string[]` — non-string / empty entries are
+  silently dropped.
+
+Each action string is passed verbatim to `wp_create_nonce()`,
+so it MUST match whatever was used to mint the original cached
+nonce. On the client side, register a `registerNonceTarget()`
+updater (`src/nonce-refresh.ts`) keyed by the same action so the
+incoming value is written to wherever your code reads from.
+
+---
+
 ## See also
 
 - [JavaScript Reference](./javascript-reference.md) — the event + postMessage side of the contract.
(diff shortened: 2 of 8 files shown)


epeicher:
@claude Can you please review this PR with a focus on security implications?

epeicher:
Addressed the two flagged items in b9762641:

**🟡 Medium — `registerNonceTarget` dead-end API trail**

Reclassified as `@internal`. The public extension surface for third-party plugins is already exposed on the facade — `wp.desktop.heartbeat.subscribe('desktop_mode_nonces', cb)` — so I rewrote the doc to use that recipe instead of inventing a second path. `registerNonceTarget` stays as a framework-only helper used by our own built-in updaters; the doc explicitly tells plugin authors not to import it. Avoids the cross-bundle import problem without growing the public API surface for a one-bug fix.

**🟡 Low — heartbeat handler fires for non-shell users**

Gated on `desktop_mode_is_enabled()` rather than `is_user_logged_in()`. Users who haven't opted into Desktop Mode (editors on classic post-edit, subscribers reading the front-end heartbeat) no longer carry the `desktop_mode_nonces` field on their ticks. Added a PHPUnit case covering the opted-out path.

**🟢 Low/Info — console.error reveals action strings**

Left as-is — agreed it's an acceptable trade-off and the action names in our default set (`wp_rest`, `desktop-mode-plugins`, `updates`) carry no sensitivity. Worth surfacing in a future plugin-author guide that custom action names should be treated as non-sensitive.

Test status: `lint`, `tsc --noEmit`, `npm run build` green; 1464/1464 JS tests pass; PHP tests still need CI (local wp-env port collision).

PR #255, opened 2026-05-20
Pre demo polishment

AllTerrainDeveloper:
A set of targeted bug fixes and a feature extension to the user activity footprint. No architectural changes; each fix is self-contained and regression-tested.

---

#### Dock-peek: native-window icons now open correctly ([`src/dock.ts`](https://github.com/WordPress/desktop-mode/blob/pre-demo-polishment/src/dock.ts))

Dock items whose target is a native window (My WordPress, Jorvy, plugin-registered launchers) carry a `windowId` but no `url`. Clicking them in dock-peek was falling through to the generic URL-based iframe path, which opened a chrome-only window with `src=""` — the user saw a title bar with no content.

The fix adds an early guard: if `item.windowId && !item.url`, the click is routed directly through `wp.desktop.openNewWindow()`, spawning a proper native instance. If `openNewWindow` returns falsy (old API, feature flag off) the existing fallback logic handles it transparently.

#### My WordPress: multiple instances no longer interfere with each other ([`src/my-wordpress/index.ts`](https://github.com/WordPress/desktop-mode/blob/pre-demo-polishment/src/my-wordpress/index.ts))

Three related issues when two My WordPress windows were open at the same time:

1. **Close clobbered live siblings.** The previous teardown used a `desktop-mode-window-closed` document listener matched on the base `WINDOW_ID`. Every sibling's close event fired the same handler — closing instance B cleared instance A's state and left it broken. Fixed by having `renderInto` return a per-instance teardown that the framework calls only for *that* specific window.

2. **Navigation in a non-foreground instance threw.** `asRenderState` compared against the singleton `activeState`. Drilling into a posts list inside a backgrounded duplicate triggered *"host body does not match active state"*. Fixed by introducing a `liveStates: Map<HTMLElement, RenderState>` that lets `asRenderState` resolve the correct state for any live instance.

3. **Drag-rejection drop targets overwrote each other.** Both instances registered a claimant with the same id (`my-wordpress-reject`), so the second registration silently replaced the first's element binding. Fixed by appending a per-mount monotonic counter (`rejectIdCounter`) to the id.

#### UI: `willReadFrequently: true` on pixel-readback canvas contexts ([`src/ui/components/wpd-crumb-chain/wpd-crumb-chain.ts`](https://github.com/WordPress/desktop-mode/blob/pre-demo-polishment/src/ui/components/wpd-crumb-chain/wpd-crumb-chain.ts), [`src/ui/util/avatar-resolve.ts`](https://github.com/WordPress/desktop-mode/blob/pre-demo-polishment/src/ui/util/avatar-resolve.ts))

Both `wpd-crumb-chain` (foreground-color picker) and `avatar-resolve` (transparent-pixel probe) create a 1×1 canvas and immediately call `getImageData`. Without the `willReadFrequently` hint Chrome backs the canvas with a GPU texture, so each `getImageData` triggers a GPU→CPU sync. The breadcrumb chain re-paints on every navigation; the avatar resolver fires once per unique URL across the page — often enough that Chrome was surfacing the "willReadFrequently" performance warning in both cases.

---

### Feature: User activity footprint now tracks post updates / revisions

**`includes/my-wordpress/user-footprint.php` · `src/my-wordpress/index.ts` · `src/my-wordpress/types.ts`**

The activity heatmap and stats previously only counted published posts and comments. Editors who spend most of their time updating existing content (the typical CMS workflow) appeared to have no activity. This brings the footprint model in line with how GitHub counts contributions — edits to existing content are first-class activity.

**What counts as an "update":** a WordPress revision saved by the user *after* the parent post's initial creation (`r.post_date_gmt > p.post_date_gmt`). The initial save is excluded so it doesn't double-count with the "posts" column. Updates are attributed to whoever hit Save (via `r.post_author`), not the post owner — consistent with tracking editorial work across a team.

**Changes in detail:**

- **`daily[].updates`** — per-day update count, added alongside the existing `posts` and `comments` fields (zero-filled for days with no saves).
- **`totals.updates`** — lifetime revision count, excluding the initial save.
- **`timeline` `post-update` kind** — revisions are collapsed per parent post (`GROUP BY r.post_parent, MAX(post_date_gmt)`) so a burst of saves on one article reads as a single row rather than flooding the timeline. Displayed with a pencil icon and "Updated `"…"`" label.
- **`timeline[].type`** — CPT slug (`'post'`, `'page'`, custom type) added to `post` and `post-update` rows so the renderer can pick a Post-vs-Page icon without a second REST lookup.
- **Streak logic** — "active day" now includes update and comment activity, not only publish days.
- **UI "Updates" stat card** — shown in the My WordPress stats panel when the lifetime count is > 0. Heatmap intensity now sums `posts + comments + updates`.

---

### Tests

- **`tests/vitest/dock-peek-native-icon-spawn.test.ts`** — verifies that a `windowId`-only dock item spawns a native window and does not fall through to the iframe path.
- **`tests/vitest/my-wordpress-multi-instance.test.ts`** — covers the three multi-instance regressions above: close isolation, independent navigation, and unique drop-target ids.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22landingPage%22%3A%22%2Fdesktop-mode%2F%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fdesktop-mode%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-255-ad7f93ecce0e00ba1be4064a8f8346ffdda3de5f.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/my-wordpress/user-footprint.php
+++ b/includes/my-wordpress/user-footprint.php
@@ -25,10 +25,18 @@
  *     hour:    [ 0..23 ],                                      // post counts, server-local hour
  *     streak:  { longest, current, longestRange:{ from, to } },
  *     timeline:[                                               // 30 most recent activity rows
- *       { kind:'post'|'comment'|'post-update', date, title, link, status, postId? }
+ *       { kind:'post'|'comment'|'post-update', date, title, link, status, postId?, type? }
  *     ],
  *     totals:  { posts, pages, comments, updates, mostProlificMonth?:{ ym, n } }
  *   }
+ *
+ * Timeline row fields:
+ * - `kind`   — discriminator: `'post'` (publish), `'comment'`, or
+ *              `'post-update'` (revision rollup).
+ * - `type`   — only set when `kind` is `'post'` or `'post-update'`.
+ *              Carries the post's CPT slug (`'post'`, `'page'`, custom
+ *              types) so the renderer can pick a Post-vs-Page icon
+ *              without a second REST lookup.
  *
  * "Updates" are revisions saved by the user AFTER a post's original
  * creation — i.e. the user opened an existing post and saved it
--- a/includes/my-wordpress/user-footprint.php
+++ b/includes/my-wordpress/user-footprint.php
@@ -19,16 +19,22 @@
  *
  *   {
  *     profile: { id, name, avatarUrl, link, roleLabels?, registered? },
- *     range:   { from, to, days },             // YYYY-MM-DD bookends + day count
- *     daily:   [ { date, posts, comments } ], // length = range.days; missing days = 0/0
- *     weekday: [ 0..6 ],                      // post counts, Sunday-indexed
- *     hour:    [ 0..23 ],                     // post counts, server-local hour
+ *     range:   { from, to, days },                              // YYYY-MM-DD bookends + day count
+ *     daily:   [ { date, posts, comments, updates } ],         // length = range.days; missing days = 0
+ *     weekday: [ 0..6 ],                                       // post counts, Sunday-indexed
+ *     hour:    [ 0..23 ],                                      // post counts, server-local hour
  *     streak:  { longest, current, longestRange:{ from, to } },
- *     timeline:[                              // 30 most recent activity rows
- *       { kind:'post'|'comment', date, title, link, status, postId? }
+ *     timeline:[                                               // 30 most recent activity rows
+ *       { kind:'post'|'comment'|'post-update', date, title, link, status, postId? }
  *     ],
- *     totals:  { posts, pages, comments, mostProlificMonth?:{ ym, n } }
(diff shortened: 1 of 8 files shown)


AllTerrainDeveloper:
@claude this is a PR that mostly fix small bugs, and also adds updates to the activity footprint. Can you review it?

AllTerrainDeveloper:
@claude I'm testing you, can you even fill the PR description based on the changes in the commits?

PR #261, opened 2026-05-21
Drag-and-drop follow-up: review fixes

AllTerrainDeveloper:
## Summary

Follow-up to #260 (Drag and drop improvements). Five code-review items from a post-merge audit:

1. **`console.log` → `console.info`** in `src/drag/iframe-drop-targets.ts`. `no-console` ESLint config allows `info`/`warn`/`error` only; this drops the inline `eslint-disable` and uses the more semantically accurate level for the drag-start diagnostic.

2. **Mid-file imports in `src/desktop-files/layer.ts`**. `buildBridgePayloadFromPlacement` had been inserted between two groups of imports, which would have tripped `import/first`. Moved the function below all imports.

3. **Block `javascript:` / `data:` / `vbscript:` URLs in the Gutenberg drop-receiver** (`src/gutenberg-drop-receiver.ts`). Same-origin postMessage is the first line of defence; this is defence-in-depth so a hostile URL surviving into a `core/file` href or `core/paragraph` anchor can't become click-to-XSS. Returns `null` (silent no-op) for hostile URLs — same shape as "empty URL → no insert".

4. **Drop `site-editor.php` enqueue** in `includes/render/assets.php`. The FSE's `core/block-editor` store isn't reliably available without an opened template in the canvas, so drops would silently time out at the receiver's 5 s `waitForEditor` poll. Also added a `desktop-mode-drop-failed` postMessage from the receiver → parent toast pipeline so any future "editor not ready" failure surfaces as user-visible feedback instead of just a `console.error`.

5. **`Map` mutation during `forEach`** in the `WINDOW_CLOSED` cleanup of `iframe-drop-targets.ts`. Spec-safe but reads as a hazard; switched to `for…of` over `Array.from()` snapshots for both `_suppressedIframes` and `_activeRegistrations`.

## Test plan

- [x] `npm run lint` clean
- [x] `npm run typecheck` clean
- [x] `npm run test:js` — 1485 passing (3 new vitests cover `javascript:` / `data:` URL rejection across post / user / attachment payloads, plus the leading-whitespace + uppercase variants)
- [x] `npm run build:desktop` + `npm run build:gutenberg-drop-receiver` — clean
- [ ] Manual: drag a media tile / post shortcut / wallpaper shortcut into an open Gutenberg post window; confirm block insertion
- [ ] Manual: open Gutenberg, wait, then drag in; confirm no spurious "Could not insert" toast (receiver still works after editor finished booting)

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22landingPage%22%3A%22%2Fdesktop-mode%2F%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fdesktop-mode%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-261-7d69aef6613868440334bb1f789a4cf8b7c1e200.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/render/assets.php
+++ b/includes/render/assets.php
@@ -45,19 +45,18 @@ function desktop_mode_enqueue_assets() {
 
 		// Block Editor cross-window drop receiver. Listens for
 		// `desktop-mode-drop` postMessages from the parent shell and
-		// inserts the matching block. Only enqueue inside the Block
-		// Editor screens — every other admin page would be paying for
-		// a bundle it never uses. `$hook_suffix` is supplied by the
-		// `admin_enqueue_scripts` action; `post.php` / `post-new.php`
-		// cover both edit + new-post flows. Gutenberg also drives
-		// `site-editor.php` (full-site editing) — included so dropping
-		// onto a template / template-part window also works.
+		// inserts the matching block. Only enqueue inside the
+		// post-edit Block Editor screens — every other admin page
+		// would be paying for a bundle it never uses.
+		//
+		// `site-editor.php` (full-site editor) deliberately omitted:
+		// the FSE doesn't expose `wp.data.dispatch('core/block-editor')`
+		// until the user opens a template in the canvas iframe, so
+		// drops arriving before that point would silently time out
+		// after the receiver's 5 s `waitForEditor()` poll. Re-enable
+		// once we have a reliable readiness signal in that context.
 		global $hook_suffix;
-		if (
-			'post.php' === $hook_suffix
-			|| 'post-new.php' === $hook_suffix
-			|| 'site-editor.php' === $hook_suffix
-		) {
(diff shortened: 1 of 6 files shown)


epeicher:
I have tested the Drag and Drop on both Mac and Windows and I have not found any issues. ✅

PR #300, opened 2026-06-07
Fix placement, gear/Help, and dock⇄desktop visibility bugs

AllTerrainDeveloper:
## Summary

Fixes a cluster of app-placement, screen-meta, and item-visibility bugs surfaced by a behavioral audit. Two were user-reported (My WordPress missing from the wallpaper; gear/Help buttons showing with nothing behind them); the rest are adjacent correctness issues found while tracing those paths. Every finding was reproduced in code and adversarially verified before fixing.

No public-API breakage. The only documented surface touched is the `desktop-mode-screen-meta` bridge message, whose change is backwards-compatible (an empty `panels: []` was already a valid payload) — `docs/javascript-reference.md` is updated in the same PR.

## Why

- **My WordPress vanished from the desktop on some installs.** Root cause: the Spatial desktop layout silently dropped *all* server-registered desktop icons. The only recovery was OS Settings → Apps → move it to the dock.
- **Gear / Help title-bar buttons appeared with nothing to show.** Two causes: the buttons were never cleared when an iframe navigated in place to a page with no screen meta, and detection keyed on element presence rather than panel content.
- **"Hide" on a dock/desktop tile didn't hide it** for single-rail items — it teleported the tile to the *other* rail.

### 🔴 High

- **Spatial layout no longer drops the pinned My WordPress icon** — `src/desktop-layout.ts`
  `repaintIcons()`'s spatial branch now keeps server icons that are either `pinned` (framework-owned, e.g. My WordPress) or carry an explicit `desktop`/`both` placement override. Non-pinned plugin icons with no override stay suppressed (the deliberate "core surface" design). This also fixes the OS Settings "On the desktop" picker silently no-opping in Spatial.

- **Right-click "Hide" stops teleporting items** — `src/item-visibility-menu.ts`
  "Hide from dock/desktop" now derives the target from the item's *current* placement via a new `computeHideTarget()` (`both` → the other rail; single-rail → `hidden`) instead of unconditionally writing the opposite rail. Native rail is derived from the rail-synthesis id prefix.

### 🟠 Medium

- **Stale gear/Help cleared on navigation** — `src/iframe-bridge-standalone.ts`, `includes/render/chromeless-bridge.php`
  Both bridges now *always* announce `desktop-mode-screen-meta` (including an empty `panels: []`) so the parent removes stale buttons when a page exposes no screen meta after an in-place same-slug navigation.

- **Visibility menu is state-aware** — `src/item-visibility-menu.ts`
  "Also show on \<rail\>" is hidden when the item is already on `both` (it was a no-op re-write).

- **OS Settings → Apps & Icons repaints live** — `src/settings/sections/apps-icons.ts`
  The tab subscribes to `wp.desktop.subscribeOsSettings` and repaints on external writes (e.g. a right-click visibility change), self-unsubscribing once the panel is torn down. Previously the row dropdown showed stale placement until the panel was rebuilt.

- **Reset no longer corrupts the DEFAULTS singleton** — `src/settings/state.ts`, `src/settings/panel.ts`
  `structuredDefaults()` now deep-clones the collection fields (`itemVisibility`, `dockOrder`, `dockPromotedPositions`), and `onReset` uses it instead of a shallow `{ ...DEFAULTS }`. Fixes the custom-gradient editor mutating the module-level `DEFAULTS` for the rest of the session after a Reset.

- **`dockOrder` survives the PHP round-trip** — `includes/os-settings.php`
  Cross-rail tiles persist with a rail-synthesis prefix (`desktop:<id>` / `dock:<id>`). `sanitize_key()` stripped the colon, silently resetting the user's order (and risking an id collision) on reload. The `dockOrder` sanitizer now preserves the colon while still rejecting anything outside the JS id charset. (itemVisibility / dockPromotedPositions keys are canonical and unaffected.)

### 🟡 Low

- **Content-less Screen Options / Help no longer surface a dead button** — both bridges
  A panel is announced only when its toggle link is present **and** the panel has renderable content (form controls for Screen Options; non-empty tab/sidebar text for Help).

- **Contract drift corrected** — `src/settings/types.ts`, `docs/javascript-reference.md`
  The `itemVisibility` docblock wrongly claimed `'both'` is never persisted; it *is* (PHP whitelists it, the menu stores it). Doc updated, plus the empty-array `screen-meta` semantics.

- **`dockPromotedPositions` no longer leaks** — `src/settings/desktop-shortcuts-sync.ts`
  When a promoted item is demoted/hidden, its persisted drag position is pruned (reentrancy-safe), so stale coordinates can't accumulate toward the 256-entry cap or silently resurrect on a future re-promote.

> Investigated and **dismissed** (no change): the "first-run seeding / absent key resolves to hidden" theory. `resolvePlacement` correctly falls back to the native rail for an unset key, so fresh and existing accounts behave identically — the real My WordPress cause was the Spatial branch above.

## Testing

- `npm run typecheck` ✓
- `npm run lint` ✓
- `npm run test:js` ✓ — **1548 passed** (added spatial-layout coverage: pinned icon kept, explicit promotion kept, non-pinned/no-override still dropped)
- `npm run build` ✓
- `php -l` ✓ on both changed PHP files
- `npm run test:php` ✓ — **926 tests / 2175 assertions** (added `dockOrder` colon-preservation + normalization tests)

## Notes for reviewers

- **Spatial design decision (#1):** I kept the deliberate suppression of *non-pinned, no-override* plugin desktop icons in Spatial — only `pinned` framework icons and explicitly-promoted icons survive — to avoid unilaterally changing the documented "core surface" behavior. If we'd rather Spatial mirror Classic/Unified and show all desktop-resolved server icons, it's a one-line change to the filter.
- The screen-meta fix is applied to **both** the standalone TS bridge and the inline PHP chromeless bridge so they stay in sync.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22landingPage%22%3A%22%2Fdesktop-mode%2F%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fdesktop-mode%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-300-6d146d2eddb0f3231eb390bb4260b8546c54ec15.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/docs/javascript-reference.md
+++ b/docs/javascript-reference.md
@@ -2791,11 +2791,13 @@ Posted by the chromeless bridge's `fetch` and `XMLHttpRequest` wrappers whenever
 ```
 
 #### `desktop-mode-screen-meta` — Stable
-Announces the screen-meta panels (Screen Options / Help) that the iframe page exposes. The parent renders corresponding title-bar buttons.
+Announces the screen-meta panels (Screen Options / Help) that the iframe page exposes. The parent renders one title-bar button per announced panel, replacing any previously rendered set.
 
 ```typescript
 { type: 'desktop-mode-screen-meta'; panels: ( 'screen-options' | 'help' )[] }
 ```
+
+The iframe sends this on every load — **including an empty `panels: []`** — so the parent can clear stale buttons when a page (e.g. after an in-place same-slug navigation) exposes no screen meta. A panel is announced only when its toggle link is present **and** the panel actually has content: a Screen Options panel with no form controls, or a Help tab registered with empty `content` and no callback, is omitted so the title bar never shows a button that opens an empty panel.
 
 #### `desktop-mode-screen-meta-state` — Stable
 Reports which screen-meta panel (if any) is currently open inside the iframe.
--- a/includes/os-settings.php
+++ b/includes/os-settings.php
@@ -464,7 +464,13 @@ function desktop_mode_sanitize_os_settings( $raw ) {
 		}
 	}
 
-	// dockOrder — ordered list of sanitize_key()-clean ids.
+	// dockOrder — ordered list of item ids. Most are sanitize_key()-
+	// clean dock slugs, but cross-rail tiles the user promoted carry a
+	// rail-synthesis prefix (`desktop:<id>` / `dock:<id>`, built by
+	// src/settings/item-placement.ts). sanitize_key() strips the colon,
+	// which silently breaks the JS order match on reload and can collide
+	// with an unrelated id — so allow the colon (and hyphen/underscore)
+	// while still rejecting anything outside the JS id charset.
 	$dock_order = array();
 	if ( isset( $raw['dockOrder'] ) && is_array( $raw['dockOrder'] ) ) {
 		$seen = array();
@@ -472,7 +478,7 @@ function desktop_mode_sanitize_os_settings( $raw ) {
 			if ( ! is_string( $id ) || '' === $id ) {
 				continue;
 			}
-			$slug = sanitize_key( $id );
+			$slug = (string) preg_replace( '/[^a-z0-9_:-]+/', '', strtolower( $id ) );
 			if ( '' === $slug || isset( $seen[ $slug ] ) ) {
 				continue;
 			}
--- a/includes/render/chromeless-bridge.php
+++ b/includes/render/chromeless-bridge.php
@@ -1949,29 +1949,55 @@ function desktop_mode_chromeless_bridge_script() {
 	}
 	window.__desktopModeScreenMetaInstalled = true;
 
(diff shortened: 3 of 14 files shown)


AllTerrainDeveloper:
@claude can you review this pull request?

PR #312, opened 2026-06-19
Improve minimized window UX with restore support and count badges

### LINKED ISSUE #311: Dock Peek Interactivity and Minimized Window UX
## 1. Unclickable Dock Peek Thumbnails

- When hovering over a dock icon that has minimized windows, the "Dock Peek" menu appears showing thumbnails/options of the minimized windows (e.g., "Window 1", "Window 2", and "Add New Window"). 
- Currently, the thumbnails for existing minimized windows are unresponsive and cannot be clicked to restore the window. Users are forced to click the dock icon itself to reopen windows, breaking the expected preview interaction flow.

https://github.com/user-attachments/assets/5248a35f-7c39-445e-b198-9e57f55eb9b0

## 2. Inadequate Minimized Window State Indicators

The visual cues on the dock icons (the edge indicators) do not sufficiently communicate the window states to the user, especially when dealing with multiple instances of the same application. 
- The default indicator (a small circle) does not convey how many instances are currently open or tucked away.
- When multiple instances exist and one is actively focused, the dock prioritizes displaying the "Focus Bar." This entirely masks the fact that there are other minimized instances in the background, as there is no distinct numerical count or visual cue for the tucked-away windows. 

## Expected Behavior
- **Interactivity:** Clicking a minimized window's thumbnail in the Dock Peek menu should immediately restore and focus that specific window.
- **Visibility of State:** Users need a clear, numerical visual cue indicating the exact number of minimized windows an application has. This count indicator must remain consistently visible even if another instance of the same application is currently active or focused, allowing the user to seamlessly track background tasks.

KarunyaChavan:
This PR fixes unresponsive minimized window thumbnails in the Dock Peek popover, resolves ghost-focusing bugs when hovering over minimized cards, and significantly improves the dock UX by introducing a dedicated, accessible counter badge for tucked-away windows. It also patches an aggressive browser caching issue that prevented dock styling updates from loading.

Closes #311 

https://github.com/user-attachments/assets/76761ef4-fd18-46db-84e1-10084178ac67

## What Changed
- **`src/dock-peek/index.ts`**: Added window restoration logic to the Dock Peek click handler and guarded the hover behavior against minimized windows.
- **`src/dock.ts`**: Added state logic to compute `data-minimized-count` and inject a dynamic `aria-label` into the tile for screen reader accessibility.
- **`assets/css/dock.css`**: Decoupled the edge state indicator from the window count. Introduced a new `::after` pseudo-element to render a dedicated, non-colliding count badge at the bottom-right of the dock tile.
- **`includes/assets.php`**: Updated the CSS enqueue registration from a static version to dynamic `filemtime` versioning.
- **`tests/vitest/*`**: Added comprehensive test coverage for all new state transitions, hover/click behaviors, and accessibility labels.

## How it fixes
- **Dock Peek Interactivity:** 

    - Previously, clicking a minimized window thumbnail in the Dock Peek menu failed to restore the window to the screen. 
    - By invoking `win.restore()` inside the `spawnFocusViewTransition` click handler, minimized instances now correctly unpack to the viewport before receiving focus.

- **Hover Ghost-Focusing:** 

    - Hovering over a minimized instance card was automatically triggering the window manager's focus logic, improperly shifting focus to an invisible/tucked window. 
    - Adding an early return (`if (win.state === 'minimized')`) inside the `pointerenter` listener cleanly prevents this unintended hijack.

- **Dedicated Minimized Badges:** 

    - The dock edge indicator (e.g., the focus bar) used to completely mask the count of tucked-away windows if another instance was actively focused. 
    - This PR completely decouples them. The edge indicator now strictly reflects the highest priority app state, while a brand new, macOS-style hollow ring badge dynamically displays the exact count of minimized instances. 
    - We anchored it to the bottom-right corner (`bottom: 0`, `inset-inline-end: -4px`) so it gracefully avoids colliding with the red unread-notification pills at the top-right.

- **Screen Reader Accessibility:** 

    - The dock tile now dynamically updates its `aria-label` (using `_n()` and `sprintf()`) so screen readers accurately announce `"[App Name], X minimized windows"` to visually impaired users, keeping them fully aware of background tasks.

- **Reliable CSS Cache Busting:** 
    - Using the static plugin version in `includes/assets.php` caused aggressive browser caching for `dock.css`. Switching to `$built_version( 'assets/css/dock.css' )` leverages file modification times, forcing the browser to reliably fetch the latest stylesheet changes after deployments.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22landingPage%22%3A%22%2Fdesktop-mode%2F%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fdesktop-mode%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-312-799b1d3d1d1a28befd8028e8a7ace267d54404f0.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
Hey @KarunyaChavan! Thanks for your contribution, we've been focused on other duties and we've just came back to Desktop Mode, and I love your focus on the thumbnails work. It definitely need some love and we are super grateful about you focusing on it. 

I would love to hear your thoughts about slightly changing the implementation of it. A second badge indicating the number of minimized tabs it might feel confusing for the user (eg, 2 badges at the same time might look crazy! 😆)

So my question is, how do you feel about actually changing the preview state of the thumbnail? I mean, removing the "content" and just showing up the title bar for the minimized window.

image

KarunyaChavan:
Hey @AllTerrainDeveloper, thanks for taking the time to review the changes and for the thoughtful feedback!

> I would love to hear your thoughts about slightly changing the implementation of it. A second badge indicating the number of minimized tabs it might feel confusing for the user (eg, 2 badges at the same time might look crazy! 😆)

Yes, I think it's better to change the current indication method for minimized windows instead of using a badge. We could take inspiration from the stacked/layered feeling in Windows OS. Let me know what you think about that.

> So my question is, how do you feel about actually changing the preview state of the thumbnail? I mean, removing the "content" and just showing up the title bar for the minimized window.

Sounds good.

AllTerrainDeveloper:
> Yes, I think it's better to change the current indication method for minimized windows instead of using a badge. We could take inspiration from the stacked/layered feeling in Windows OS. Let me know what you think about that.

I'm open to anything, we can aim for creativity or simplicity.

KarunyaChavan:
Hey @AllTerrainDeveloper,

Please refer to the video below for the updated behavior of the minimized windows UX:

https://github.com/user-attachments/assets/c34e2a77-f4f0-4b92-b878-1d6d16e2cd5a

## Changes Overview

### 1. Persist tile highlight while the Peek popover is open

On `trunk`, moving the cursor from the dock icon into the Peek popover causes the dock tile to lose its highlight, making the interaction feel disconnected. This change introduces the `data-peek-active` state on the dock tile, ensuring the icon remains highlighted for the entire time the Peek popover is visible. The behavior now aligns with the Windows taskbar experience.

|                                                              Trunk                                                             |                                                            This Branch                                                           |
| :----------------------------------------------------------------------------------------------------------------------------: | :------------------------------------------------------------------------------------------------------------------------------: |
| Trunk behavior | Updated behavior |

### 2. Dock stacked card visual

When multiple windows of the same application are open, the dock now displays a stacked card visual, similar to the Windows taskbar, providing a clearer indication that multiple windows are available.

<div align="center">
Dock stacked card visual
</div>

----

I'd love to get your thoughts on the preview state of the thumbnails.

As you can see in the video above, we're currently showing only the title bars for **minimized** windows, whereas for **non-minimized** windows we display the window content. Do you think we should make the behavior consistent for both states, or do you prefer keeping the current distinction?

AllTerrainDeveloper:
Nice! I'll review and test it tomorrow morning, I had many duties today :)

src/dock-peek/index.ts
```diff
@@ -348,20 +350,23 @@ function buildInstanceCard(
...
 	card.addEventListener( 'pointerenter', () => {
-		if ( deps.windowManager.getFocused() === win ) {
-			return;
+		if ( win.state === 'minimized' ) {
+			win.restore();
+		} else if ( deps.windowManager.getFocused() !== win ) {
+			deps.windowManager.focus( win );
 		}
-		deps.windowManager.focus( win );
 	} );
```
AllTerrainDeveloper, line 384: The `pointerenter` handler restores the window here, but never clears `card.dataset.state`, which was stamped to `minimized` at build time (lines 353 to 356 above). The CSS rule for `[data-state="minimized"]` keeps collapsing the card to just its titlebar after that, so the peek card stays visually collapsed even though the underlying window is no longer minimized, which works against the stated goal of this change (restore on hover so the peek card actually shows the window). Clearing the attribute alongside the restore call fixes it:

```suggestion
	card.addEventListener( 'pointerenter', () => {
		if ( win.state === 'minimized' ) {
			win.restore();
			delete card.dataset.state;
		} else if ( deps.windowManager.getFocused() !== win ) {
			deps.windowManager.focus( win );
		}
	} );
```

assets/css/dock.css
```diff
@@ -77,6 +77,24 @@
...
 
+/* Stacked card hint on hover — a second card peeks from behind the
+   icon when the tile has multiple open windows. Only visible on hover so
+   it never conflicts with focused tiles. */
+.desktop-mode-dock__item--stacked:hover .desktop-mode-dock__item-primary::before,
+.desktop-mode-dock__item--stacked[data-peek-active] .desktop-mode-dock__item-primary::before {
+	content: "";
+	position: absolute;
+	top: -4px;
+	left: -4px;
```
AllTerrainDeveloper: This hint uses a hardcoded physical `left` offset. Every other absolutely positioned decoration in this file (badges, dots, chips around lines 222, 465, 534, 761) uses `inset-inline-start` or `inset-inline-end` so it mirrors correctly on RTL locales. As written, this hint will sit on the wrong edge in RTL. Switching to the logical property matches the rest of the file:

```suggestion
	inset-inline-start: -4px;
```

src/dock-peek/index.ts
```diff
@@ -398,6 +403,9 @@ function spawnFocusViewTransition(
 	const vtName = `desktop-mode-peek-card-${ win.id }`;
 	const focus = (): void => {
 		dismiss();
+		if ( win.state === 'minimized' ) {
+			win.restore();
+		}
```
AllTerrainDeveloper: Small cleanup, not blocking: this minimized check and restore is duplicated verbatim between here and the `pointerenter` handler above (lines 365 to 366). Pulling it into a small shared helper would keep both call sites in sync if the restore condition ever changes, something like:

```ts
function restoreIfMinimized( win: WPWindow, card?: HTMLElement ): void {
	if ( win.state !== 'minimized' ) {
		return;
	}
	win.restore();
	if ( card ) {
		delete card.dataset.state;
	}
}
```

Then both call sites become `restoreIfMinimized( win, card )` and `restoreIfMinimized( win )`. Not a blocker, just flagging so the two paths do not drift apart.

AllTerrainDeveloper (commented):
Ran a high effort review over this branch's diff against trunk. Two correctness issues and one small cleanup, pinned to the lines below.

AllTerrainDeveloper:
Pushed 2e005d6 addressing the review above:

- Cleared the peek card's collapsed `data-state="minimized"` when the underlying window is restored (hover and click paths), via a shared `restoreIfMinimized()` helper so the two call sites cannot drift apart again.
- Switched the stacked card hint in dock.css from a hardcoded `left` offset to `inset-inline-start` so it mirrors correctly in RTL.

Verified locally: `npm run build`, `npm run lint`, `npm run typecheck`, and `npm run test:js` (full suite, 1843 tests) all pass.

AllTerrainDeveloper:

PR #338, opened 2026-07-13
Fix Overview keyboard navigation and focus trapping

### LINKED ISSUE #331: Overview keyboard navigation issues
Overview mode has two keyboard accessibility issues that make navigation difficult for keyboard-only users:

1. Focus is not constrained to the Overview UI, causing `Tab` navigation to traverse hidden page elements before reaching the Overview controls.
2. Desktop tile close buttons are unreachable using the keyboard because they are not part of the browser's sequential focus order.

Together, these issues prevent efficient keyboard interaction and do not meet expected accessibility behavior for a modal-like interface.

---

### Description

When Overview mode is opened, keyboard focus is not restricted to the Overview interface. Although large portions of the WordPress admin UI are visually hidden, they remain part of the document's sequential tab order.

As a result, pressing `Tab` causes focus to move through numerous invisible elements - including the left admin navigation (admin side-panel) , page content, and window iframes - before eventually reaching the desktop tiles.

### Steps to Reproduce

1. Enable Desktop Mode.
2. Open two or more windows.
3. Enter Overview mode.
4. Press `Tab` repeatedly.

### Expected Behavior

Keyboard focus should remain within the Overview interface, allowing users to navigate only between the available Overview controls (Ones visible in the UI).

### Actual Behavior

Focus traverses hidden elements throughout the page before eventually reaching the Overview controls, resulting in dozens of invisible tab stops.

---

### Description

Each desktop tile includes a close ("X") control, but it cannot be reached using the `Tab` key. Keyboard users can focus the desktop tile itself but cannot navigate to its close action.

### Steps to Reproduce

1. Create two or more virtual desktops.
2. Enter Overview mode.
3. Use `Tab` to navigate through the desktop tiles.

### Expected Behavior

Both the desktop tile and its close control should be independently reachable using keyboard navigation.

### Actual Behavior

Focus skips the close control entirely and moves directly to the next focusable element.

---

## Impact

These issues significantly reduce the usability of Overview mode for keyboard-only users by:

* Requiring navigation through numerous invisible elements before reaching the Overview controls. (Confuses user)
* Making desktop removal inaccessible without a pointing device - as 'X' isn't reachable.
* Creating inconsistent keyboard behavior within a modal-like interface.
* Introducing accessibility barriers that are inconsistent with expected focus management and keyboard interaction patterns.

KarunyaChavan:
This PR fixes two a11y issues that make the Overview mode keyboard-inaccessible: Tab traverses the entire hidden admin chrome before reaching desktop tiles, and the close X button on each tile is unreachable via keyboard.

Closes #331 

## Why? 
Two root causes: 

- no inert is applied to background elements when Overview opens.
- the close X is nested inside the tile <button>, which browsers skip in Tab order.

## How? 

- On overview enter, inert is set on the admin menu, docks, widgets, footer, and all #wpbody-content children  and removed on exit. 
- The close X is extracted from inside the tile `<button>` into a sibling `<button>` under a wrapper `<div>`, so both are independently reachable by Tab. 
- The global Enter handler now delegates to native <button> clicks instead of intercepting them, fixing a third issue where pressing Enter on the close button would exit overview instead of closing the desktop. 
- The inert-toggling logic is deduplicated into a shared constant and helper.


AllTerrainDeveloper (changes requested):
Hey! Thank you again! I can see a regression added by this pull request. If we have 3 opened windows and one of them is minimized, if we click on the minimized one, it's not restored.

See video:
https://github.com/user-attachments/assets/ef337ca2-2dbb-42c8-b0c4-411d33f77307

A second thing is that we have a rule which is to add tests with each new change :)

Would you mind?

KarunyaChavan:
> I can see a regression added by this pull request. If we have 3 opened windows and one of them is minimized, if we click on the minimized one, it's not restored.

@AllTerrainDeveloper This is actually a core bug - you can reproduce it on the latest trunk as well. I already discussed this in #311, and I've opened PR #312 to address it.

AllTerrainDeveloper:
Oh Thank you! Didn't realise! too much work lately on Desktop Mode 🤣

AllTerrainDeveloper (approved):

PR #343, opened 2026-07-13
fix: live-refresh the dock when a post type is added off the plugins page

### LINKED ISSUE #325: New post types not visible until refresh entire browser window
After adding a new custom post type, it is not accessible (not present in the side menu). Only after refreshing the browser window does it appear.

```
Plugin version:    0.9.2
WordPress version: unknown
User agent:        Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/148.0.0.0 Safari/537.36
Viewport:          2418x1393
Platform:          MacIntel
Current URL:       https://advancetrac.infinitemonkeys.ca/wp-admin/plugins.php?plugin_status=all&paged=1&s&desktop_mode_portal=1&desktop_mode_portal_intent=1
```

</details>

itzmekhokan:
## Summary

Fixes #325.

A newly-registered custom post type didn't show up in the dock / side menu until the user reloaded the entire browser window.

**Root cause.** The chromeless bridge only emits the full menu-refresh payload from four pages — `plugins.php`, `plugin-install.php`, `update.php`, `themes.php` — plus the explicit `wp.desktop.refreshMenu()` probe. That covers plugin activation, but a CPT registered through a settings-based tool (CPT UI, Pods, ACF, …) is saved on its own `admin.php?page=…` / `options.php` screen, none of which is on that allowlist. So the shell never received a fresh menu payload and the new top-level menu stayed invisible until a full reload rebuilt the shell from `$menu`.

## Fix — a lightweight menu signature

Rather than emit the full payload on every navigation (wasteful, and the reason the allowlist exists), every chromeless page now ships a cheap fingerprint of the admin menu; the shell only spends a refresh probe when it actually changes.

- **`desktop_mode_menu_signature()`** (`includes/core/payload.php`) — md5 over the capability-passing top-level + submenu slugs and their badge-stripped titles, i.e. exactly the add / remove / rename events the dock cares about. Transient badge counts (update notifications, moderation queues) are stripped so they don't churn the hash.
- **Bridge** (`includes/render/chromeless-bridge.php`) — emits a `desktop-mode-menu-signature` postMessage on every chromeless page that isn't already sending a full payload. The full payload now embeds its own `menuSig` so the shell can adopt it.
- **Boot config** (`includes/render/assets.php`) — seeds `desktopModeConfig.menuSig` so the shell has a baseline from the first paint (no wasted probe on the first window opened).
- **Shell** (`src/boot/menu-refresh.ts`) — tracks the last-known signature and calls the existing `wp.desktop.refreshMenu()` only when a reported signature differs, with an in-flight guard so a burst of navigations can't stack probes.

Net effect: off-allowlist menu changes (CPTs, and any plugin that adds a menu on save) now surface live; unchanged menus cost nothing beyond a tiny message.

## Testing

- `npm run typecheck`, `npm run lint`, `npm run build` — clean
- `npm run test:js` — 1608 tests pass
- PHP: `php -l` clean; the added lines are PHPCS-clean against `phpcs.xml.dist`
- **Manual, in `wp-env`:**
  - Registered a CPT while the shell was running (off the plugins page). After any fresh chromeless navigation the bridge posts `desktop-mode-menu-signature`, the shell detects the change, runs one refresh probe, and the new tile appears in the dock — **no browser reload**.
  - Negative case: loading a chromeless page when the menu is unchanged posts the signature but triggers **no** refresh probe.
  - Regression: plugin activate/deactivate live-refresh still works (full payload path unchanged).

## Docs

Updated `docs/javascript-reference.md` (new `desktop-mode-menu-signature` bridge message + `desktop-mode-plugins-changed` catalog entry, `refreshMenu()` note) and `docs/api-index.md` (postMessage inventory).

### DIFF
--- a/includes/core/payload.php
+++ b/includes/core/payload.php
@@ -1166,7 +1166,89 @@ function desktop_mode_build_menu_payload() {
 		$payload[ $key ] = function_exists( $builder ) ? $builder() : array();
 	}
 
+	// A cheap structural fingerprint of the admin menu the shell uses to
+	// decide whether a live refresh is warranted. Shipped in every full
+	// payload so the shell can seed / update its last-known signature
+	// without recomputing it client-side (which would risk drift from
+	// the server's capability-gated view). See
+	// desktop_mode_menu_signature().
+	$payload['menuSig'] = desktop_mode_menu_signature();
+
 	return $payload;
+}
+
+/**
+ * Cheap structural fingerprint of the current admin menu.
+ *
+ * The chromeless bridge emits the *full* menu payload only from the
+ * handful of pages whose completion commonly mutates the admin menu
+ * (activation / install / theme switch). That leaves a gap: a custom
+ * post type registered through a settings-based tool (CPT UI, Pods,
+ * ACF, …) saves on its own `admin.php?page=…` / `options.php` screen,
+ * none of which is in that list, so the new top-level menu never
+ * reaches the live dock until a full browser reload rebuilds the shell
+ * (GH#325).
+ *
+ * Building the full payload on *every* chromeless page just to catch
+ * that case would be wasteful — most navigations don't touch the menu.
+ * Instead every chromeless page ships this lightweight signature; the
+ * shell compares it against its last-known value and only spends a
+ * `wp.desktop.refreshMenu()` probe when it actually changed.
+ *
+ * The hash covers the capability-passing top-level + submenu slugs and
+ * their (badge-stripped) titles — i.e. exactly the add / remove /
+ * rename events the dock cares about. Transient badge counts (update
+ * notifications, moderation queues) are stripped so they don't churn
+ * the signature; those have their own refresh path.
+ *
+ * @since 0.9.4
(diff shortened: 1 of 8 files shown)


includes/core/payload.php
```diff
@@ -1166,9 +1166,91 @@ static function ( $item ) {
...
+ * their (badge-stripped) titles — i.e. exactly the add / remove /
+ * rename events the dock cares about. Transient badge counts (update
+ * notifications, moderation queues) are stripped so they don't churn
+ * the signature; those have their own refresh path.
+ *
+ * @since 0.9.4
+ *
+ * @return string 32-char md5 fingerprint, or '' when the menu is
+ *                unavailable (non-admin context).
+ */
```
AllTerrainDeveloper, line 1208: This is a great idea. Thank you!

AllTerrainDeveloper (approved):
Excellent!

PR #345, opened 2026-07-14
fix: delete a sticky note on close instead of only hiding it

### LINKED ISSUE #344: Closing a sticky note doesn't delete it — the note reappears on the next desktop render
When you close a desktop sticky note using the × button in its title bar, the note is only torn down from view — the backing `wp_guideline` post is never deleted. Any subsequent desktop re-render (opening a window, right-clicking a desktop icon, or reloading the page) restores the note.

### Steps to reproduce
1. Open the desktop shell (`/wp-admin/index.php?desktop_mode_portal=1`).
2. Right-click the empty desktop wallpaper → **New sticky note**.
3. Type a word or two into the note.
4. Click the **×** (close) button in the note's title bar — the note disappears.
5. Trigger any desktop re-render: open a window (double-click a desktop icon), right-click a desktop icon, or reload the page.

### Expected
Clicking × permanently removes the sticky note.

### Actual
The note disappears momentarily, then **reappears** on the next desktop render / page reload. Reproduced consistently.

### Likely cause
Sticky notes are persisted as `wp_guideline` CPT posts. The title-bar × appears to only remove the note's DOM/window without issuing (or persisting) the REST delete for the post, so the sticky-notes layer re-hydrates it from the server on the next render.

### Validation
Run before and after clicking ×; the row does **not** go away:

`npx wp-env run cli wp post list --post_type=wp_guideline --post_status=any --fields=ID,post_title`

The note only truly disappears after `wp post delete <ID> --force`. In DevTools, `localStorage['desktop-mode-sticky-notes-geometry']` also keeps its `"guideline:<id>"` entry after ×.

### Environment
- Plugin version: 0.9.4
- WordPress version: 7.0.1
- Reproducible from a fresh `wp-env` (`npm run env:start`)

### Screenshot
Image

itzmekhokan:
## Summary

Fixes #344.

The sticky-note title-bar **×** was labelled *"Hide sticky note"* and only called `layer.forget()` — an in-memory eviction. The backing `wp_guideline` post was never touched, so the note **reappeared on the next heartbeat tick, page reload, or any desktop re-render**. Since the × was the only removal affordance, a note could never actually be deleted.

## Fix

Make the × delete the note, following the same **soft-trash + Undo** pattern the desktop-files layer (`trashPlacementWithUndo`) already uses:

- **`rest.ts`** — add `deleteStickyNote()` (REST *trash*, not `force`, so it stays recoverable) and `restoreStickyNote()` (untrash by flipping status back to `private`). `fetchStickyNotes` only queries `status: 'private'`, so a trashed note never re-hydrates onto the desktop.
- **`layer.ts`** — the × now trashes the note with optimistic eviction and an **Undo** toast, and is relabelled *"Delete sticky note"*. A `dismissedIds` guard drops any in-flight heartbeat/reload echo for a just-deleted id so it can't resurrect the note before the trash propagates.
- A note created and closed *inside* the save debounce is persisted first (`save()` now hands callers its in-flight promise) so there's always a server row to trash — closing the create-then-delete race behind the intermittent resurrection.

No behaviour change for the "empty, never-saved note" case (still dropped locally with no round-trip).

## Testing

- `npm run typecheck`, `npm run lint` — clean
- `npm run test:js` — **1610 pass** (adds 2 tests: the × trashes + evicts the note, and a stale heartbeat echo can't re-add a deleted note)
- **Manual, in `wp-env`:**
  - Clicking × deletes the note and shows a *"…deleted"* toast with **Undo**; the `wp_guideline` post moves to `trash`.
  - **The note no longer returns after a full page reload** (the #344 repro) — verified `0` notes and the post staying `trash`.
  - Undo / `restoreStickyNote` flips the post back to `private` and re-adds the note.


AllTerrainDeveloper:
Hey Thanks for your contribution! I'll ask the person who made the Sticky Notes, not sure if he really intended to hide for the session or close them forever. @artpi?

AllTerrainDeveloper:
I'll reach Artur internally 😆

mmtr:
Closing this one out as per https://github.com/WordPress/openstation/issues/344#issuecomment-5192164720.

PR #349, opened 2026-07-14
Fix selection identity & staleness bugs that let destructive actions hit the wrong items

AllTerrainDeveloper:
## Context

A user selected a comment in the native Comments app, hit Trash, and a *different* comment was trashed. The immediate comments bug was fixed in #293 (clear selection after bulk moderation), but an audit of every native app found the same bug class alive in three forms — including a critical one in the Recycle Bin. This PR fixes all of them in one pass.

### 1. Recycle Bin — cross-type numeric-id collision (critical)

`src/recycle-bin/index.ts` keyed rows by bare `row.id`, but the bin lists posts/pages/attachments (wp_posts ids) alongside comments (wp_comments ids) — so **post #5 and comment #5 shared one selection key**. Ticking the post also selected the comment, and bulk "Delete forever" permanently purged both. Realtime refreshes could even slot a colliding item *under* an existing selection.

Row identity is now the composite `type:id` throughout: `getRowId`, `collectSelectedItems()` (which builds the `[{id, type}]` REST payload), and `itemsFingerprint()`. The type filter and search also clear the selection when they replace the view.

### 2. `<wpd-table>` — select-all swept rows hidden by client-side filters

`selectAll()` iterated the full `data` buffer, ignoring live column filters. In the Recycle Bin (Title / "By" column filters): filter 200 rows down to 2, click select-all intending "these 2", Delete forever → all 200 purged, 198 of them invisible.

`selectAll()` and the header checkbox tri-state now operate on the **visible (filtered) row set** — "checked" means "every row the user can see is selected". Tables without client-side filters are unaffected.

### 3. Selection hygiene — stale ids riding into bulk actions

`<wpd-table>` deliberately keeps its selection across `data` reassignment, so ids selected on a previous page / search / filter lingered invisibly and were swept into the next bulk action:

- **Comments** (this is the remaining path for the originally-reported bug on post-#293 builds): selection now clears on every panel data replacement — pagination, search, per-page, manual refresh, heartbeat reload, tab re-activation.
- **Posts & Pages** (same code, both modes): cleared on status/search/pagination/per-page/sort/author-tag-filter changes. Deliberately *not* cleared inside `refresh()` itself, so background broadcast-driven reloads don't wipe an in-progress selection.
- **Users**: cleared on status/search/pagination/per-page changes *and* after the bulk role apply (previously never cleared anywhere).
- **Plugins**: cleared on status-filter/search changes — a plugin ticked under one filter could ride hidden into bulk **Delete**.

## Audited and confirmed safe (no changes needed)

My WordPress (per-tile id closures, id-keyed removal), media grid, desktop files, sticky notes, tag cloud / category mindmap term actions, user-edit window, plugins per-row actions + update queue, and all REST layers (no index-based request/response mapping anywhere).

## Tests

- `wpd-table.test.ts`: `selectAll()` with an active filter selects only visible rows; header tri-state tracks visible rows when a selected row is filtered out.
- New `tests/vitest/recycle-bin-selection-identity.test.ts`: mounts the real `renderRecycleBin()` + `<wpd-table>` in jsdom with mocked REST; asserts post #5 / comment #5 get distinct `type:id` row keys, selecting one does not select the other, and bulk purge sends exactly `[{id: 5, type: 'post'}]`.
- Full suite: 193 files / 1816 tests green; lint + typecheck + build clean.

## Docs

`docs/examples/data-table.md`: documented the visible-rows semantics of `selectAll()`, the "ids must be unique across mixed row kinds" rule, and the clear-selection-on-query-change pattern for destructive consumers (Selection section, methods table, Common pitfalls).

## Manual test plan (on :8889)

1. Trash a post and a comment that share the same numeric id (or just any post + any comment). Open Recycle Bin → tick only the post → only that row shows checked → Delete forever → confirm dialog says 1 item → only the post is purged.
2. Recycle Bin: type in the Title column filter, click the header select-all → only the visible rows select.
3. Comments: select a comment on page 1 → Next page → bulk bar is hidden, nothing pre-selected → select one → Trash → only that one is trashed.
4. Posts/Pages/Users/Plugins: select rows, then change search/filter/page → selection resets.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22landingPage%22%3A%22%2Fdesktop-mode%2F%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fdesktop-mode%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-349-1184c6beeadd6dd80bb187a684a5c80377f33e24.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/docs/examples/data-table.md
+++ b/docs/examples/data-table.md
@@ -334,7 +334,8 @@ If editing is the primary use case, request a first-class `column.editor` API �
 | `select(id)` / `deselect(id)` | Mutate the selection by id. |
 | `selectAll()` / `clearSelection()` | Bulk operations (multi-mode). `selectAll()` selects only the rows passing the active client-side filters. |
 | `selection` | Get/set the selection set. |
-| `selectedRows` | Resolved row objects matching `selection`. |
+| `selectedRows` | Resolved row objects matching `selection` (resolves against the full `data` buffer). |
+| `visibleRows` | Rows passing the active client-side filters — the set `selectAll()` operates on. Destructive bulk consumers should resolve `selection` against this, not `data`. |
 | `getRowId` | Stable-id extractor (default: index). |
 | `sort` | Get/set the active `{ key, direction }` (or `null`). |
 | `filters` | Get/set the filter map. |
@@ -395,7 +396,7 @@ If editing is the primary use case, request a first-class `column.editor` API �
 - **Editable cells losing focus.** Reassigning `table.data` on every keystroke triggers a full body repaint — the new `<input>` is a different element, so focus is lost. Commit on blur/Enter, or hold edits in a side buffer until save.
 - **Selection going stale on data refresh.** Without `getRowId`, ids are array indices — selections drift if rows reorder. Set `getRowId = ( row ) => row.id` whenever rows have a natural identifier.
 - **Colliding ids across mixed row kinds.** Two rows whose `getRowId` returns the same value are one row as far as selection is concerned — ticking one ticks both. Qualify the id (``( row ) => `${ row.type }:${ row.id }` ``) when a list mixes entities from independent id sequences.
-- **Stale selection feeding destructive bulk actions.** Selection survives `data` reassignment by design. If your bulk actions read `table.selection` (trash, delete…), call `clearSelection()` whenever the query (page, search, filter) changes, or resolve the selection against the currently rendered rows before acting.
+- **Stale selection feeding destructive bulk actions.** Selection survives `data` reassignment by design. If your bulk actions read `table.selection` (trash, delete…), call `clearSelection()` whenever the query (page, search, filter) changes, **and** resolve the selection against `table.visibleRows` (not `data`) before acting — a data-driven change can hide a selected row without any filter event firing, and rows the user can't see must never be swept into a destructive action.
 
 ## See also
 
--- a/docs/examples/data-table.md
+++ b/docs/examples/data-table.md
@@ -135,12 +135,18 @@ table.addEventListener( 'wpd-table-selection-change', ( e ) => {
 // Programmatic API
 table.select( 'alice@a.com' );
 table.deselect( 'alice@a.com' );
-table.selectAll();                          // multi only
+table.selectAll();                          // multi only — selects the VISIBLE rows
 table.clearSelection();
 table.selection = new Set( savedIds );      // bulk replace
 ```
 
+`selectAll()` (and the header select-all checkbox) selects the rows passing the active client-side filters — never rows a filter is currently hiding. The header checkbox tri-state follows the same rule: "checked" means every *visible* row is selected. Tables without client-side filters are unaffected.
+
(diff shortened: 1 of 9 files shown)


AllTerrainDeveloper:
## Adversarial verification pass (2 independent reviews of the branch diff)

Two fresh-eyes adversarial reviews were run against the branch with instructions to falsify each fix. Results:

**Confirmed holding (could not be falsified):** composite `type:id` identity end-to-end (client keys ↔ server `{id,type}` dispatch); `type` slugs are `sanitize_key`-clean so the composite can't be ambiguous; all selection-clear points; the refresh `seq` guard closes overlapping-refresh races (the prune block contains no `await`); the posts trash action independently guards through `table.data` + skips already-trashed rows, so even the deliberate "don't clear on background broadcast refresh" trade-off cannot misfire destructively; the plugins TDZ concern is unreachable (elements connect after `const table` executes); comments undo acts on stored ids, immune to query changes; the users `targetIds` shadowing is complete.

**Falsified and fixed in the follow-up commits (`1b0c40c`, `1184c6b`):**
1. Pin-to-Desktop matched restore results by bare numeric id — post #5 + comment #5 selected with one restore failing still pinned the failed one. Now restores per-ref (unambiguous success signal; same server work).
2. A *data-driven* visibility change (realtime refresh replacing a selected row so it stops matching an active column filter) hid a row without firing `wpd-table-filter-change`, leaving it purgeable while invisible. Bulk collection + selection pruning now resolve against the new public `table.visibleRows` (rows passing active client-side filters).
3. Recycle Bin column-filter edits (Title / "By") didn't clear selection; the Empty-bin confirm claimed "current view" while the endpoint purges the entire bin — copy corrected.
4. Comments: the post-bulk clear had become dependent on the follow-up fetch succeeding — now clears synchronously before the refresh. Keyboard j/k now emits selection-change so the "N selected" chip tracks the cursor `a`/`s`/`d` act on.
5. Plugins: background reloads left bulk buttons holding stale row snapshots — `paintTable()` now rebuilds the bulk bar from live selection + fresh rows (delete lists always reflect current status).

**Known-and-accepted (documented, server-guarded, non-destructive):** posts/users selections deliberately survive *background* refreshes (clearing would wipe an in-progress selection); a phantom id surviving that path is filtered out by the trash guard / 404s server-side. `selectedRows` and the `wpd-table-selection-change` `rows` detail still resolve against the full buffer by contract — flagged in docs with `visibleRows` as the destructive-consumer alternative.

Gates after each commit: eslint, tsc, 1821 vitest tests, full Vite build — all green.

PR #366, opened 2026-07-17
Normalize AI Copilot tool schemas for the provider

### LINKED ISSUE #362: AI Copilot: one third-party ability with a valid-but-unsupported input_schema 400s the entire assistant
# AI Copilot: one third-party ability with a valid-but-unsupported `input_schema` 400s the entire assistant

**Version:** 0.9.5

## Summary

Since 0.9.4, the Copilot's tools are WordPress Abilities and **every** read-only ability on the site is offered automatically — "No opt-in: register a read-only ability and the assistant can use it" (`includes/ai-copilot/abilities.php:51-55`).

The converter passes the ability's `input_schema` through **unmodified** as the tool's `parameters` (`includes/ai-copilot/search.php:743-751`). Several JSON Schema constructs that are perfectly valid — and that the Abilities API accepts — are **not** supported inside Anthropic's `input_schema`. When any one of them is present, the provider rejects the request and **Ask AI dies completely**: it isn't the offending tool that's dropped, it's the whole assistant.

There's no opt-out. A plugin author cannot prevent their ability being offered, so a schema shape they had every reason to consider correct silently breaks a feature they may not even know exists.

## Impact

Ask AI is **100% unusable** on the affected site. Not degraded — every query returns a 400 before the model runs. Any plugin registering a read-only ability with one of the shapes below takes the assistant down for everyone on that install.

## Reproduction

```php
add_action( 'wp_abilities_api_init', function () {
    wp_register_ability( 'demo/thing', array(
        'label'               => 'Demo',
        'description'         => 'Demo.',
        'permission_callback' => '__return_true',
        'execute_callback'    => fn() => array( 'ok' => true ),
        'input_schema'        => array(
            'type'       => 'object',
            'properties' => array(),   // ← empty PHP array → JSON []
        ),
        'meta' => array( 'annotations' => array( 'readonly' => true ) ),
    ) );
} );
```

Open Ask AI → ask anything → `Bad Request (400)`.

## The three shapes that break it

All three are valid JSON Schema. All three are accepted by the Abilities API. All three 400 the Copilot.

| # | `input_schema` contains | Provider error |
|---|---|---|
| 1 | `'type' => array( 'object', 'null' )` | `tools.N.custom.input_schema.type: Input should be 'object'` |
| 2 | top-level `anyOf` / `oneOf` / `allOf` | `tools.N.custom.input_schema: does not support oneOf, allOf, or anyOf at the top level` |
| 3 | `'properties' => array()` (empty PHP array → JSON `[]`) | `tools.N.custom.input_schema.properties: Input should be an object` |

These aren't contrived. In our plugin + theme they're all load-bearing:

- **(1)** `['object','null']` lets an ability be invoked with `null` input — its GET/no-args run-path. 21 of our abilities use it.
- **(2)** `anyOf` expresses a real constraint: *"supply `post_id` **or** `slug`"*.
- **(3)** An ability that takes no arguments naturally declares `'properties' => array()` — PHP has no distinct empty-map literal, so it encodes to `[]` instead of `{}`.

## Root cause

`includes/ai-copilot/search.php:743-751`

```php
$input_schema    = $ability->get_input_schema();
$builtin_tools[] = array(
    'type'        => 'function',
    'name'        => $tool_name,
    'description' => (string) $ability->get_description(),
    'parameters'  => ! empty( $input_schema )
        ? $input_schema                                              // ← raw, unvalidated
        : array( 'type' => 'object', 'properties' => (object) array() ),
);
```

Note the fallback on the last line: it **already** casts empty properties to `(object) array()` — the code plainly knows an empty PHP array serialises to `[]` and that the provider needs `{}`. That knowledge just never reaches the branch above it, where a plugin's schema goes through untouched.

Desktop Mode's own abilities all declare a plain `'type' => 'object'` with non-empty properties, which is exactly why this never shows up in-house — **only third-party abilities trigger it.**

## Why this is unusually hard to diagnose

- The error names a **tool index** (`tools.30`), not an ability name, a namespace, or a plugin. There is no way to map index → culprit from the error alone.
- Fixing one shape reveals the next, and the index **moves**: we saw `tools.12` → `tools.29` → `tools.30` across three fixes. Nothing tells you how many remain.
- Nothing in the message mentions *abilities*, *Desktop Mode*, or *Copilot*. A site owner sees only `Bad Request (400)` in the Ask AI box.
- The plugin author has no signal at all. Their ability works fine over REST and MCP; they never learn they broke a feature in a different plugin.

## Suggested fix

Normalise in the converter, immediately before the tool definition is built. Roughly:

```php
$schema = $ability->get_input_schema();
if ( ! is_array( $schema ) || empty( $schema ) ) {
    $schema = array( 'type' => 'object', 'properties' => (object) array() );
} else {
    // The provider's tool schema is a strict subset of JSON Schema.
    $schema['type'] = 'object';                                    // no unions
    unset( $schema['oneOf'], $schema['allOf'], $schema['anyOf'] ); // not at top level
    if ( isset( $schema['properties'] ) && array() === $schema['properties'] ) {
        $schema['properties'] = (object) array();                  // {} not []
    }
}
```

This loses nothing. The converter projects an ability into a **tool schema**; it does not touch the ability. `WP_Ability::execute()` still validates against the real schema, and `permission_callback` still gates it — so a dropped `anyOf` is still enforced server-side, and the model is told about it in the ability's description. Only combinators at the **top level** need removing; nested ones are fine.

Two further suggestions:

1. **Fail soft per-tool.** If a schema can't be made conformant, skip that one tool and log it — one bad ability shouldn't disable the assistant for the whole site.
2. **Name the culprit.** Mapping the provider's `tools.N` back to the ability name in the surfaced error would turn an undiagnosable 400 into a one-line bug report.

## Notes

- Same class of mismatch applies to MCP hosts, which forward tool schemas to the same validator — so normalising here helps beyond Anthropic.
- Happy to open a PR if the approach looks right.

Environment
```
Desktop Mode: 0.9.5
Third-party abilities: 28 read-only (18 plugin + 10 theme), all
meta.annotations.readonly, all offered automatically to the Copilot.
```

juanlentino:
Fixes #362.

Thanks @AllTerrainDeveloper for confirming the approach on the issue. This is the normalization fix; the fail-soft-per-tool and name-the-culprit ideas are left as the follow-ups you flagged.

## The problem

`desktop_mode_ai_run_search()` advertises each built-in tool with its ability's `input_schema` passed straight through as the tool `parameters` (`includes/ai-copilot/search.php`). The Abilities API accepts the full breadth of JSON Schema, but the provider's tool-schema validator does not — and it rejects the **whole** request, not just the offending tool. So a single tool with a legal-but-unsupported schema makes Ask AI return a 400 for **every** query: unusable, not degraded, and the error names a `tools.N` index rather than the plugin, so it's painful to trace.

Three shapes, all valid JSON Schema, trigger it in the wild:

1. **`type` as an array** — e.g. `['object','null']`, an ability's GET/null run-path. The provider wants the literal `"object"` at the top level.
2. **A top-level `oneOf` / `anyOf` / `allOf`** — e.g. "supply `post_id` OR `slug`". Rejected with *"does not support oneOf, allOf, or anyOf at the top level"*.
3. **An empty `properties`** — a no-args tool's `[]`, where an object schema needs `{}`.

## The fix

A small pure helper, `desktop_mode_ai_normalize_tool_schema()`, projects a schema onto the provider-supported subset:

- forces top-level `type` to `"object"`
- strips **top-level** `oneOf` / `anyOf` / `allOf` (nested combinators are left intact — those are real constraints the provider accepts)
- coerces an empty `properties` array to an object
- empty / non-array input becomes `{"type":"object","properties":{}}`

It's applied to every tool in the final list, **after** the `desktop_mode_ai_tools` filter — so it covers the complete set the provider receives: built-in abilities, command tools, and anything a plugin injected. No single tool, from any source, can 400 the request.

**Nothing loses enforcement.** This reshapes only the copy advertised to the model. `WP_Ability::execute()` still validates arguments against the real schema and `permission_callback` still gates execution — the model is simply told the constraint in prose (the tool description) instead of a schema construct the provider can't parse. The projection is idempotent, so a plugin that already normalizes on the filter is unaffected.

## Tests

`tests/phpunit/tests/aiToolSchemaNormalization.php` covers each of the three shapes, nested-combinator preservation, empty/non-array input, and idempotence. The helper is pure, so these run without the network or the ability registry.

## Notes

- **Fail-soft-per-tool** (drop the offending tool and name it, rather than 400-ing) and **naming the culprit** are the resilience follow-ups you called out — happy to open a separate PR for those; they're a different concern from this correctness fix.
- I tagged `@since 0.9.6` as a guess against the current `0.9.5` release — please adjust to your intended version.
- Verified the helper's behaviour directly and confirmed `phpcs` reports no new violations on the change. I couldn't run the full PHPUnit suite locally (needs the wp-env test DB), so I'd lean on CI for the `test:php` leg.

---

*Prepared by @juanlentino with Claude Code assistance.*

### DIFF
--- a/includes/ai-copilot/search.php
+++ b/includes/ai-copilot/search.php
@@ -609,7 +609,9 @@ function desktop_mode_ai_normalize_tool_schema( $schema ) {
 	unset( $schema['oneOf'], $schema['allOf'], $schema['anyOf'] );
 
 	// An empty PHP array encodes as `[]`; an object schema's properties need `{}`.
-	if ( isset( $schema['properties'] ) && array() === $schema['properties'] ) {
+	// A schema with no `properties` at all (e.g. one whose only content was a
+	// stripped top-level combinator) gets an empty object for the same reason.
+	if ( ! isset( $schema['properties'] ) || array() === $schema['properties'] ) {
 		$schema['properties'] = (object) array();
 	}
 
--- a/includes/ai-copilot/search.php
+++ b/includes/ai-copilot/search.php
@@ -564,6 +564,59 @@ function desktop_mode_ai_search_resumable_tools() {
 }
 
 /**
+ * Projects a tool's `parameters` schema onto the provider-supported subset.
+ *
+ * The Abilities API (and plugin-supplied tools) may use the full breadth of JSON
+ * Schema, but the provider's tool-schema validator does not — and it rejects the
+ * whole request, not just the offending tool, so ONE tool with a
+ * legal-but-unsupported schema makes the entire assistant return a 400 before the
+ * model runs. Three shapes that are valid JSON Schema, but rejected here, have
+ * been seen in the wild (see the linked issue):
+ *
+ *   1. `type` as an array, e.g. ['object','null'] — an ability's GET/null
+ *      run-path. The provider wants the literal string "object" at the top level.
+ *   2. A top-level `oneOf` / `anyOf` / `allOf`, e.g. "post_id OR slug". Rejected
+ *      with "does not support oneOf, allOf, or anyOf at the top level".
+ *   3. `properties` as an empty PHP array, which encodes to JSON as `[]` where an
+ *      object schema needs `{}`.
+ *
+ * This reshapes only the copy advertised to the model. The ability itself is
+ * untouched: `WP_Ability::execute()` still validates arguments against the real
+ * schema and `permission_callback` still gates execution, so nothing loses
+ * enforcement — the model is simply told the constraint in prose (the tool
+ * description) instead of in a schema construct the provider can't parse. Only
+ * the TOP level is constrained; a combinator nested inside a property is a real
+ * constraint the provider accepts, so it is left intact.
+ *
+ * @since 0.9.6
+ *
+ * @param mixed $schema A tool parameters schema (array), or empty/non-array.
+ * @return array A provider-safe object schema for a tool `parameters` block.
+ */
+function desktop_mode_ai_normalize_tool_schema( $schema ) {
+	if ( ! is_array( $schema ) || empty( $schema ) ) {
+		return array(
+			'type'       => 'object',
+			'properties' => (object) array(),
+		);
+	}
+
+	// Top-level tool parameters must be the literal "object", never a union.
+	$schema['type'] = 'object';
+
+	// Strip top-level combinators — the provider rejects them outright, and one
+	// such tool 400s the whole request. Nested combinators are left alone.
+	unset( $schema['oneOf'], $schema['allOf'], $schema['anyOf'] );
+
+	// An empty PHP array encodes as `[]`; an object schema's properties need `{}`.
+	if ( isset( $schema['properties'] ) && array() === $schema['properties'] ) {
+		$schema['properties'] = (object) array();
+	}
+
+	return $schema;
+}
+
+/**
  * Runs the agentic content-search loop.
  *
  * The model receives focused tools — search_posts, search_pages,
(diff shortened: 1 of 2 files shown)


AllTerrainDeveloper:
Reviewed this locally, the projection had one remaining gap: a schema whose only content is a top-level combinator (e.g. `{ oneOf: [ { properties: … }, { properties: … } ] }` with nothing duplicated at the top level), or a bare `{ "type": ["object","null"] }`, normalized to `{"type":"object"}` with no `properties` key at all, still an incomplete object schema for the same strict validator this PR targets, and the model would see a tool with zero argument info.

AllTerrainDeveloper:

PR #375, opened 2026-07-20
Redesign session-expiry recovery: one login prompt, in-place recovery

epeicher:
## What it does

Redesigns what happens when the WordPress login session expires while the desktop is open:

- **One login prompt instead of N.** Chromeless iframes no longer load core's `wp-auth-check` modal — the parent shell owns a single "session expired" prompt over the whole desktop.
- **In-place recovery instead of a full shell reload.** After re-login, cached nonces are refreshed live via the heartbeat, window iframes reload in place, and the shell itself never reboots — windows, positions, and wallpaper survive. Recovery completes ~1 s after the login form is submitted.
- **No more global `fetch`/XHR monkey-patching.** DevTools request initiators point at the real callers again instead of `desktop-mode-parent-auth-recovery.js:68`, and a permission `403` from a live session can never pop the login prompt.
- New public events for plugins: `desktop-mode-auth-lost` / `desktop-mode-auth-restored` document CustomEvents plus `HOOKS.AUTH_LOST` / `HOOKS.AUTH_RESTORED`.

|Before no login popup | After: It displays login popup| It restores the windows
|-------|------|------
|CleanShot 2026-07-20 at 12 07 14@2x |  CleanShot 2026-07-20 at 12 05 17@2x| CleanShot 2026-07-20 at 12 05 38@2x

## Rationale

The old inline script (`shell.php`, since 0.8.3) predates the heartbeat nonce-refresh system (0.8.7) and was built on the assumption that stale nonces could only be fixed by hard-reloading the shell. It also wrapped `window.fetch` + `XMLHttpRequest` globally — every shell request appeared to originate from the recovery script — and treated *any* same-origin 401/403 (including plain `rest_forbidden` permission denials) as a session signal. Meanwhile every open window ran its own copy of core's login modal, and windows that missed the re-auth stayed stale.

## Implementation

The recovery state machine lives in `src/auth-recovery/index.ts` and keys off authoritative Heartbeat state only:

- **Loss**: the `wp-auth-check` response field (attached server-side by core on every tick) flips to `false`.
- **Recovery**: the flag flips back to `true`, an iframe bridge posts `desktop-mode-reauth-detected` (its heartbeat noticed first), or a `nonces_expired` response arrives during a known outage — core only ever sends that field to an authenticated session, making it the earliest possible re-auth signal.
- **Recover**: force a functional tick (fresh nonces ride it), reload every same-origin iframe except core's mid-handshake login iframe, fire the events. Racing signals (own tick + one message per open window) collapse into one run via a cooldown.

Supporting changes:

- `includes/core/routing.php` — `desktop_mode_chromeless_suppress_auth_check()` returns `false` from `wp_auth_check_load` on chromeless requests.
- `includes/nonce-refresh.php` — the `desktop_mode_nonces` map now also rides core's `wp_refresh_nonces` filter, so the tick that says "your nonces expired" already carries the replacements (one round-trip heal). Both paths attach `desktop_mode_auth: { uid }`.
- `src/boot/tracked-fetch.ts` — 401/403 responses through `wp.desktop.fetch` call `noteAuthFailure()`, which only *accelerates* the next heartbeat (debounced); it never decides anything.
- **User switch** is the one remaining hard reload: if the interim login authenticated a different account (`uid` mismatch), in-place recovery would leave user A's desktop issuing user B's requests, so the shell reloads.

MutationObserver on `#wp-auth-check-wrap` covers the modal-hide path (core only forces a follow-up tick on the post editor), and the tick debounce is trailing-edge — deferred requests fire when the cooldown lapses instead of stranding the two-tick recovery chain on the 60 s schedule.

Docs updated: `javascript-reference` (new *Session expiry & recovery* section), `bridge-protocol` (documents `desktop-mode-reauth-detected`, which had never been documented), `api-index`, `hooks-reference`, `event-driven-framework`.

## Testing instructions

```bash
npm run test:js    # includes 15 new tests in tests/vitest/auth-recovery.test.ts
npm run test:php   # includes 7 new tests (auth-check suppression, wp_refresh_nonces rider)
```

Manual, on a dev site with a couple of windows open:

1. In the shell console: `window.MARKER = 1`, then destroy your sessions (`wp user session destroy <user> --all`).
2. Trigger any request (or wait a tick): **one** prompt appears; windows behind it show no modals of their own.
3. Log back in: the prompt closes, window iframes reload, REST works immediately, and `window.MARKER` is still set — the shell never reloaded.


AllTerrainDeveloper (approved):
It works as expected!

PR #381, opened 2026-07-21
Add real file/folder storage on the desktop

epeicher:
## What it does

Closes the "shortcuts only" gap in Files-on-the-Desktop: users can now upload arbitrary files or whole folder trees onto the desktop and real bytes land on the server, tied to that user. Files download back unmodified; folders download as an on-demand `.zip`. Owners can share an uploaded file with specific users, who get read + download only (no move, no rename, no delete, stricter than the folder-sharing write tier).

Drop-destination defaults follow the surface: the desk or a folder means Desktop storage (a folder drop lands *inside* that folder), WordPress admin windows (Media, Posts, Pages) keep the Media Library flow. The upload dialog carries a selector either way.

https://github.com/user-attachments/assets/ea60ab12-3451-4329-aed9-3f8fabf7727c

### Screenshot of the new status bar at the bottom for folders
CleanShot 2026-07-22 at 12 28 09@2x

## Rationale

Desktop tiles were references to WP entities, so there was no way to bring actual files into the desktop. The design (per the plan in `docs/plans/`) reuses everything already load-bearing instead of inventing a parallel system: hierarchy is the existing folders + placements tables, sync is the existing heartbeat delta, sharing is the shares table's `target_type` column that shipped for exactly this. Bytes on disk are a dumb flat blob store.

## Implementation

**Storage.** New `{prefix}desktop_mode_stored_files` table (schema v13). Bytes live flat at `uploads/desktop-mode-files/<user_id>/<uuid>` with server-generated, extensionless names, so user input never composes a disk path and a direct hit cannot dispatch to a PHP handler. Defense in depth: dual-syntax `.htaccess` + `index.php`, downloads only via authenticated endpoints with `Content-Disposition: attachment` + `nosniff`, and a documented nginx `deny all` snippet.

**Intake.** `POST /desktop-mode/v1/files/uploads`, one file per request through `wp_handle_upload()` with a scoped `upload_dir` redirect. Validation is the user-scoped WP MIME policy plus a hard executable denylist matched against every dot-segment (`shell.php.gif` is rejected). Folder trees send `relativePath` per file; the server resolves segments to folder rows mkdir-p style, deduped by parent + name, so parallel uploads cannot race folder creation. Oversize bodies map to explicit 413s, including the empty-`$_FILES`-with-`CONTENT_LENGTH` case PHP produces past `post_max_size`. Receive and register are separate internal steps, kept as the seam for a future resumable layer.

**Downloads.** `GET .../uploads/<id>/download` and `GET .../folders/<id>/download` authenticate via cookie + `_wpnonce` query param (URLs minted at click time) and stream through a `rest_pre_serve_request` short-circuit. Zips build with `ZipArchive` into a swept temp file (feature-gated on `class_exists`, capped via filter, UTF-8 entry names, case-insensitive dedupe, empty dirs preserved). Not-found and no-access both answer 404.

**Ownership + sharing.** Upload placements are owner-locked at the store level: folder write-collaborators cannot move/rename/trash them (`desktop_mode_files_upload_owner_locked`), and `canTrash` shaping hides the affordances. Single-file shares reuse the shares table with `target_type='file'`, capability hard-forced to `read`, invites riding the existing heartbeat `shares.pending` channel with `targetType: 'file'` shapes. Every folder-share query gained an explicit `target_type='folder'` predicate; without it a file share whose id collided with a folder id would leak into folder-share reads (covered by a regression test).

**Lifecycle.** Purging the owner's last placement deletes bytes, row, shares, and recipient placements (with tombstones so tiles scrub live). This is the documented exception to the "references, not copies" contract in `docs/files-on-desktop.md`. A daily sweep reconciles disk/DB drift in both directions; `deleted_user` purges that user's storage.

**Client.** Folder-tree drag-drop uses `webkitGetAsEntry` traversal with the looped `readEntries` 100-entry batching; the desktop sink fires the same `desktop-mode.drop.*` hook chain as the Media path so the progress HUD and subscribers work unchanged. Drop classification resolves folder windows via their `wp-window-<windowId>` element ids. The previous marker (`data-window-id`) never existed on window roots, which also made drops below a folder's tiles land on the desk; dropping onto a closed folder tile now routes into that folder too.

## Testing instructions

```bash
npm run env:start && npm run test:php   # 1269 tests, 54 new across 4 suites
npm run test:js                          # 2083 tests, 4 new files
npm run lint && npm run typecheck && npm run build
```

Manual pass on wp-env: drag a file onto the wallpaper (Desktop default, tile at drop point), drag a nested folder in (tree recreated, empty dirs included), drop a file anywhere on an open folder window and on a closed folder tile (lands inside the folder), right-click for Download and Download as .zip, share a file with a second user and verify the invite banner, read-only enforcement, and that revoke scrubs the recipient tile. Probe a stored file URL directly while logged out and confirm a 404.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22landingPage%22%3A%22%2Fdesktop-mode%2F%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fdesktop-mode%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-381-55e3792d5c40fcdcfa1bdb180177202cb189a822.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
@epeicher it works nicely! I have some suggestions tho. 

If we drag just file(s) let's default to upload to media. That means, imagine I am dragging from my MacOS desktop, 3 different files: a jpg, a png and a mp4. If all of them matches the filter that media library supports, lets default to media please. For other cases, default to "desktop".

image

I found another issue (this one is existing one, but worths fixing it in this PR? Or if you prefer we can narrow this PR to exact functionality and work later on it.

The issue is that if I drag a file and I don't start uploading it yet, and i drag another, the window should update rather than have 2 different upload windows :) 

image

epeicher:
Thanks for the review and the testing @AllTerrainDeveloper! I will work on those findings

AllTerrainDeveloper (approved):

PR #382, opened 2026-07-21
Feat: Allow selecting window from Overview view

### LINKED ISSUE #380: Allow selecting window from Overview view
This seems like a golden missed opportunity: when viewing all open windows with Arrange windows → Overview, allow clicking the open window to select it and make it active (move to the top of the stack). Currently, no matter which window is clicked the overview closes with the windows remaining in the same order.

```
Plugin version:    0.9.5
WordPress version: unknown
Viewport:          2408x1393
Platform:          MacIntel
Current URL:       https://ourtowns.infinitemonkeys.ca/wp-admin/edit.php?post_type=page&desktop_mode_portal=1&desktop_mode_portal_intent=1
```

</details>

CookieDarb:
## What?
Closes #380

Allows users to select and focus a specific window by clicking its thumbnail in Overview mode, bringing it to the top of the stack without forcing it to maximize.

## Why?
Previously, when viewing all open windows via "Arrange windows → Overview", clicking a window's thumbnail would close the overview but leave the windows in their original order. Allowing to be able to use Overview as a window switcher to bring a specific background window to the front can be useful. Additionally, the underlying mechanism was hardcoded to force the selected window into a maximized state upon exiting, which may disrupt user workflow.

## How?
- `src/window-manager/overview.ts`: 
  - Introduced `inertWindowChildren` to apply the `inert` attribute to the *children* of window elements rather than the root wrappers. This safely traps keyboard focus within the overview while keeping the window roots interactive for pointer events (clicks and hover states).
  - Updated the `exitOverview` call within the pointer up handler to omit the `maximize: true` argument, allowing the window to simply focus and restore to its original geometry.
  - Updated `HOOKS.OVERVIEW_EXITING` and `HOOKS.OVERVIEW_EXITED` event payloads to correctly report `reason: 'select'` and emit the `windowId` whenever a window is selected, regardless of whether it is maximizing.
- `tests/vitest/desktops.test.ts`: Added a new test block (`'clicking a window thumbnail in overview selects and focuses it without forcing maximize'`) to simulate a thumbnail click using `MouseEvent` and verify that the correct window focuses while retaining a `'normal'` state. Updated existing inert assertions to match the new child-element logic.

## Testing Instructions
1. Open multiple overlapping windows (e.g., Pages, Media, Settings).
2. Ensure at least one window is partially or fully obscured by the currently active window.
3. Enter Overview mode via the top bar menu (Arrange windows → Overview).
4. Click the thumbnail of one of the background windows.
5. Verify that Overview mode closes, and the window you clicked is now active and brought to the very front.

## Screencast

https://github.com/user-attachments/assets/db80294c-4fac-469f-adc6-1c7ba471d7be

## Use of AI Tools
AI assistance: Yes
Tool(s): Antigravity
Model(s): Gemini
Used for: Diagnosing the issue, updating tests, and assistance with implementation.

### DIFF
--- a/src/window-manager/overview.ts
+++ b/src/window-manager/overview.ts
@@ -3,7 +3,7 @@
  *
  * Animate every eligible window to a grid thumbnail, plus a top bar
  * showing one tile per virtual desktop. Clicking a thumbnail exits
- * overview and fullscreens the clicked window. Pressing Escape or
+ * overview, focusing and bringing the clicked window to the front. Pressing Escape or
  * clicking the backdrop exits without selection.
  *
  * The lifecycle is big (enter/exit + click + key handlers + top bar
@@ -74,9 +74,9 @@ function inertWindowChildren( mgr: WindowManager, inactive: boolean ): void {
 
 /**
  * Enter overview mode — animate every eligible window to a grid
- * thumbnail layout. Clicking a thumbnail exits overview and
- * fullscreens the clicked window. Pressing Escape or clicking the
- * backdrop exits without selection.
+ * thumbnail layout. Clicking a thumbnail exits overview,
+ * focusing and bringing the clicked window to the front. Pressing Escape
+ * or clicking the backdrop exits without selection.
  */
 export function enterOverview( mgr: WindowManager ): void {
 	if ( mgr._overviewActive ) {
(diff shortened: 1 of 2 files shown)


src/window-manager/overview.ts
AllTerrainDeveloper: We should update this comment too :) That would make consisten the change, with the new behaviour and the docs :)

AllTerrainDeveloper (commented):
Thank you for the contribution! Would you mind update the comment I have pointed to? It's a tiny nit :) 

Again, thank you!

src/window-manager/overview.ts
CookieDarb: Hi! Thanks for the review. I have now updated all the relevant comments to match the new behaviour. Do let me know if there are any other feedbacks.

AllTerrainDeveloper (approved):
Thank you!

AllTerrainDeveloper:

CookieDarb:
Glad to be there :D 
Thank you!

PR #390, opened 2026-07-22
Fix content graph taxonomy fallback

### LINKED ISSUE #356: Display Categories for Pages in the Content Graph
Currently, the Content Graph groups Posts by their assigned categories, but Pages are always displayed under "Uncategorized" because WordPress pages do not support categories by default.

**Current Behavior:**

Posts are grouped based on their assigned categories.
Pages are displayed under "Uncategorized" regardless of their organization.

**Expected Behavior:**

If page categories are enabled, pages should be grouped by their assigned categories, similar to posts.
If categories are not supported for pages, pages should either:
Be grouped under a separate Pages section without showing "Uncategorized", or
Use an alternative grouping method (e.g., Parent Page or Page Hierarchy).

**Acceptance Criteria:**

Posts continue to be grouped by categories.
Pages are not incorrectly displayed under "Uncategorized".
The grouping for pages should accurately reflect the available taxonomy or hierarchy.
The Content Graph UI remains consistent and intuitive.

Image

CookieDarb:
## What?
Closes #356  

Fixes the bug where Pages and other Custom Post Types (CPTs) without taxonomy support are incorrectly grouped into a fallback "Uncategorized" or "Untagged" cluster in the Content Graph.

## Why?
Previously, the Content Graph client-side logic grouped any node with an empty `category_ids` array into `cat:uncat` ("Uncategorized"), and empty `tag_ids` into `tag:untagged` ("Untagged"). 
While this makes sense for Posts (which support categories but might have their terms cleared), it creates misleading clusters for Pages and CPTs that don't support those taxonomies at all. Pages were visually dumped into the same "Uncategorized" and "Untagged" buckets as standard Posts.

## How?
- `includes/content-graph/window.php`:
  - Updated the backend `desktop_mode_content_graph_post_types()` config to expose a `taxonomies` object for each post type, indicating whether it supports `category` and `post_tag`.
- `src/content-graph/types.ts`: 
  - Updated the frontend `PostTypeDescriptor` interface to require the new `taxonomies` field.
- `src/content-graph/scene.ts`: 
  - Added `postTypeSupportsTaxonomy()` to check if a post type supports the active facet.
  - Updated `deriveGroupKeys()`: if a post type does *not* support the active taxonomy, it now generates a distinct grouping key (e.g., `cat:type_page` or `tag:type_page`) instead of falling back to the generic "Uncategorized" or "Untagged" pools.
  - Updated cluster label resolution to render the real post type name (e.g., "Pages") for these isolated clusters.
- `tests/phpunit/tests/contentGraphGroupBy.php`:
  - Fixed and expanded PHPUnit tests to verify the backend data contract for taxonomy fallbacks. Added coverage to guarantee that taxonomy-supporting posts correctly inject the default category (`term ID 1`), the fallback term resolves in the catalog, and CPTs without taxonomy support correctly bypass this injection.

## Testing Instructions
1. Open the Content Graph in the desktop mode environment.
2. Select **Group by category** from the toolbar.
3. Observe that Pages are now grouped in their own distinct "Pages" cluster, rather than being grouped under "Uncategorized".
4. Select **Group by tag** from the toolbar.
5. Observe that Pages are similarly grouped under their own "Pages" cluster, while untagged Posts correctly fall under "Untagged".
6. If you test with a custom post type that does not support categories or tags, verify that it also forms its own distinct cluster based on the post type name.

## Screenshots
**Group by Category**
Default_Group_By_Cat

**Group by Tag**
Default_Group_By_Tag

## Use of AI Tools
AI assistance: Yes
Tool(s): Antigravity
Model(s): Gemini, Claude
Used for: Diagnosing the issue, implementing the fix and tests. Final decisions and edits were made by me.

### DIFF
--- a/includes/content-graph/window.php
+++ b/includes/content-graph/window.php
@@ -48,7 +48,7 @@ function desktop_mode_content_graph_user_can_use() {
  *
  * @since 0.8.2
  *
- * @return array[] Each entry: `array( 'slug', 'label', 'icon' )`.
+ * @return array[] Each entry: `array( 'slug', 'label', 'icon', 'taxonomies' )`.
  */
 function desktop_mode_content_graph_post_types() {
 	$types  = get_post_types( array( 'public' => true ), 'objects' );
@@ -61,15 +61,19 @@ function desktop_mode_content_graph_post_types() {
 			continue;
 		}
 		$result[] = array(
-			'slug'  => (string) $type->name,
-			'label' => (string) $type->labels->name,
-			'icon'  => (string) ( ! empty( $type->menu_icon ) ? $type->menu_icon : 'dashicons-admin-post' ),
+			'slug'       => (string) $type->name,
+			'label'      => (string) $type->labels->name,
+			'icon'       => (string) ( ! empty( $type->menu_icon ) ? $type->menu_icon : 'dashicons-admin-post' ),
+			'taxonomies' => array(
+				'category' => is_object_in_taxonomy( $type->name, 'category' ),
+				'post_tag' => is_object_in_taxonomy( $type->name, 'post_tag' ),
+			),
 		);
 	}
 
 	/**
 	 * Filter the list of post types shown in the Content Graph filter
-	 * bar. Each entry must declare `slug`, `label`, and `icon`. Removing
+	 * bar. Each entry must declare `slug`, `label`, `icon`, and `taxonomies`. Removing
 	 * an entry hides it from the filter bar AND excludes it from the
 	 * graph entirely.
 	 *
--- a/includes/content-graph/window.php
+++ b/includes/content-graph/window.php
@@ -73,7 +73,10 @@ function desktop_mode_content_graph_post_types() {
 
 	/**
 	 * Filter the list of post types shown in the Content Graph filter
-	 * bar. Each entry must declare `slug`, `label`, `icon`, and `taxonomies`. Removing
+	 * bar. Each entry declares `slug`, `label`, `icon`, and optionally
+	 * `taxonomies` (`array( 'category' => bool, 'post_tag' => bool )`);
+	 * entries missing `taxonomies` get it derived from
+	 * `is_object_in_taxonomy()` after filtering. Removing
 	 * an entry hides it from the filter bar AND excludes it from the
 	 * graph entirely.
 	 *
--- a/includes/content-graph/window.php
+++ b/includes/content-graph/window.php
@@ -82,7 +82,17 @@ function desktop_mode_content_graph_post_types() {
 	 * @param array[] $result Default: every public post type except attachment.
 	 */
 	$filtered = apply_filters( 'desktop_mode_content_graph_post_types', $result );
-	return is_array( $filtered ) ? array_values( $filtered ) : $result;
+	$filtered = is_array( $filtered ) ? array_values( $filtered ) : $result;
+
+	foreach ( $filtered as $i => $entry ) {
+		$slug                         = isset( $entry['slug'] ) ? (string) $entry['slug'] : '';
+		$filtered[ $i ]['taxonomies'] = array(
+			'category' => isset( $entry['taxonomies'] ) ? ! empty( $entry['taxonomies']['category'] ) : is_object_in_taxonomy( $slug, 'category' ),
+			'post_tag' => isset( $entry['taxonomies'] ) ? ! empty( $entry['taxonomies']['post_tag'] ) : is_object_in_taxonomy( $slug, 'post_tag' ),
+		);
+	}
+
+	return $filtered;
 }
 
 /**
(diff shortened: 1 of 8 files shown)


includes/content-graph/window.php
```diff
@@ -61,15 +61,19 @@ function desktop_mode_content_graph_post_types() {
 			continue;
 		}
 		$result[] = array(
-			'slug'  => (string) $type->name,
-			'label' => (string) $type->labels->name,
-			'icon'  => (string) ( ! empty( $type->menu_icon ) ? $type->menu_icon : 'dashicons-admin-post' ),
+			'slug'       => (string) $type->name,
+			'label'      => (string) $type->labels->name,
+			'icon'       => (string) ( ! empty( $type->menu_icon ) ? $type->menu_icon : 'dashicons-admin-post' ),
+			'taxonomies' => array(
```
epeicher, line 67: The scene actually gets its `postTypes` from this window-config path (`index.ts` passes `cfg.postTypes` into `GraphScene`), but only `rest.php` got a missing-`taxonomies` fallback. Any existing filter callback that adds/replaces entries per the old contract (`slug`/`label`/`icon`) leaves `taxonomies` undefined, and `pt.taxonomies[ taxonomy ]` in `scene.ts` throws, breaking group-by entirely.

Suggest normalizing after `apply_filters()` on line 84 so both consumers are covered (the `rest.php` fallback can then be dropped):

```php
$filtered = is_array( $filtered ) ? array_values( $filtered ) : $result;
foreach ( $filtered as $i => $entry ) {
	$slug = isset( $entry['slug'] ) ? (string) $entry['slug'] : '';
	$filtered[ $i ]['taxonomies'] = array(
		'category' => isset( $entry['taxonomies'] ) ? ! empty( $entry['taxonomies']['category'] ) : is_object_in_taxonomy( $slug, 'category' ),
		'post_tag' => isset( $entry['taxonomies'] ) ? ! empty( $entry['taxonomies']['post_tag'] ) : is_object_in_taxonomy( $slug, 'post_tag' ),
	);
}
return $filtered;
```

The `! empty()` cast also guards filters that pass `taxonomies` in the `register_post_type` list shape (`array( 'category' )`).

src/content-graph/scene.ts
```diff
@@ -941,14 +943,34 @@ export class GraphScene {
 		}
 	}
 
+	private postTypeSupportsTaxonomy( typeSlug: string, taxonomy: 'category' | 'post_tag' ): boolean {
+		const pt = this.postTypes.find( ( t ) => t.slug === typeSlug );
+		// Unknown type (not registered in the scene's post-type list) is
+		// treated as not supporting the taxonomy so it gets its own
+		// cat:type_<slug> / tag:type_<slug> cluster rather than silently
+		// merging into the shared Uncategorized / Untagged pool.
+		return pt ? pt.taxonomies[ taxonomy ] : false;
```
epeicher: I suggest to check config entries without `taxonomies` (old filter contract):

```suggestion
		return pt?.taxonomies?.[ taxonomy ] ?? false;
```

includes/content-graph/window.php
```diff
@@ -61,15 +61,19 @@ function desktop_mode_content_graph_post_types() {
...
+				'category' => is_object_in_taxonomy( $type->name, 'category' ),
+				'post_tag' => is_object_in_taxonomy( $type->name, 'post_tag' ),
+			),
 		);
 	}
 
 	/**
 	 * Filter the list of post types shown in the Content Graph filter
-	 * bar. Each entry must declare `slug`, `label`, and `icon`. Removing
+	 * bar. Each entry must declare `slug`, `label`, `icon`, and `taxonomies`. Removing
```
epeicher: This contract change needs a matching update in `docs/hooks-reference.md` (repo rule: filter changes ship with their doc). The `desktop_mode_content_graph_post_types` section (~line 3055) still says "Each entry declares `slug`, `label`, and `icon`." Suggested wording to be added to `docs/hooks-reference.md`:

```md
Each entry declares `slug`, `label`, `icon`, and `taxonomies` (`array( 'category' => bool, 'post_tag' => bool )`), used to keep types without a taxonomy out of the shared Uncategorized/Untagged clusters. Entries added without `taxonomies` get it derived via `is_object_in_taxonomy()`.
```

epeicher (commented):
Thanks for the changes @CookieDarb! This is a well described PR 👍 

I have tested them, and they work as expected. In the review, I have suggested a couple of changes: one is a check for empty taxonomies, and the other is in the documentation. Once they are addressed, this PR is ready to be merged 🚀

CookieDarb:
Thanks for the feedback! I've addressed the suggestions and also included a PHPUnit test to verify the normalization of legacy callbacks.

epeicher (approved):
Thanks for addressing the comments @CookieDarb! Changes LGTM! 

I will add a couple of minor nit changes and merge this, so it will be included in the next release 🙌

PR #414, opened 2026-07-24
Fix window drag bounds

### LINKED ISSUE #404: Window can be dragged out of bounds
We should decide what's the better option for the case when the user intentionally or unintentionally leaves a window out of the visible area:

https://github.com/user-attachments/assets/64fbda92-9f2e-4023-b32a-725f74fac3e5

This is somehow related with #399 andit would require a bit of thinking:

1. What happens if the whole web browser is resized?
2. What happens if the window is out of bounds (or at least, its title bar completely out)? Maybe MutationObserver would help to keep track of the tile bar mostly inside the wallpaper area

CookieDarb:
## What?
Closes #404 

Allows windows to be dragged partially off-screen without losing access to them, and smoothly recovers stranded windows during browser resizes. 

## Why?

This allows to replicate how macOS and Windows naturally handle desktop boundaries: allowing users to push large windows partially out of the way to clear visual space. 

Previously, floating windows could be dragged completely off-screen and lost. A prior attempt to fix this (PR #406) strictly trapped windows inside the viewport, which felt a bit restrictive. Additionally, making the browser window smaller would leave floating windows completely stranded outside the visible desktop area with no way to retrieve them without refresh or making the browser window large. 

## How?

- **Revert**: Reverted the strict boundary constraints introduced in PR #406, replacing them with a grab-margin based approach.
- `src/window/constants.ts`: 
  - Added a 40px `GRAB_MARGIN` constant to ensure a minimum clickable grab area is always maintained.
- `src/window/pointer.ts`: 
  - Introduced `clampWindowPosition()` to let windows bleed off the left, right, and bottom edges, stopping only when just 40px of the title bar remains visible. The top edge strictly locks at `0`.
  - Updated `handleDragStart()` to enforce these new boundaries during drag moves.
- `src/window-manager/index.ts`: 
  - Updated `reflowStatefulWindows()` to include standard floating windows. If a browser resize leaves a window stranded off-screen, it smoothly pulls it back into the 40px grab area.
- `tests/vitest/drag-unstate.test.ts`:
  - Added dedicated unit tests for the new clamping constraints and drag bounds.

## Testing Instructions

1. Open a standard floating window.
2. Drag the window past the left, right, and bottom edges of the desktop. Verify it stops moving once only 40px of the title bar is left visible.
3. Drag the window to the top edge and verify it strictly stops at the top boundary.
4. Drag a window to the bottom-right corner, then shrink your browser window so the window is stranded outside the visible desktop area.
5. Verify the window smoothly reflows back into view, stopping at the 40px grab margin limit.

## Screencast

**Before**

https://github.com/user-attachments/assets/ab16343c-eb3b-4a65-ac3d-0fe63a6a12da

**After**

https://github.com/user-attachments/assets/eee58751-72bb-4d6d-a58c-9f5f964cc5ed

## Use of AI Tools
AI assistance: Yes
Tool(s): Antigravity
Model(s): Gemini and Claude
Used for: Reviewing the existing implementation and suggesting the changes. Final decisions and edits were made by me.

### DIFF
--- a/src/window-manager/index.ts
+++ b/src/window-manager/index.ts
@@ -66,6 +66,7 @@ import {
 } from './snap-zones';
 import { cancelOverviewTimers, enterOverview, exitOverview } from './overview';
 import { loadNativeWindowGeometry } from './native-window-geometry';
+import { clampWindowPosition } from '../window/pointer';
 
 /** Base z-index for desktop windows. */
 const BASE_Z_INDEX = 100;
@@ -379,9 +380,12 @@ export class WindowManager {
 	/**
 	 * Re-apply state-driven bounds to any window whose geometry is
 	 * derived from the desktop area's dimensions: maximized (full
-	 * area) and snapped-left / snapped-right (half area). Called from
-	 * the desktop-area ResizeObserver so shrinking the browser window
-	 * drags the stateful windows along with it.
+	 * area) and snapped-left / snapped-right (half area). Also
+	 * clamps normal (floating) windows to the GRAB_MARGIN boundaries
+	 * so they are not stranded off-screen when the viewport shrinks.
+	 *
+	 * Called from the desktop-area ResizeObserver so shrinking the
+	 * browser window drags the windows along with it.
 	 *
 	 * Inlines the geometry writes instead of calling `applySnap` —
 	 * that method emits `_emitChange('state')` which would spam the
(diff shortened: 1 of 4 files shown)


AllTerrainDeveloper (approved):
WOAH! It feels amazing. It's very clean and clever! Thank you! ❤️

PR #403, opened 2026-07-23
Add Focus Timer widget

Alexismlg:
Fixes #410

## Summary

Adds a **Focus Timer** widget — a focus/countdown timer you add from the widget picker. Pick a duration, optionally link it to one of your open windows, and start. When time is up the linked window **shakes** (via the public `Window.shake()`) and a gentle alarm rings until you press **Stop**.

Built following the `starter-widget` template and using only public APIs + runtime globals — no framework internals touched.

## Features

- Duration presets (5 / 15 / 25 / 45 min) plus ±1-minute steppers (min 1 min, max 3 h).
- Optionally **link the timer to any open window** — that window shakes when time is up (looped `Window.shake()` until dismissed). Only currently-open windows are offered, and closing the linked window while the timer runs cancels it with a toast (the runtime listens for `desktop-mode-window-closed`).
- A Web Audio alarm chime that rings until dismissed — **synthesized in the browser, no audio files bundled**.
- Show / hide the remaining time while running.
- The timer keeps running across widget re-renders / re-docks (page-wide runtime singleton) and resumes across a page reload while still counting down (mirrored to `ctx.storage`).

## Implementation

- `src/plugins/focus-timer-widget/`
  - `index.ts` — the view. Uses `wpd-select` (window picker) + `wpd-checkbox-label` (toggle) and native buttons, with `__`/`sprintf` from `src/i18n`.
  - `timer.ts` — a DOM-independent state machine (idle / running / paused / finished) held on `window.__desktopModeFocusTimer` so it survives re-mounts; persists to `ctx.storage`.
  - `alarm.ts` — Web Audio alarm; the AudioContext is primed on the Start click (a user gesture) so it can fire minutes later.
  - `desktop.ts` — a thin typed accessor over `wp.desktop.windowManager` to list + shake windows.
  - `styles.css` — layout only; overrides `--desktop-mode-text` / `--desktop-mode-muted` on the dark-glass card (same approach as the Note Pad widget) so the `wpd-*` components stay legible.
- `includes/widgets/widget-focus-timer.php` — registration (no REST, no server state, no external services).
- Wired into `desktop-mode.php`, `vite.config.js` (`widget-focus-timer` target), and `package.json` (`build:widget-focus-timer` + the `build` chain).

Pure client-side — no REST routes, no server state, no network calls.

## Testing

- `npm run build:widget-focus-timer`, `npm run typecheck`, and `npm run lint` (on the new sources) all pass.
[...]

PR #416, opened 2026-07-24
Focus Timer: react to the linked window closing

### LINKED ISSUE #410: Pomodoro widget should listen to window events
Image

Has no sense to show a closed window to interact with. Similarly, what happens if a window with timer is closed? should the timer be cancelled?

Alexismlg:
Fixes #410

Follow-up to #403 (which added the Focus Timer widget) — addresses the review feedback in #410.

## Changes

- The window picker now lists **only currently-open windows** — a closed window is nothing to shake, so it is never offered as a choice (drops the previous dangling "(closed)" entry).
- The timer runtime now **listens for `desktop-mode-window-closed`**. If the linked window is closed while the timer is running or paused, the timer is **cancelled** (back to idle) and a toast explains why. The listener lives in the page-wide runtime rather than the view, so it fires even when the widget card has been re-docked or torn down.

## Testing

- `npm run typecheck`, `npm run lint` (on the new sources), and `npm run build:widget-focus-timer` all pass.
- Verified in wp-env: closing a *different* window does not affect a running timer; closing the *linked* window cancels it (→ idle) with a toast; the picker no longer lists closed windows.

### DIFF
--- a/src/plugins/focus-timer-widget/desktop.ts
+++ b/src/plugins/focus-timer-widget/desktop.ts
@@ -27,6 +27,7 @@ interface WindowManager {
 
 interface DesktopApi {
 	windowManager?: WindowManager;
+	showToast?( opts: { message: string; type?: string } ): unknown;
 }
 
 function desktopApi(): DesktopApi | undefined {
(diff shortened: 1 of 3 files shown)


AllTerrainDeveloper (approved):
Thanks! Works as expected!

PR #422, opened 2026-07-25
Add a window corner-radius setting (Sharp / Default / Round)

Alexismlg:
## Summary

Adds an **OS Settings → Appearance** control, **"Window corners"**, that lets the user choose how rounded window corners are:

- **Sharp** — 0px
- **Default** — 8px (current behaviour, unchanged for existing users)
- **Round** — 16px

The choice writes the existing `--desktop-mode-window-radius` custom property on the shell root during the settings apply pass, so every open window reflows its corners **live** (no reload). It persists per user like the other appearance settings.

## Why

Corner radius was a fixed token (`--desktop-mode-window-radius: 8px`) with no way to change it short of custom CSS. Exposing it as a segmented control is a small, self-contained personalisation that fits alongside Accent color / Dock size.

## Implementation

Follows the **Dock size** section pattern exactly:

- `src/settings/constants.ts` — `WINDOW_RADII` presets + `windowRadius` default.
- `src/settings/types.ts` — `WindowRadiusId` type + `windowRadius` on `OsSettingsState`.
- `src/settings/state.ts` — default + validated coercion on load.
- `src/settings/labels.ts` — `translateWindowRadiusLabel`.
- `src/settings/sections/window-radius.ts` — new `<wpd-segmented>` section (mirrors `dock-size.ts`).
- `src/settings/panel.ts` — mounts the section under the Appearance tab (after Dock size).
- `src/settings/index.ts` — apply pass sets `--desktop-mode-window-radius` on `<html>` (next to the dock-size / accent vars).
- `includes/os-settings.php` — server-side default, sanitise (allow-list of the three ids), and output, mirroring `dockSize`.

Only affects **windows** — widgets and the dock keep their own radii.

## Testing

- `npm run typecheck`, `npm run lint`, `php -l` on the PHP file, and builds of the `desktop` + `os-settings-panel` targets all pass.
- Verified in wp-env: the control renders in Appearance; picking Sharp/Default/Round changes `--desktop-mode-window-radius` to `0/8/16px` live; the choice survives a page reload (persisted server-side).

### DIFF
--- a/includes/os-settings.php
+++ b/includes/os-settings.php
@@ -20,6 +20,9 @@ const DESKTOP_MODE_OS_SETTINGS_META_KEY = 'desktop_mode_os_settings';
 /** Valid dock-size IDs — mirrors the TS `DOCK_SIZES` constant. */
 const DESKTOP_MODE_OS_SETTINGS_DOCK_SIZES = array( 'compact', 'default', 'large' );
 
+/** Valid window-radius IDs — mirrors the TS `WINDOW_RADII` constant. */
+const DESKTOP_MODE_OS_SETTINGS_WINDOW_RADII = array( 'sharp', 'default', 'round' );
+
 /** Valid desktop-layout IDs — mirrors the TS `DESKTOP_LAYOUTS` constant. */
 const DESKTOP_MODE_OS_SETTINGS_DESKTOP_LAYOUTS = array( 'classic', 'unified', 'spatial' );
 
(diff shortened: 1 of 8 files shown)


AllTerrainDeveloper (approved):
LGTM! Thank you!

https://github.com/user-attachments/assets/e29cb287-af9c-4ad1-b606-7417331dc3f6

PR #424, opened 2026-07-27
Fix Cmd+K entity search not finding posts in Commands mode

### LINKED ISSUE #411: Cmd+K post search works in Classic Admin but not in Desktop Admin Mode
When I press Cmd+K in Classic Admin Mode, I can quickly search for any post or page by typing its title and open it directly. But when I switch to Desktop Admin Mode, pressing Cmd+K opens the Site Assistant, and searching for posts by title no longer works — I can't find them at all.

## Steps to reproduce
  1. Be in Classic Admin Mode
  2. Press Cmd+K - WordPress's command palette opens
  3. Start typing a post title - matching results appear instantly
  4. Click a result - you're taken to that post's edit screen
  5. Now switch to Desktop Admin Mode
  6. Press Cmd+K - the Site Assistant overlay opens instead
  7. Type the same post title in the Commands search bar - no matching posts appear

### In Classic Mode:
- Cmd+K opens WordPress Core's built-in command palette (from @wordpress/commands). 
- Core ships a set of entity search commands (powered by @wordpress/core-commands) that can find posts, pages, and other content by title via REST API queries. 
- This works out of the box - no setup needed.

### In Desktop Mode:
- Cmd+K is intentionally intercepted and replaced with the Site Assistant. 
- This is by design: Core's command palette contains navigation commands that would bypass Desktop Mode's window system and break the desktop experience. 
- The Site Assistant opens in its place with two modes:
  - Commands mode - shows registered commands (admin menu navigation links, plugin commands, etc.).
  - Ask AI mode - natural language search that requires an AI provider to be configured.

The problem is that neither mode replicates the post-search functionality that Core's command palette provides for free. The Commands mode only searches command names, not post titles. The Ask AI mode can theoretically find posts, but only if the user has set up and enabled an AI provider - a much higher bar than what Classic Mode requires.

## Why this matters
Searching for a post by title via Cmd+K is one of the most common workflows for WordPress admins. It's fast, keyboard-driven, and universal. Desktop Mode currently breaks this flow without offering an equivalent replacement, which means users who switch to Desktop Mode lose a productivity tool they likely use daily.

## Looking for suggestions
This seems like a gap that was probably overlooked when the Site Assistant replaced Core's command palette - the entity search commands from @wordpress/core-commands were never ported over. I'd be curious to hear thoughts on what the best path forward looks like. Some possibilities that come to mind:
- Could the Site Assistant's Commands mode be extended to show live post/page/media search results inline when typing? This would restore the functionality without requiring any AI setup.
- Could we hook into the same REST API endpoints (/wp/v2/search) that Core's entity search uses, and surface results as clickable items in the existing command results list?
- Is there a simpler approach that I'm not seeing?

Would love to hear what the team thinks makes the most sense here - especially in terms of keeping the implementation simple and maintainable while making Desktop Mode a clear upgrade over Classic Mode rather than a trade-off.

KarunyaChavan:
This PR fixes the Cmd+K command palette so typing a post or page title in Commands mode shows matching results you can click or press Enter to open in the editor - *matching the behavior of Classic Admin Mode's Cmd+K*.

Closes #411 

## Why?

- In Classic Admin Mode, pressing Cmd+K and typing a post title instantly showed results and let you open the post editor with one click. 
- Desktop Mode's Cmd+K showed "No commands matching…" instead. 
- Users switching to Desktop Mode expect the same quick-content-search muscle memory to work. 
- The REST API search existed but the results were never fetched, displayed, or made clickable.

## What changed?
- New REST search (_fetchRemoteCommands): When you type plain text in Commands mode, the palette now calls /wp/v2/search (200ms debounced) and turns matching posts/pages into clickable command items.
- Click handler rewrite: Previously the click handler used findCommand(slug) which only finds registered slash-commands. Entity search results aren't registered commands, so clicks did nothing. Now it looks up by data-index from the live match list instead.
- Result styling: Entity results get a regular font (not the monospace used for slash-commands) via the new .is-entity-result CSS class, so they visually read as content search results.
- Admin URL config: Added adminUrl to the assistant config so entity result links resolve to absolute wp-admin/post.php?post=N&action=edit URLs for opening in legacy windows.

## Steps for Testing:
1. Press Cmd+K to open the command palette (make sure you're in Commands mode, not Ask AI)
2. Type the title of an existing post or page - matching results should appear below the command list
3. Click a result - it should open the post/page editor in a legacy window
4. Switch to Ask AI mode and type a natural language question like "How do I create a post?" - pressing Enter should submit the question to the AI, not open a random post


KarunyaChavan:
We implemented entity search (posts/pages) in the Cmd+K palette — type a title, see results, click to Edit. This closes the gap with Classic Admin Mode.

One detail we didn't replicate: Classic mode shows "Edit" as a right-aligned badge on each result row. We left it out because:
- Each result already shows "Post" or "Page" as a description line below the title
- Entity results already get a different font (regular vs. monospace for slash-commands) via CSS
- Adding a third visual indicator felt like clutter

Would you like us to add "Edit" text anyway to match classic mode exactly? Or is the current approach (description + font difference) sufficient?

AllTerrainDeveloper (approved):
> Would you like us to add "Edit" text anyway to match classic mode exactly? Or is the current approach (description + font difference) sufficient?

We don't necessarily need to replicate everything, if we can improve it :) That's what you did :) Thank you!

And because I tested yours, I found a bug of mine :) 
image

I'll quickly fix the theming issue, which is not applied in the command palette

PR #438, opened 2026-07-28
Make the command palette follow the active desktop theme

AllTerrainDeveloper:
The Site Assistant / command palette overlay was themed only halfway. Its text tones already read `--wpd-fg` / `--wpd-fg-muted`, but the panel itself was a hardcoded `rgba( 255, 255, 255, 0.97 )` and every hairline was a literal `rgba( 0, 0, 0, … )`. Under a dark desktop theme that put the theme's *light* foreground on a *white* card — result rows came out nearly invisible.

The search field was worse than untouched. It's a raw `<input type="text">` living in the parent shell, so core's `forms.css` reaches it:

```css
input[type="text"] { background-color: #fff; color: #2c3338; border: 1px solid #8c8f94; … }
```

That selector weighs (0,1,1) and beat the bare `.desktop-mode-ai__input` class at (0,1,0), so core painted its own white bordered box with a focus ring — ignoring both the palette and the stylesheet's own chromeless design intent.

## What changed — `assets/css/ai-assistant.css`

- **Panel** takes `--wpd-surface`, `--wpd-fg` and `--wpd-border` (the hairline ring), plus the desktop-theme `DIALOG` texture slot. Background is longhand so the texture slot isn't reset by a shorthand.
- **Backdrop** picks up the `SCRIM` texture slot alongside the `--wpd-scrim` colour it already had.
- **Every remaining literal border** resolves through `--wpd-border` / `--wpd-border-strong`, including the ones mixed with the accent via `color-mix()`.
- **Input rule scoped to the overlay root** — `.desktop-mode-ai .desktop-mode-ai__input` weighs (0,2,0) and beats core. Same fix as the native-window form-control rule in `window-chrome.css`. Border, radius, box-shadow and padding are reset; `min-height: 30px` keeps the hit target core was providing.
- **Active mode pill** prefers `--wpd-surface-elevated` so it stays legible against a header that is itself `--wpd-surface`.

Every token keeps the literal that was already there as its fallback, so **an unthemed shell is pixel-identical** — with one deliberate exception: the search field now renders chromeless, as its stylesheet always intended.

The three black atmospheric shadows stay literal on purpose: they read as depth against the wallpaper, not as part of the surface.

## Docs

- `DIALOG` / `SCRIM` slot rows name the command palette as a consumer.
- The `--wpd-*` palette section notes that body-mounted shell overlays (toasts, dialogs, context menus, the palette) read the palette too.
- The core-CSS note points at both tokenized form-control overrides.

## How to test

1. OS Settings → Themes, activate a dark desktop theme.
2. Hit ⌘K, switch to **Commands**, type something that returns post results.
3. Panel, footer, borders and the search field should all sit in the theme's palette; non-selected result rows must be readable.
4. Deactivate the theme and reopen — everything back to the default look, except the search box, which is now borderless/transparent instead of core's white box.

## Checks

`npm run build` · `npm run lint` · `npm run typecheck` · `npm run test:js` (2448 tests) — all green. No PHP touched.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22landingPage%22%3A%22%2Fdesktop-mode%2F%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fdesktop-mode%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-438-2a0790819f56f35e37164e6cd9d7e218d9411b1e.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/assets/css/ai-assistant.css
+++ b/assets/css/ai-assistant.css
@@ -7,10 +7,21 @@
  * entrance, large search input, accent-tinted header.
  *
  * z-index 10000 puts this above --desktop-mode-z-adminbar (9991) so the
- * overlay truly covers the full viewport. The panel itself is white/light
- * regardless of wallpaper, providing consistent contrast.
+ * overlay truly covers the full viewport.
  *
- * @since 0.14.0
+ * ## Desktop-theme awareness
+ *
+ * The overlay mounts on `document.body`, so it sits inside the
+ * `body.desktop-mode-desktop-theme-<slug>` half of a compiled theme's
+ * doubled selector and every `--wpd-*` token below resolves against the
+ * active theme. Each site reads `var( --wpd-x, <the literal that was
+ * always there> )`, so an unthemed shell paints exactly as before.
+ *
+ * Getting this half-right is worse than not doing it at all: when the
+ * text tones followed the theme but the panel stayed hardcoded white,
+ * a dark theme's light `--wpd-fg` landed on white and the results were
+ * unreadable. Any new surface added here MUST take its background,
+ * border and text from the palette — not from a literal.
  */
 
 /* ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


AllTerrainDeveloper (commented):
it correctly renders the command palette now

Image

PR #440, opened 2026-07-28
Fix inline reply/edit in the native Comments window (shadow-DOM row lookup)

Alexismlg:
## Summary

Inline **reply** and **edit** never opened in the native Comments window — clicking them (or pressing `r` / `e`) did nothing, no editor appeared.

`openReplyFor()` / `openEditFor()` located the target row with a light-DOM lookup:

```js
const tr = state.tableHost?.querySelector( 'tr[data-row-id="…"]' );
tr?.after( editorHost ); // inserts the editor after the row
```

But `<wpd-table>` renders its rows inside its **(open) shadow DOM**. `querySelector` doesn't cross the shadow boundary, so `tr` was `null`, `tr?.after()` was a no-op, and the editor ended up in a **detached node** that never reached the page.

The native Comments window is **opt-in** (Settings → Features → *Use the native Comments experience*) and **off by default**, which is why this went unnoticed when the table moved to shadow DOM — there's no test covering the interaction and little real-world exposure.

## Fix

- **Re-anchor the editor** as a docked bottom sheet over the table body (`inlineHostFor()`), instead of trying to insert it after a shadow-DOM `<tr>`. A single host keeps reply and edit mutually exclusive, and a context line names the target comment (e.g. *"Replying to Marta Ruiz"*).
- **Surface a visible `Reply` action** in the bulk bar, shown when exactly one comment is selected (mirroring the `r` shortcut's guard). Inline reply was previously reachable **only** via the keyboard shortcut — undiscoverable for pointer users.
- **Fix keyboard `j` / `k` scroll-into-view**, which hit the same light-DOM row lookup, by reaching through the table's shadow root.

## Notes

- The docked sheet sits at `z-index: 50` so it clears the table's sticky header/column cells (`<wpd-table>` doesn't isolate its stacking context, so its sticky cells — z-index up to 40 in the shadow tree — share the positioned ancestor's context; without this the sticky Author column bled through the sheet).
- No behavior change for the classic (chromeless `edit-comments.php`) experience — this only touches the native window bundle.

## Testing

- `npm run typecheck`, `npm run lint`, `npm run build` clean.
- `npm run test:js` green (unrelated pre-existing `desktop-files-restore-sync` failure aside).
- Manually verified in wp-env: with the native Comments window enabled, `Reply` and `Edit` now open the docked editor, the Author column no longer bleeds through, and `j`/`k` scroll the focused row into view.

### DIFF
--- a/assets/css/comments-window.css
+++ b/assets/css/comments-window.css
@@ -108,6 +108,8 @@
 	min-block-size: 0;
 	display: flex;
 	overflow: hidden;
+	/* Anchors the docked inline reply/edit sheet (position: absolute). */
+	position: relative;
 }
 
 .desktop-mode-comments__body wpd-table {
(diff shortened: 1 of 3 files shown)


AllTerrainDeveloper:
Wow, we don't know for how long it has been broken 😅. I can see there is no test for it. Maybe adding a test would help us next time we break it. Don't you think so?

AllTerrainDeveloper:
I'm closing this in favour of the awesome work you did in the other pull request :)

PR #441, opened 2026-07-28
Redesign the native Comments window as a two-pane conversation view

Alexismlg:
## Summary

Reworks the **opt-in native Comments window** from a flat moderation table into a **two-pane conversation view**, and wires it into the surrounding desktop (per-post scoping, editor jump, connection spline).

- **Left rail** — the list of *conversations* (top-level comments per tab): avatar, author, snippet, post, status dot, reply count.
- **Right pane** — the selected thread's **full nested reply chain** (every depth), each message carrying **always-visible actions** (Reply / Edit / Approve · Unapprove / Spam / Trash), plus a docked reply composer that repoints to whichever message you reply to.

Built entirely on existing Desktop Mode / wp-admin tokens and the `<wpd-*>` kit — no new visual vocabulary, and no AI.

## Why

Testing the native window surfaced three limitations in the table UI, all fixed here:

1. **Threading was one level deep.** The conversation pane renders the entire nested tree, built client-side from a single thread fetch.
2. **Reply / edit were keyboard-only** and undiscoverable. Every message now has visible action buttons — including nested replies.
3. The surface read like `wp-admin` rather than a Desktop Mode app.

## Opt-in, default unchanged

Opt-in and **OFF by default** (`nativeCommentsEnabled = false`, untouched). Users who don't opt in still get the classic `edit-comments.php` screen. This PR only changes what's *inside* the native window, which appears only when a user enables it under **Settings → Features → "Use the native Comments window"**.

## Desktop integration (added on top of the base redesign)

- **Per-post scoping.** `edit-comments.php?p=<id>` (WP's "comments on this post") now opens the native window scoped to that post — the rail shows only that post's conversations with an *"On: `<title>` / Show all"* banner. The Comments URL remap parses `?p=` in `onMatch` and stashes it in a small cross-bundle store the renderer reads (mirrors the user-edit target pattern); a store subscription re-scopes an already-open window, since native windows render once.
- **Related menu → native windows.** The "Related" title-bar menu now consults `tryNativeUrlRemap()` first (as the shell's link interceptor already does), so a related link opens the opted-in native window instead of a chromeless iframe of the classic page. This is a small framework-level change in `desktop.ts` — it benefits every native window (Posts/Users/…), and is the "revisit when native windows accept deep-link hints" the old comment there called for (the hint is now the remap's `onMatch`). **Flagging for review** as it touches shared code.
- **Editor jump.** A pencil in the conversation header opens the post's block editor as a Desktop Mode window (same link-interceptor path as the Drafts widget).
- **Connection spline.** When scoped to a post, the window announces its content identity via `wp.desktop.relations.set()` so the window-links spline to the post's editor is drawn — parity with what the classic iframe got from the chromeless bridge.

## Implementation

- `src/comments-window/conversation.ts` (new) owns all rendering + interaction; `index.ts` is a thin registry entry.
- `src/comments-window/rest.ts` gains `fetchThread()` (approve + hold merged, since `wp/v2/comments` rejects a multi-value `status`) and an optional `post` filter on `fetchComments`.
- `src/comments-window/post-filter.ts` (new) — the cross-bundle post-filter store.
- `src/desktop.ts` — the Comments remap `onMatch` + the Related-menu remap consult.
- `includes/comments-window/window.php` emits the two-pane shell; `assets/css/comments-window.css` adds the layout.
- Reuses the existing REST helpers (`bulkModerate`, `postReply`, `updateCommentContent`) wholesale — transport + capability gates unchanged.

## Security / capabilities

No change to the capability model. The Edit button renders only when the server reports `current_user_can( 'edit_comment', $id )` (`desktop_mode_can_edit`), and every mutation re-checks the capability server-side.

## Commit layout

1. Base conversation redesign (two-pane, nested thread, per-message actions, composer).
2. Post-filter store + `fetchComments` `post` param.
3. Renderer: editor pencil, per-post filter UI, window-links identity.
4. Framework: Related menu → native windows + Comments remap reads `?p=`.
5. Review fixes: `fetchThread` covers all four statuses (spam/trash threads render); consume-once post filter (no stale scope on a plain reopen); `loadRail` sequence guard; post-filter subscription dropped on window close; tabs toggle `aria-selected`; dead template nodes removed.

## Not in this PR (follow-ups)

Keyboard moderation (`j`/`k`/`a`/`s`/`d`) + undo; author-insights drawer; realtime "N new" pill; rail pagination / surfacing pending *replies* in the rail; first-run intro dialog; pruning the now-dead table styles.

## Testing

- `npm run typecheck`, `npm run lint`, `npm run build` clean; `npm run test:js` green (unrelated pre-existing `desktop-files-restore-sync` failure aside).
- Verified in wp-env with the native window enabled: nested threads at all depths, per-message actions without flicker (silent reloads), `edit-comments.php?p=` opens filtered with the banner, the Related menu opens the native filtered window, the editor pencil opens the block editor as a window, and the connection spline draws to the editor. With the toggle off the classic screen is unchanged.


AllTerrainDeveloper (approved):
This is freaking amazing. I've been testing and reviewing and I found a few things, if you don't mind for the sake of simplicty and velocity I'll do those:

1. I'll use links button there

Image

2. I've changed the header to have the same external icon as Gutenberg (and the link is now clickable so no need to add extra button

Image

3. Cliking on a conversation that is opened, reloads it.

4. Improved loading speed by adding promises.all 

5. We are not using the `wpd-` web components.
There is a huge list of wpd components that are reusable to keep UI consistent:

Image

OS Settings -> Components serves as an example library (Kind of story book self maintained)

Image

Please if you think I've made something worse or out of your vision, let me know and we will gladly check it

Alexismlg:
Thank you so much for testing, reviewing, and jumping in on the polish — really appreciate it! 🙏

All five make sense to me and none of them go against the vision, quite the opposite:

1. **Links button** — 👍
2. **Gutenberg-style external icon + clickable header** — cleaner than my extra pencil button, happy to drop it.
3. **Reloading an open conversation on click** — good catch.
4. **`Promise.all` speed-up** — 
5. **Using the `wpd-*` components** — 100% the right call; I'd actually flagged it as a follow-up in the description, so I'm really glad you're wiring it up properly with the kit. Thanks for the pointer to OS Settings → Components too.

Please go ahead with all of them — nothing feels worse or out of vision. Thanks again for the collaboration! 🚀

PR #443, opened 2026-07-29
Fix(window): refine page-title-action button redundancy and layout positioning

### LINKED ISSUE #391: Visual inconsistency and redundancy with the "Add New" button
## Description
In Desktop Mode, the in-page "Add New" button (e.g., "Add Post", "Add Page", "Add User") on admin pages looks visually inconsistent and redundant.

1. **Redundancy**: The window already provides tab navigation at the very top (e.g., `Posts | Add Post | Categories | Tags`). Having a secondary "Add Post" button directly below it feels repetitive.
2. **Visual Inconsistency (Normally)**: Under normal circumstances, the "Add Post" button floats awkwardly on the exact same line as the view filters (e.g., `All (6) | Published (4) | Trash (2) Add Post`). It looks misaligned and out of place.
3. **Behaviour With Admin Notices**: When an admin notice (like an error or success message) appears, the layout changes. The "Add Post" button is pushed up onto its own row, above the notice. While this looks much cleaner, the button itself is still redundant.

## Screenshots

**Without Admin Notice**

Image

**With Admin Notice**

Image

## Proposed Solution
There are two potential ways to improve this experience:
1. **Remove the button:** Hide the in-page button completely on screens where a matching window tab (like the "Add Post" tab) already exists, removing the redundancy.
2. **Fix the layout**: Adjust the styling so that the button always stays neatly on its own top row, whether an admin notice is visible or not.

If this UI is not intended, I'd be happy to work on a confirmed fix.

CookieDarb:
## What?
Closes #391 

This PR proposes a proposal to refine the visual experience of in-page action buttons (like "Add New", "Add Post", "Add User") in Desktop Mode. It removes these buttons when they duplicate tabs in the window header, and correctly positions them on their own row when they are kept.

## Why?
Having an in-page "Add New" button is redundant when the window already has a dedicated tab for the same action in the top header. Additionally, it floats awkwardly next to filter links without breathing room unless an admin notice pushes it onto its own line. This proposal makes the experience cleaner and more consistent.

## How?
- `includes/render/chromeless-bridge.php`:
  - Extracted all active submenu tab URLs in PHP (using the `desktop_mode_dock_item` filter to ensure custom tabs like "Add Theme" are included).
  - Passed these URLs to the iframe configuration payload.
  - Updated the in-iframe `MutationObserver` to intercept `.page-title-action` elements and remove them if their destination matches a tab URL (comparing pathnames and query parameters like `post_type`, `taxonomy`, `page`, and `action`).
- `assets/css/chromeless.css`:
  - Removed the page-specific override for `themes.php`.
  - Applied a universal layout style to `.page-title-action` to ensure preserved buttons always clear filter links and sit on their own row.
- `tests/phpunit/tests/desktopModeRender.php`:
  - Added a unit test to verify submenu URL extraction and injection into the script configuration.
- `docs/plugin-compat-layer.md`:
  - Updated the "Add Theme" tab documentation to reflect the new dynamic URL matching behavior.

## Testing Instructions
1. Navigate to **Posts** (`edit.php`) in Desktop Mode. Verify that the in-page "Add New" button is removed (since it matches the "Add New" tab).
2. Navigate to **Appearance** (`themes.php`). Verify that the "Add Theme" button is removed (since it matches the "Add Theme" tab).
3. Navigate to a screen with no submenus (e.g. a standalone plugin page or custom CPT without menus). Verify that the button remains visible and sits on its own row, cleanly separated from content below.
4. Navigate to **WooCommerce Orders**. Verify that the "Add order" button remains visible, because there is no equivalent tab in the header.

### 1. When tab is present
Screenshot 2026-07-29 at 11 38 01 AM

### 2. When tab is present + admin notice
Screenshot 2026-07-29 at 11 38 18 AM

### 3. When tab is absent (WooCommerce Orders list)
Screenshot 2026-07-29 at 11 38 56 AM

### 4. When tab is absent (WooCommerce Coupons list)
Screenshot 2026-07-29 at 11 39 22 AM

## Use of AI Tools
AI assistance: Yes
Tool(s): Antigravity
Model(s): Gemini
Used for: Reviewing the existing implementation and suggesting the changes. Final decisions and edits were made by me.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-443-90e364841679d59e6a6a29ffc0295f5fbcc08e19.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper (dismissed):
This worth some discussions around 🙂 

I remember some users were lost because they couldn't find the add buttons, as in the beginning I was also removing them.

Also seems we are removing tooooo much :)
image

image

We might need probably more work around some areas

CookieDarb:
Hi, thanks for the review!

> I remember some users were lost because they couldn't find the add buttons, as in the beginning I was also removing them.
> Also seems we are removing tooooo much :)

My bad! The mapping logic ended up being a bit too aggressive, and given the context, I think its best not to remove the buttons for now. 

However, if needed I can update the PR(or open new) to apply only the positioning changes, such that the "Add New" buttons always stays neatly on its own top row, rather than floating on the same line as the view filters. 

Screenshot 2026-07-29 at 6 42 41 PM

Let me know if that is preferred.

AllTerrainDeveloper:
No worries :) 

I think we can dedupe for now, the Add Post would be just a few pixels far away from where the original button was. Let's deal with the change, deduping is always nice.

mmtr:
@CookieDarb no worries about the delay. I've picked this up and pushed to your branch so the PR stays yours.

> The mapping logic ended up being a bit too aggressive

That was the whole problem, and it came down to `isUrlMatch()` falling through to `return true` when the path wasn't `admin.php` or `edit.php`. That is what removed the "Upload Plugin" button in the screenshot above: `plugin-install.php?tab=upload` shares a path with the `plugin-install.php` tab.

A button is now hidden only when its href exactly matches one of the current parent menu's submenu URLs. Upload Plugin stays, WooCommerce's "Add order" stays (it carries `action=new`, the Orders tab doesn't), and a screen with no submenu strip contributes no URLs at all, so plugin pages that only have the button keep it.

The layout change you proposed is in too, kept buttons sit on their own row.

Two things I ran into while testing:

- Core's in-page toggles are excluded by class. Their href is only a no-JS fallback, so matching on it was hiding the Media grid's inline uploader.
- Clicking "Upload Plugin" was already broken before this PR: our link interceptor beat `plugin-install.js` and navigated to `?tab=upload` instead of opening the drop zone in place, which left you on a page with no plugin cards and no way back. Fixed here since it's the same button.

Your branch was a fair way behind trunk, which has been rebranded since (`desktop_mode_*` is `openstation_*`, `.desktop-mode-chromeless` is `.os-chromeless`), so I merged trunk in and moved the implementation to `includes/render/chromeless-title-actions.php`. Happy to walk through any of it.

Some screenshots:

Before | After
--- | ---
Screenshot 2026-08-07 at 16 20 58 | Screenshot 2026-08-07 at 16 22 01
Screenshot 2026-08-07 at 16 21 12 | Screenshot 2026-08-07 at 16 22 20
Screenshot 2026-08-07 at 16 21 23 | Screenshot 2026-08-07 at 16 22 45
Screenshot 2026-08-07 at 16 21 28 | Screenshot 2026-08-07 at 16 22 54
Screenshot 2026-08-07 at 16 21 37 | Screenshot 2026-08-07 at 16 22 50"Upload Plugin" button remains visible

PR #452, opened 2026-07-30
Overview: show minimized windows in grid to fix badge count mismatch

### LINKED ISSUE #437: Overview: window count badge includes minimized windows when they are not shown in the grid
The Overview top-bar count badge and the grid of window thumbnails disagree on what constitutes a "visible" window when only some windows on the desktop are minimized.

### Case 1 — One window, minimized

1. Open a single window (e.g. Dashboard).
2. Minimize it.
3. Enter Overview.

**Result:** The minimized window is restored (the "Show Desktop" all-minimized path fires) and the grid shows it. The count badge says `1`.  _Matches_.

### Case 2 — Two windows, one minimized

1. Open two windows (e.g. Dashboard and Posts).
2. Minimize only the Dashboard window.
3. Enter Overview.

**Result:** Only one thumbnail tile appears in the grid (Posts), but the count badge in `.desktop-mode-overview-top-bar__tile-wrapper` still reads `2`.  _Mismatch_.

## Open question

What should the Overview do with partially minimized windows?

**Option — Show minimized windows with a visual indicator:** Keep minimized windows in the grid but render them with a dimmed/deprioritized style and let the count badge stay accurate at `2`.

Or any other behavior should Overview adopt? Happy to submit a PR once the direction is confirmed.

KarunyaChavan:
This PR fixes the window count badge in Overview mode including minimized windows while the grid filtered them out, causing a mismatch between the badge number and the visible thumbnail tiles. Instead of filtering minimized windows from the grid, they now render as dimmed thumbnails (60% opacity) so the badge and grid stay consistent.

Closes #437 

## Why: 

- When entering Overview with partially minimized windows (e.g., 2 open, 1 minimized), the count badge displayed 2 but the grid only showed 1 thumbnail — a confusing mismatch that undermines Overview as a workspace overview tool. 
- The fix also removes the brittle "all-minimized" auto-restore path (Show Desktop → Overview), since the grid now renders minimized windows directly.

## What Changed:

* **Overview:** Minimized windows now appear in the Overview grid. Removed the "all minimized" auto-restore behavior, restored the selected minimized window before focusing it on exit, and added helper functions to keep thumbnails renderable while preserving fullscreen state.

* **`src/window/index.ts`:** Updated the `minimize()` transition handler so Overview thumbnails don't accidentally hide their content during opacity transitions.

* **`src/window-manager/desktops.ts`:** Updated Overview relayout to include minimized windows and reuse the new layout helper functions.

* **`assets/css/window-overview.css`:** Dimmed minimized thumbnails (`opacity: 0.6`) so they're easy to distinguish, and removed a redundant `pointer-events` rule.

* **`tests/vitest/desktops.test.ts`:** Updated existing tests for the new minimized-window behavior and added coverage for the badge count regression, thumbnail rendering, transition handling, and fullscreen restoration.

## Steps for Testing:

1. Open two windows (e.g. Dashboard and Posts).
2. Minimize only the Dashboard window.
3. Enter Overview.
4. Verify: Both windows appear as grid tiles — Posts at full opacity, Dashboard at ~60% opacity.
5. Verify: The count badge in .desktop-mode-overview-top-bar__tile-wrapper reads 2, matching the number of visible tiles.
6. Click the minimized Dashboard tile to exit Overview.
7. Verify: Dashboard restores to its normal (non-minimized) state and receives focus.
8. Fullscreen edge case: Open a window, toggle fullscreen, minimize it, enter Overview. Select it. Verify: The window exits to fullscreen (not a thumbnail-sized layout).
9. Regression — "Show Desktop → Overview": Minimize all windows on the active desktop. Enter Overview. Verify: All windows appear as dimmed thumbnails in the grid (not an empty grid), and the count badge matches the grid tile count.


KarunyaChavan:
**Note:** These changes introduce a behavioral change.

Previously, if all open windows were minimized before entering Overview, they would all be restored automatically once user entered Overview.

With this change, minimized windows remain minimized when entering Overview. If the user selects a minimized window, only that window is restored. If no window is selected, all windows remain minimized, resulting in behavior similar to Show Desktop on Windows.

tests/vitest/desktops.test.ts
```diff
@@ -287,6 +302,86 @@ describe( 'WindowManager — virtual desktops', async () => {
 		).toBe( true );
 	} );
 
+	test( 'enterOverview makes completed-minimize windows renderable for thumbnails', async () => {
+		// Regression guard for the "minimized thumbnail renders blank" bug.
+		// After the minimize transition fires, `content-visibility: hidden`
+		// and `iframe.style.visibility = 'hidden'` are set on the window.
+		// enterOverview must reverse these so the overview thumbnail shows
```
AllTerrainDeveloper, line 309: Awesome

AllTerrainDeveloper (approved):
LGTM and very happy with those improvements! Thank you!

Image

PR #459, opened 2026-07-30
Add window reveal animations

AllTerrainDeveloper:
https://github.com/user-attachments/assets/4087bb5b-fa56-45e5-bca6-4f31c3b98953

image

Windows now uncover their content with a clip-path animation when they finish loading, instead of fading in. It plays right after the loading spinner clears — and on reloads and in-window navigation too, same as the spinner does.

Twelve reveals to pick from in **OS Settings → Effects**, plus a speed setting next to it. Off by default, so nobody gets one they didn't choose.

Sweep · Rise · Diagonal · Iris · Diamond · Curtain · Shutter · Blinds · Slats · Mosaic · Radar · Camera shutter

The animation runs on a layer in the shell that sits above the window content without wrapping it, so it can never interfere with the page it's revealing. Iframe and native windows behave identically.

**Camera shutter** is a real iris: six equilateral wedges tiling a hexagon over the window, sliding tangentially. Their shared edges offset by the same amount, so they stay flush the whole way while a centred hexagonal aperture opens between them. It's rendered as SVG rather than clip-path layers, because a lens iris has a cyclic overlap — every leaf over the next, the last back under the first — that a linear paint order can't represent.

## Theming

| Token | Default |
|---|---|
| `--desktop-mode-window-reveal-surface` | white |
| `--desktop-mode-window-reveal-edge` | `transparent` (off) |
| `--desktop-mode-window-reveal-edge-thickness` | unset |
| `--desktop-mode-window-reveal-duration` | unset |

The edge is a band that travels with the animation's leading edge, drawing whatever shape the active reveal has. It's off unless a theme colours it, and costs nothing while it is.

Themes can also recommend a reveal and a speed in their manifest, so the Themes tab button is now **"Apply X's recommended layout and effects"**.

## For plugin authors

`wp.desktop.registerWindowReveal()` takes a `from`/`to` clip-path pair, a `layers` array for effects whose parts overlap, or a `render()` that builds the DOM outright while the shell keeps the timing. Recipe in [`docs/examples/window-reveal.md`](docs/examples/window-reveal.md).

## Testing

Turn a reveal on in OS Settings → Effects and open or reload a window. Reloading a cached window exercises the fast path where the spinner never appears.

2681 JS tests and 1700 PHP tests pass. The reveals have been checked visually.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22landingPage%22%3A%22%2Fdesktop-mode%2F%22%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fdesktop-mode%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-459-f04adddd4112e6e7497c8b4451f0a9dee57be184.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


epeicher (approved):
Thanks for adding those animations, they look cool! I have tested them and they work as expected. I have also reviewed the code and I have not found any blockers. LGTM!

PR #472, opened 2026-08-01
CPT folders grouped by plugin, WooCommerce integration, and a public tile-drop seam

AllTerrainDeveloper:
Custom post types become browsable sections in the site window, grouped into a folder named after the plugin or theme that registered them. Post-kind tiles render their featured image instead of a uniform dashicon — which turns a WooCommerce catalogue into a photo grid.

https://github.com/user-attachments/assets/45df13e9-236a-44bb-bf44-d3a360c5216f

image
image
image

## Site window

- Every non-builtin `show_ui` post type the current user can edit becomes a section (`includes/my-wordpress/post-types.php`), grouped by registrant (`owner.php`): plugin, mu-plugin, or theme. Display names come from `get_file_data()` and `wp_get_theme()` — never `get_plugins()`, which doesn't exist yet at `init` and scans the plugins directory.
- **New `group` route.** The root collapses a plugin's sections into one folder that drills into its members. Breadcrumbs and the back button behave at every depth; a grouped section's parent is its folder, not the root.
- **A bridge for non-REST types.** Types with `show_in_rest => false` are re-exposed at `desktop-mode/v1/post-type/<slug>` by subclassing Core's `WP_REST_Posts_Controller`, so `_fields`, `_embed`, `search`, `status` and the pagination headers all behave as on `wp/v2` and the bundle needs no special-casing.
- Window registration moves to `init` 99 — the entity list is frozen into the window config and only emitted at `admin_enqueue_scripts`, so a CPT registered later never reached the bundle.

### Security on the bridge

These types opted out of REST deliberately, so the subclass is narrower than its parent:

- Core's `get_items_permissions_check()` returns `true` for any non-`edit` context — i.e. **public read**. Both read checks and the delete check are overridden to require the type's `edit_posts` capability in every context.
- Only `GET` collection, `GET` item and `DELETE` item are registered. No create, no update.
- `desktop_mode_my_wordpress_post_type_rest_enabled` vetoes the bridge per type.

Two Core seams need overriding for a non-REST type to work at all, neither reachable by filter: `check_is_post_type_allowed()` reads `show_in_rest` (left inherited, every row is filtered out and the collection returns empty), and `rest_get_route_for_post()` bails *before* applying its own filter, so `self` / `collection` links are fixed in `prepare_links()`.

## WooCommerce

Inert unless `class_exists( 'WooCommerce' )`; the 11 KB bundle only enqueues for users who can open the site window on a store. Nothing in `src/my-wordpress/index.ts` knows WooCommerce exists — it all runs through the public hook contract, so a third-party plugin would write identical code.

- Folder labelled **Woo** ("WooCommerce" wraps onto two lines in an 88px tile) with WooCommerce's own mark re-emitted as `currentColor` so it follows the desktop theme.
- **Orders** served through `wc_get_orders()`, which is storage-agnostic. HPOS keeps orders out of `wp_posts`, so the generic post-type path returned an empty folder on any modern store.
- **Bands**: orders by status (attention first), products by stock then category, coupons by validity. Ordering is computed server-side and replayed as `post__in`, so a band's rows arrive together instead of the grouping visibly reshuffling as the user scrolls.
- **Right pane**: price/stock/sales for products, total/customer/line items for orders, validity/discount/usage for coupons — including total discount actually given, which WooCommerce doesn't surface. Everything with an edit screen is a link to it.
- Out-of-stock, backorder, low-stock and sale products carry a `<wpd-ribbon>` on the tile.

## Public API — accepting drops on your own desktop icon

Promotes the files-layer tile-payload seam to `wp.desktop.files.registerTilePayloadHandler( type, handler )`. This is the supported answer to "my plugin's icon rejects everything dropped on it": the drop-target registry allows one target per element and the layer's reject claimant is installed last, so a competing `DropTarget` registered during `desktop-mode.files.tile-rendered` is always displaced.

```js
wp.desktop.files.registerTilePayloadHandler( 'shortcut', {
    appliesTo: ( { placement } ) => placement.file.ref === 'my-plugin',
    accept: ( data ) => data.kind === 'attachment',
    acceptLabel: 'Open in My Plugin',
    onDrop: ( session ) => open( session.payload.data ),
} );
```

Handlers may now share a payload type, resolved first-matching-`appliesTo`. The previous one-handler-per-type map meant a second `'shortcut'` handler **silently replaced the in-tree agent drop targets** — a live bug for the next plugin to use this.

## Bugs found along the way

| Bug | Cause |
|---|---|
| CPT→plugin attribution never populated | Resolved plugin files via `get_plugins()`, which doesn't exist until after `init`. Would also have attributed every type to Desktop Mode, since the backtrace starts in our own `payload.php`. Also repairs the dock's CPT attribution, silently dead until now. |
| Coupon previews rendered blank | `appendPostArticle` read `detail.content.rendered` unguarded; any post type without editor support (`shop_coupon`) threw before the article was appended. |
| Tiles overlapping the next band | Canvas height was measured by walking `host.children`, but `place()` runs while the tile is still detached — a partially filled last row overflowed its band. |
| Search breaking the grid | The swap emptied the root container, detaching band sections the band map still pointed at; results were appended into orphaned canvases with stale occupied cells. |
| Products ignoring band order | WooCommerce Blocks' `ProductQuery::update_rest_query` hooks `rest_product_query` at the same priority and `array_merge`s `orderby` back to the request default. Moved to 99. |
| Store panel said 120 orders, band said 155 | The panel counted processing + on-hold; the band also counts pending. The panel now reads its statuses off the band definition. |
| `myWordpressUserStats` order-dependent | Six fixtures shared a `post_date` with no tie-break, so which one fell out of the 5-item `recent` list depended on insert order. |

## Testing

- **PHP**: 1906 tests green. New coverage for post-type eligibility and grouping, the bridge controller (401 logged-out, 403 without the capability, `X-WP-Total`, `DELETE` trashes, `POST` absent, filter veto), the registration tracker, and the WooCommerce module's inert-without-WooCommerce behaviour.
- **JS**: 2774 tests green. New coverage for folder grouping, thumbnails, the preview/group/band/tile seams, and multiple handlers per payload type.
- The WooCommerce data paths were verified against a real 181-product / 675-order store via WP-CLI; the test environment has no WooCommerce, so its PHPUnit coverage is the inert path plus the pure helpers.

## Docs

`hooks-reference.md`, `javascript-reference.md`, `architecture.md`, `plugin-compat-layer.md`, plus two new examples: [custom post types in the site window](docs/examples/my-wordpress-cpt-section.md) and [accepting drops on your desktop icon](docs/examples/tile-drop-handler.md).

## Notes for review

- `includes/core/payload.php` is shared with the dock. The tracker fix is contained and strictly improves attribution, but it's the one change here that touches a file outside the my-wordpress module.
- Product category ordering applies up to 20,000 products; past that it falls back to stock-status ordering and drops the category bands rather than let them churn.
- Coupon "discount given" scans the last 500 paid orders — bounded, so an approximation on a very large store.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fdesktop-mode%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fdesktop-mode%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-472-11b39f7594c969f17ffcfc25e010d4f9728973c4.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22git%3Adirectory%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fdesktop-mode%22%2C%22ref%22%3A%22trunk%22%2C%22refType%22%3A%22branch%22%2C%22path%22%3A%22extensions%2Fdesktop-mode-popup-siege%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%2C%22targetFolderName%22%3A%22desktop-mode-popup-siege%22%7D%7D%2C%7B%22step%22%3A%22runPHP%22%2C%22code%22%3A%22%3C%3Fphp%5Cnrequire_once%20&#039;%2Fwordpress%2Fwp-load.php&#039;%3B%5Cn%24options%20%3D%20get_option(%20&#039;desktop_mode_extended_options&#039;%2C%20array()%20)%3B%5Cn%24options%20%3D%20is_array(%20%24options%20)%20%3F%20%24options%20%3A%20array()%3B%5Cn%24options%5B&#039;games&#039;%5D%20%3D%20true%3B%5Cnupdate_option(%20&#039;desktop_mode_extended_options&#039;%2C%20%24options%2C%20false%20)%3B%5Cnupdate_option(%20&#039;blogname&#039;%2C%20&#039;WP%20Desktop%20Mode%20%E2%80%94%20Popup%20Siege%20Demo&#039;%20)%3B%22%7D%2C%7B%22step%22%3A%22writeFile%22%2C%22path%22%3A%22%2Fwordpress%2Fwp-content%2Fmu-plugins%2Fpopup-siege-demo.php%22%2C%22data%22%3A%7B%22resource%22%3A%22literal%22%2C%22name%22%3A%22popup-siege-demo.php%22%2C%22contents%22%3A%22%3C%3Fphp%5Cn%2F**%20Automatically%20stage%20Popup%20Siege%20on%20the%20disposable%20Playground%20demo.%20*%2F%5Cnadd_action(%5Cn%5Ct&#039;admin_enqueue_scripts&#039;%2C%5Cn%5Ctstatic%20function%20()%20%7B%5Cn%5Ct%5Ctif%20(%20!%20function_exists(%20&#039;desktop_mode_is_enabled&#039;%20)%20%7C%7C%20!%20desktop_mode_is_enabled()%20)%20%7B%5Cn%5Ct%5Ct%5Ctreturn%3B%5Cn%5Ct%5Ct%7D%5Cn%5Ct%5Ctwp_add_inline_script(%5Cn%5Ct%5Ct%5Ct&#039;desktop-mode&#039;%2C%5Cn%5Ct%5Ct%5Ct&#039;window.wp.desktop.ready(function()%7Bvar%20games%3Dwindow.wp.desktop.games%3Bif(games%26%26typeof%20games.launch%3D%3D%3D%5C%22function%5C%22)%7BPromise.resolve(games.launch(%5C%22popup-siege%5C%22)).catch(function()%7Bwindow.wp.desktop.openWindow(%5C%22desktop-mode-games%5C%22%2C%7Bsource%3A%5C%22popup-siege-demo%5C%22%7D)%3B%7D)%3B%7Delse%7Bwindow.wp.desktop.openWindow(%5C%22desktop-mode-games%5C%22%2C%7Bsource%3A%5C%22popup-siege-demo%5C%22%7D)%3B%7D%7D)%3B&#039;%2C%5Cn%5Ct%5Ct%5Ct&#039;after&#039;%5Cn%5Ct%5Ct)%3B%5Cn%5Ct%7D%2C%5Cn%5Ct100%5Cn)%3B%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
@claude can you review it?

AllTerrainDeveloper:
Thanks — this was a genuinely useful review. All four main points were real; fixed in 11b39f7, along with every smaller item. Details below.

## 1. Query overrides were global — correct, and the worst of the four

You're right that this is the same collision the priority-99 fix was written to solve, just pointed at a new victim. A Product Collection block sorting by price would have silently got band order instead.

Section list requests now carry a `desktop_mode_bands=1` marker, declared as `listQuery` on the section descriptor and sent by `fetchEntityList()`; both filters no-op without it. I went with a request param over `get_route()` because the coupon collection lives under `desktop-mode/v1` while products are on `wp/v2` — one check covers both, and it generalises to any section that needs server-side query scoping.

Verified against a real store:

```
site window (banded):  accessories, accessories, accessories, accessories, accessories
other REST caller:     boots, pantry, pantry, outdoors, hiking
```

## 2. `getGroups()` dropped client-registered groups — correct

The docblock described merge behaviour the code didn't implement. Server and derived groups are now merged **per group**, not all-or-nothing: anything an entity references that the server didn't ship gets derived and slotted by its `groupOrder`. Server groups keep the order PHP gave them, so the `desktop_mode_my_wordpress_post_type_groups` filter's ordering isn't undone by a client-side re-sort.

## 3. Backtrace on every request — correct, and an unintentional regression

Exactly as you describe: the old `get_plugins()` guard was doing perf work by accident, and removing it turned a front-end no-op into a real cost on any site with CPTs. Now gated on `is_admin()` behind a new `desktop_mode_track_type_registrants` filter, with the reasoning in the docblock so the next person doesn't remove it again. Only admin surfaces read the map, so nothing loses attribution.

## 4. Panel could hang on a malformed response — correct

`paintPanel()` now has a rejection path, and the row builders read defensively. Worth stating plainly: `desktop_mode_my_wordpress_woo_summary` is documented as a third-party filter over that exact payload, so a plugin dropping a field is *expected* input, not a corrupt-server edge case. There's a test for it.

## Smaller items

- **Uncached discount scan** — cached, 5 minutes, matching the sibling plan queries.
- **Component reuse** — `pill()` is now `<wpd-badge>`; deleted the CSS it duplicated. I left `panel`/`row` as-is for now: `<wpd-panel>`/`<wpd-row>` are card and list-row primitives, and this is a `<dl>` of label/value pairs inside a preview pane. Happy to convert if you disagree — say the word and I'll do it in a follow-up rather than expand this PR further.
- **Test gap (no JS test imported the bundle)** — correct, and it's why the `wc_get_product_type_label()` typo reached the browser. Added `tests/vitest/my-wordpress-woocommerce.test.ts`: 16 tests over band resolution, the four ribbon states, ribbon survival across a tile repaint, and the panel's failure paths including the malformed payload above.
- **Test gap (route registration vs. permission callback)** — correct, the description overstated it. Added a test asserting the routes aren't registered at all for logged-out and subscriber users. The 401/403 tests stay as defence-in-depth, since routes are registered per-request.

**Now:** 1911 PHP tests, 2790 JS tests, lint and typecheck clean.

PR #482, opened 2026-08-04
Hide desktop icons and files layer during Overview

### LINKED ISSUE #453: Tiles (desktop icons) remain visible during Overview
## Observation

When entering **Overview**, window thumbnails are displayed against a dimmed backdrop. However, desktop shortcut icons (`.desktop-mode-icons`) and files-layer tiles (`wpd-tile` / `.desktop-mode-files-layer`) remain fully visible and interactive alongside them.

With many open windows this is easy to miss because the thumbnails cover most of the desktop. But with only a few windows (for example, two), the uncovered area exposes desktop tiles, and the left window thumbnail can overlap them, making the UI appear visually broken rather than intentionally designed. (in Classic Desktop layout)

<p align="center">
  Classic Desktop Overview showing desktop tiles visible
</p>

## Why Spatial layout makes this more noticeable

The issue is more apparent in the **Spatial** desktop layout because:

* Core dock items are synthesized as desktop shortcuts on the wallpaper surface.
* User-pinned shortcuts also remain visible.
* The files layer contributes additional `wpd-tile` elements.

As a result:

* There are significantly more desktop tiles than in the Classic or Unified layouts.
* The likelihood of window thumbnails overlapping desktop tiles during Overview increases substantially.

<p align="center">
  Spatial Desktop Overview showing more tile overlap
</p>

## Evidence that this may be unintentional

Every other non-window surface is explicitly hidden when entering Overview:

| Surface                   | How it's hidden                                                        |
| ------------------------- | ---------------------------------------------------------------------- |
| Widgets                   | `opacity: 0; pointer-events: none` (`desktop.css`)                     |
| Sticky notes              | `opacity: 0; pointer-events: none` (`desktop.css`)                     |
| Notes                     | `display: none` (`notes.css`)                                          |
| Dock (all placements)     | `width: 0; opacity: 0` (`window-overview.css`)                         |
| Admin menu, dock, widgets | Set `inert` via JavaScript (`overview.ts` → `OVERVIEW_INERT_ELEMENTS`) |

Desktop shortcut icons and the files layer are the only desktop chrome surfaces that are not covered by either the CSS rules or the `inert` list.

That makes this appear more like an oversight than an intentional design decision.

## Questions

1. **Should desktop shortcut icons and `wpd-tile` elements remain visible during Overview?**

   * The existing pattern suggests **no** - Overview appears intended to be a chrome-free task-switching experience.

2. **Is there a use case for keeping desktop tiles visible during Overview?**

Happy to open a PR once there's consensus on the expected behavior.

KarunyaChavan:
This PR fixes desktop shortcut icons (wpd-tile elements and legacy icons) remaining visible and overlapping window thumbnails when the user enters the Overview (zoom-out grid) view. The icons and files layer now fade out during Overview, matching how widgets, sticky notes, and the dock already behave.

Closes #453 

## Why?
Overview is a task-switching view - its job is to show only open windows so the user can pick one. 
But the desktop icons and file tiles stayed visible at full opacity beside the thumbnails, and with few open windows (e.g. 2), the left thumbnail could visually collide with a tile, which looked broken. 
The Spatial desktop layout made this worst, since it renders core menu items as desktop icons on the wallpaper, producing more tiles to collide with. 
Every other non-window surface (widgets, notes, dock) already hides during Overview, so leaving icons and tiles visible was an inconsistency, not a deliberate choice.

## What Changed?

| **Before**                                                                                                             | **After**                                                                                                             |
| ---------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------- |
| Before | After |

- Added a CSS rule in `assets/css/window-overview.css` that fades .desktop-mode-icons and .desktop-mode-files-layer to opacity: 0 with pointer-events: none (0.22s ease transition) when .desktop-mode-area--overview is active, matching the existing widget/sticky-note fade pattern.
- Scoped the selector with the direct-child combinator (>) so files layers mounted inside folder-native-window bodies are excluded - folder thumbnails in the grid keep their tiles visible.
- Extended the prefers-reduced-motion block to collapse the new transition to 0.01ms for accessibility.

### DIFF
--- a/assets/css/window-overview.css
+++ b/assets/css/window-overview.css
@@ -160,6 +160,23 @@
 	}
 }
 
+/*
+ * Desktop shortcut icons & files-layer tiles —
+ * hidden during overview so they don't compete with window
+ * thumbnails or overlap them (especially noticeable with few
+ * windows or Spatial layout). Matches the fade-out pattern
+ * used by widgets and sticky notes. Scoped to DIRECT children
+ * of the desktop area only: folder windows mount their own
+ * files layer inside the window body, and that must stay
+ * visible in its thumbnail.
+ */
+.os-area--overview > .os-icons,
+.os-area--overview > .os-files-layer {
+	opacity: 0;
+	pointer-events: none;
+	transition: opacity 0.22s ease;
+}
+
 /* ---------------------------------------------------------------
  * Overview thumbnail labels.
  *
@@ -268,6 +285,10 @@
 @media ( prefers-reduced-motion: reduce ) {
 	.os-area--overview,
 	.os-window--overview {
+		transition-duration: 0.01ms;
+	}
+	.os-area--overview > .os-icons,
+	.os-area--overview > .os-files-layer {
 		transition-duration: 0.01ms;
 	}
 }


assets/css/window-overview.css
```diff
@@ -155,11 +147,28 @@
...
+ * files layer inside the window body, and that must stay
+ * visible in its thumbnail.
+ */
+.desktop-mode-area--overview > .desktop-mode-icons,
+.desktop-mode-area--overview > .desktop-mode-files-layer {
+	opacity: 0;
+	pointer-events: none;
+	transition: opacity 0.22s ease;
+}
+
```
AllTerrainDeveloper, line 179: Hey! @KarunyaChavan The code needs to be updated. There is no reference at all in the newrebranding for "desktop-mode" in CSS files. Can you update your fork?

assets/css/window-overview.css
```diff
@@ -155,11 +147,28 @@
 }
 
 @media ( prefers-reduced-motion: reduce ) {
-	.os-dock {
+	.desktop-mode-dock {
 		transition-duration: 0.01ms;
 	}
 }
```
AllTerrainDeveloper, line 162: Image

AllTerrainDeveloper (changes requested):
Hey! Thank you! Sadly we do a massive rework yesterday that has caused your fork to be staled. I tested it and it works but once merged will not :)

KarunyaChavan:
I rewrote the branch onto the rebranded trunk as a single commit so the PR shows the fix with zero merge noise. I've verified that changes are now working, thank you! 🙇

AllTerrainDeveloper (approved):
Works as expected!

https://github.com/user-attachments/assets/42e78ec0-1c94-4403-b601-dbc7bfba22fe

Thank you very much!

PR #485, opened 2026-08-04
Woo: a Customers section, and relations that run both ways

AllTerrainDeveloper:
Adds a **Customers** section to the Woo folder, a **Customer window**, and the reverse half of the relations graph — plus the framework seams those needed, and three bugs found on the way.

## Customers

Renders through the built-in `user` entity kind — avatar tiles, drag-out, the dossier pane — with lifetime spend, order count, AOV, first/last order and a band (VIP → lapsed → repeat → new → none, the two actionable ones first) on every row. VIP is derived (3× store AOV) rather than fixed, because "spent over 500" means nothing without knowing whether the store sells postcards or pianos.

The whole section costs **one grouped query** over the order store (HPOS or legacy), cached and flushed on any order change; band ordering, per-row facts and folder counts all read it, so a page of customers fires no per-row order queries. The facts field is registered on the core `user` REST resource, so the built-in Users section gets them for free.

Double-clicking a customer opens the **Customer window**: identity, four stat cards, what they buy most, recent orders (each opening in its own window), addresses.

## Relations, both directions

An order announces `links` to its customer, products and coupons. The new half: a **product** reaches its orders, its buyers and the coupons that discount it; a **coupon** reaches the orders that redeemed it and who redeemed them. An order names its products, so that direction always worked — *is this selling, and to whom?* is the question a merchant actually asks, and the catalogue screen had no answer.

Reviews and profile screens gained content identities, so those windows tie rather than opening unconnected.

## Framework changes (all generic)

- **`WindowConfig.params`** — open-time arguments for a native window, persisted with the session and staged back on restore. A native window is addressed by id, so a singleton that retargets had nowhere to record its subject; the profile window came back showing whoever was logged in after F5.
- **`os.my-wordpress.user-dossier-sections`**, **`…user-preview-actions`**, **`…user-activate`** — a user pane is not always an author pane. Post counts and a publishing sparkline are four zeroes above the figure a merchant came for.
- `preview-extras` and `list-tile` now fire for **user** tiles; the user pane gained a `meta` slot under the identity header.
- URL remaps can declare a **`params`** hook, and **`os_person_view`** lets a specific view claim a person-URL without racing the built-in profile remap.

## Bugs fixed

- **`loadVendorScript` deduped only against tags it had injected**, so naming an already-enqueued handle as a native window's `script` evaluated the bundle twice. `@wordpress/hooks` appends on a repeated namespace, so every subscriber ran twice — two Store panels, two badges. Nothing in the symptom pointed at script loading.
- **Panel links carried `target="_blank"`.** In a tab that threw admin screens out of the shell; inside the installed PWA a same-origin admin URL is in scope, so clicking a product name **relaunched the app**.
- **The session sanitizer dropped unknown window fields**, so `params` never survived the POST — and `sanitize_key()` would have lowercased camelCase names, storing `customerid` for a client reading `customerId`.
- `openAdminWindow` called `manager.open()` with no `id`, which throws.

## Tests

1998 PHPUnit, 3717 Vitest, all green. New: `myWordpressWoocommerceCustomers.php`, `window-manager-params.test.ts`, `vendor-loader.test.ts`, plus additions to the session, window-links, remap and Woo suites.

## Known gap

The Customers grid is band-*ordered* but not band-*headed* — that chrome lives in the post-kind list renderer and the user-kind one doesn't have it yet. The order is right; the band shows as a tile badge instead. Noted in `docs/plugin-compat-layer.md`.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-485-ddcac09570657c88f3a59bf8f24595a6cc50cd31.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/assets/css/my-wordpress-woocommerce.css
+++ b/assets/css/my-wordpress-woocommerce.css
@@ -204,3 +204,308 @@ os-ribbon.os-woo-ribbon {
 	font-size: 20px;
 	font-weight: 600;
 }
+
+/* ----- Customer tiles ----- */
+
+/*
+ * The band badge, for VIP and lapsed only.
+ *
+ * It lives INSIDE the 48px avatar box, straddling its bottom edge —
+ * the way a status dot sits on a contact photo. That costs the tile
+ * no vertical space, so the icon stays an icon: a face, a name, and
+ * at most one mark. Everything else about a customer (spend, orders,
+ * last purchase) is one click away in the pane, which has room to say
+ * it properly.
+ *
+ * Two rejected alternatives, both tried: `<os-ribbon>`, the 45°
+ * corner banner the Products grid uses — right across a product
+ * photo, vandalism across a face, covering a third of the avatar at
+ * this size. And a line of text under the name — spend and order
+ * count stacked above a 48px avatar crowd the tile past reading, and
+ * they push the name down the grid.
+ */
+.os-woo-customer-avatar {
+	position: relative;
+	/* The badge is wider than 48px and must not be clipped. */
+	overflow: visible;
+}
+
+.os-woo-customer-band {
+	position: absolute;
+	inset-block-end: -4px;
+	inset-inline-start: 50%;
+	transform: translateX( -50% );
+	z-index: 1;
+	padding: 0 5px;
+	border-radius: 999px;
+	font-size: 8px;
+	font-weight: 700;
+	line-height: 14px;
+	letter-spacing: 0.04em;
+	text-transform: uppercase;
+	white-space: nowrap;
+	pointer-events: none;
+}
+
+/*
+ * A solid surface under the tone, because this sits ON a photograph:
+ * the 12%-alpha wash `<os-badge>` uses would take whatever colour the
+ * avatar happens to be. The tone survives as the text and the ring,
+ * read from the same tokens (and the same fallback literals) the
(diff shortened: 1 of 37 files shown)


AllTerrainDeveloper:
@claude can you review?

AllTerrainDeveloper:
@claude one more review?

PR #488, opened 2026-08-04
Rebrand: Polish the leftovers, and announce the rename

mmtr:
## Proposed changes

Four fixes and one addition left over from the rebrand.

**Tab strip edge gradients.** The submenu tab strip masked both edges unconditionally. They now paint only on an edge that is actually hiding a tab, re-measured on scroll, on resize, and when tabs are added or removed.

Before | After
--- | ---
Screenshot 2026-08-04 at 18 24 29 | Screenshot 2026-08-04 at 18 24 47

**Notice link colour.** Links inside `<os-notice>` (for example the `Settings > Connectors` link in OpenStation Settings > Features) painted WordPress Blue at rest and `#135e96` on hover, close to illegible on the notice wash. They now read Nebula, lifting to Starlight.

Before | After
--- | ---
Screenshot 2026-08-04 at 17 14 44 | Screenshot 2026-08-04 at 18 25 23

**OpenStation Settings.** The `OS Settings` tile and window are renamed, and the tile wears the logomark instead of `dashicons-desktop`. Renamed across user-facing strings and docs; code comments and test fixtures still say "OS Settings" (around 380 occurrences), left for a separate mechanical sweep so this diff stays readable.

Before | After
--- | ---
Screenshot 2026-08-04 at 18 24 17 | Screenshot 2026-08-04 at 18 23 53

**Rename announcement.** A one-off dialog explaining that Desktop Mode is now OpenStation, shown once to each user who was actually using the plugin before the rename. Fresh installs and users who joined afterwards never see it. It only exists inside the shell, so it cannot appear in the classic admin or in a window iframe. New filter `openstation_install_predates_rebrand` suppresses it site-wide.

Screenshot 2026-08-04 at 18 23 28

**`bin/setup-wp-env.sh`.** Activates the right plugin slug again.

The first three are the rebrand landing unevenly: cosmetics that were invisible on the old light chrome became visible on the dark one, and a colour that was tied to the user's accent picker stopped being legible once the surfaces went dark.

Two of them had a cause worth writing down, because in both cases the obvious fix does nothing:

- A slotted link belongs to the **document** tree, not the shadow tree that renders it. CSS Scoping hands normal declarations to the outer tree on a collision, so `wp-admin/css/common.css` `a { color: #2271b1 }` beats `::slotted( a )` regardless of specificity. Setting the component's token alone had no effect; the declaration had to move to a tree that can win.
- The dock paints an icon's **alpha**, not its colours (`_makeSvgIcon()` masks every image icon with `currentColor` so plugin brand colours cannot break the monochrome rail). The brand's app chip is a mark knocked out of a tile, so masking one yields the tile: a solid white rounded square. The icon is a `currentColor` silhouette instead, which both painters handle correctly and which stays legible on a light title bar without carrying its own ground.

The announcement exists because someone who installed under the old name opens wp-admin to a differently named, differently coloured shell. Without a word of explanation that reads as a compromised site rather than a release.

The `setup-wp-env.sh` fix is a slug the rename sweep caught by accident, even though the rebrand commit listed it as not-to-be-renamed. The script runs under `set -euo pipefail`, so the failed activate aborted it: the Gutenberg Guidelines experiment never got enabled and the plugin was left **inactive** on every fresh `wp-env start`.

### Tab strip gradients

1. Open a window with a submenu that fits, e.g. **Settings**.
2. Make sure there is no grey fade at either end of the tab strip.
3. Narrow the window until the tabs overflow.
4. Make sure a fade appears on the right only.
5. Scroll the strip to the middle. Make sure both ends fade.
6. Scroll it fully right. Make sure only the left end fades.
7. Widen the window until the tabs fit again. Make sure both fades disappear.

### Notice link colour

1. Go to **OpenStation Settings > Features** on a site with no AI provider configured.
2. Find the "This feature requires an AI provider configured in Settings > Connectors" notice.
3. Make sure the link is pink (Nebula), not blue, and that hovering it turns it near-white rather than a darker blue.
4. Go to **OpenStation Settings > Appearance**, switch the accent to WordPress Blue, and return to Features. Make sure the link colour did not change, since it no longer follows the accent.

### OpenStation Settings tile

1. Make sure the dock tile and its window title both read "OpenStation Settings".
2. Make sure the tile shows the OpenStation logomark (a circle with a blob cut), the same weight as the dashicons beside it, and not a filled square.
3. Open the window and make sure the title bar shows the same mark.

### Rename announcement

The announcement only fires for a user who was using the plugin before the rename, so a fresh dev site will not show it. To simulate one:

```
wp user meta update admin desktop_mode_mode 1
wp eval 'openstation_migrate_flag_rebrand_notice( 0 );'
wp user meta delete admin desktop_mode_seen_intros
```

To reset between runs, `wp user meta delete admin desktop_mode_seen_intros` alone is enough.

1. Reload the shell. Make sure the dialog appears about a second after the desk settles, not mid-paint.
2. Make sure it is styled like the first-run welcome card: blurred backdrop, white card, dark shimmering hero with the logomark.
3. Press <kbd>Esc</kbd>. Make sure it closes and focus returns to where it was.
4. Reload. Make sure it does **not** come back.
5. Run the `wp user meta delete` line again, reload, and dismiss with **Got it** instead. Reload and make sure it stays gone.
6. Repeat once more and dismiss by clicking the backdrop, then the × chip. Both should stick.
7. Click inside the card (for example select some text). Make sure it does **not** close.
8. Go to **OpenStation Settings > Features** and hit "Reset what's-new dialogs". Reload and make sure the announcement is back.
9. On a second user account that was also flagged, make sure the announcement still appears after the first user dismissed it.
10. Create a fresh user, enable OpenStation for them, and make sure they never see it. They joined after the rename, so there is nothing to explain.
11. Switch to the classic admin (`?desktop_mode_classic=1`) and make sure the dialog never appears there, and that `announce.css` is not among the loaded stylesheets.
12. On a user who is not flagged, view source and make sure `announce.css` is not enqueued at all.

### `bin/setup-wp-env.sh`

1. Run `npm run env:start` on a fresh instance, or reset an existing one:

```
wp plugin deactivate desktop-mode gutenberg
wp option delete gutenberg-experiments
./bin/setup-wp-env.sh
```

2. Make sure the script exits 0 with no "plugin could not be found" error.
3. Run `wp plugin list` and make sure both `desktop-mode` and `gutenberg` are active.
4. Run `wp option get gutenberg-experiments` and make sure it contains `gutenberg-guidelines => true`.

### Regression sweep

- `npm run test:js`, `npm run lint`, `npm run typecheck`
- `npm run lint:php`, `npm run test:php`

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-488-590ee51781842a4dda2fe4ba16303d53ae9b1042.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper (approved):
Thank you! Super valuable!

PR #501, opened 2026-08-05
Selection: hold more than one thing at a time

AllTerrainDeveloper:
https://github.com/user-attachments/assets/beb7c0dc-272f-4d4f-907a-b242fee3c938

Every tile surface in the shell was single-select, and each one had re-implemented it — six selection states and four near-identical context-menu builders between the wallpaper, folder windows and the five canvases in the site folder. You could not act on two things at once anywhere.

This is one selection framework (`src/selection/`) that all of them use, and one rule for the hard part.

## The rule: who decides what a mixed set can do

- **The item's own adapter** answers *"what can be done to this one thing?"* — that's the tile-menu builder each surface already had, filters and all. Unchanged.
- **The framework** answers *"what do these N have in common?"* — intersect by action id, and keep an id only when **every** contributing action declares `multi: true`.

So selecting a post and an image leaves *Move to Trash* and drops *Navigate into* (post-only) and *Download* (image-only).

`multi` is opt-in on purpose: an action written against one item — *Rename…*, *Share folder…* — would misbehave run twelve times, and its author never agreed to that. **Existing menu entries keep working untouched** and simply stay single-item until they opt in. `multiId` merges the same deed under different labels, which is how a folder's `delete-folder` and a file's `remove` end up sharing one Trash entry.

Actions can ship a `bulk( items )` runner. That's what turns twelve REST calls, twelve toasts and twelve Undo buttons into one of each.

## What you get

**Selection** on every canvas — wallpaper, folder windows, and every My WordPress list: click, Ctrl/Cmd-click, Shift-click, rubber band, Cmd+A, Escape, arrow keys. Status bars gain a live "N selected"; preview panes summarise a set instead of previewing an arbitrary member.

**WordPress's own bulk actions** on the site folder's lists:

| Section | Actions |
|---|---|
| Posts / Pages / CPTs | **Edit…** (the bulk-edit modal), Publish, Switch to Draft, Copy link, Move to Trash |
| Media | Detach, Copy file URL, Delete permanently |
| Users | Change role…, Delete… , Copy author archive link |

Bulk edit matches core where it counts: **only the fields you changed are written** (every control starts on "— No change —"), and **categories and tags are added** to what each entry already has rather than replacing it. Deleting users asks core's "delete their content or reassign it" question first, because the REST endpoint refuses without an answer.

The status actions are the intersection rule in the wild: *Publish* only appears for entries that aren't published, *Switch to Draft* only for those that aren't drafts — select one of each and neither shows.

**Multi-item drag.** Three icons to the Trash is one toast with one Undo; three to a folder is one move; three posts from My WordPress onto the wallpaper is three shortcuts. The ghost is a stack with a count badge and the hover chip says the count.

## Compatibility

Drag payloads grew **one optional field per type** — `placements` on `desktop-file`, `items` on `shortcut` — both **absent for a single-item drag**. Every drop target written before this, ours and any plugin's, is untouched and correct: it reads `data.placement` and acts on the tile the user pointed at. Targets that want sets call `dragPlacements()` / `dragShortcutItems()`, which fall back to the grabbed item — one code path for "one" and "many".

Targets that do support sets apply their accept gates to **every** member and refuse the whole drop when any one fails. Accepting a set and handling part of it reports success for an operation that half-happened.

## Also: one icon grid

The grid had drifted into two — the desktop ran a 96×110 pitch and the site folder 108×112 — and the desktop's 8px of air was fiction: the tile had no `box-sizing`, so its padding fell outside its declared width, filled the cell, and icons touched.

There is now one grid, declared as tokens in `variables.css` (so a desktop theme can retune it) and mirrored in `grid.ts` for the layout maths, with a test that parses the stylesheet to prove the mirror stays faithful. **The cell is derived, never declared**: `cell = tile + gap`. Tile height is fixed rather than minimum, because the tile box *is* the selection ring and a box that grows with its label gives a row of selected icons a ragged edge.

> **Heads-up for reviewers:** stored icon coordinates were on the old pitch, so expect a one-time reshuffle. Icons snap to the new grid on the next drag or *Sort by → Clean up*.

## Four bugs found while building it

Each has the test that would have caught it:

- **`<os-tile>` rebuilt its children on any attribute change.** A selection change mid-gesture destroyed the node a `mousedown` landed on, so the browser synthesized no `click` — no `dblclick`, and tiles stopped opening. Selection attributes now take a children-preserving path (also: Cmd+A over 200 icons costs 200 class toggles, not 200 subtree rebuilds).
- **The marquee's `.os-window` exclusion** — added so a title-bar drag can't rubber-band the wallpaper behind it — also matched every canvas *inside* a window, where the window is an ancestor rather than a descendant. Scoped with `background.contains()`.
- **The marquee anchor was in viewport coordinates**, so scrolling mid-drag froze the band. It's content-space now, redraws on scroll, and auto-scrolls at the edges.
- **A marquee released over an iframe stranded the gesture**, because the `pointerup` fired in that document. The band takes pointer capture, with `lostpointercapture` and window `blur` as nets. It also suppresses native text selection while it runs — the bare canvas is selectable where a tile is not, so the browser was quietly running its own selection into whatever the pointer crossed.

## Testing

`npm run lint`, `npm run typecheck`, `npm run test:js` (**3904 tests, 317 files**) and `npm run build` all green. **9 new test files**: the selection model, the action resolver, the controller (gestures, marquee, scroll, capture, suppression), both surfaces, bulk trash, bulk actions, multi-drag, and grid metrics.

No PHP changed — this is entirely client-side over existing REST routes — so `lint:php` / `test:php` are unaffected.

Manual QA on a live site: multi-select and bulk actions across Posts, Users, Media and folder windows; drag-to-Trash and drag-to-folder with Undo; rubber band on the wallpaper, in folder windows, and mid-scroll.

## Docs

Updated in the same change: `javascript-reference.md` (the `wp.os.selection` surface, `os-selection-changed`, the `multi` / `multiId` / `bulkLabel` / `bulk` fields, the built-in bulk-action table, multi-item drag payloads), `files-on-desktop.md` (selection, dragging a selection, the grid), `components-reference.md`, `api-index.md`, `desktop-themes.md` (the grid tokens), and a new `examples/multi-selection-action.md`.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-501-7ab895cfd0f41d75d0dbe2cb085dc47184efe4d7.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/assets/css/desktop-files.css
+++ b/assets/css/desktop-files.css
@@ -152,7 +152,25 @@
 	align-items: center;
 	justify-content: flex-start;
 	gap: 6px;
-	width: 88px;
+	/*
+	 * `box-sizing` is load-bearing, not housekeeping. Without it the
+	 * horizontal padding is ADDED to the declared width, so an 88px
+	 * tile occupies 96px — which was exactly the desktop's old cell
+	 * pitch, and why icons sat edge to edge with no gap at all while
+	 * the maths insisted there were 8px between them.
+	 *
+	 * The tile is the size the grid says it is. Padding lives inside.
+	 */
+	box-sizing: border-box;
+	width: var( --os-tile-w, 88px );
+	/*
+	 * Fixed, not minimum. The tile box IS the selection ring, and a
+	 * box that grows with its label gives a row of selected icons a
+	 * ragged top edge — one height per label that happened to wrap
+	 * to two lines. Every tile occupies its cell; short labels leave
+	 * the slack inside the ring, where it reads as padding.
+	 */
+	height: var( --os-tile-h, 104px );
 	padding: 8px 4px;
 	border: 0;
 	background: transparent;
@@ -754,6 +772,58 @@
 	--os-tile-selected-bg: rgba( 34, 113, 177, 0.12 );
 }
 
+/* Type breakdown under a multi-selection's count in a folder
+ * window's preview pane. The count is the headline; this is the
+ * detail that tells you which actions the menu will offer. */
+.os-files-preview__selection-breakdown {
+	font-size: 12px;
+	opacity: 0.75;
+}
+
+/* While a rubber band is live, nothing in the shell is selectable.
+ *
+ * The band starts on bare canvas, which — unlike a tile — IS
+ * selectable, so the browser begins its own text selection on the
+ * same press and keeps extending it as the pointer travels. Inside
+ * the canvas that stays invisible; drag out over another window and
+ * it starts highlighting that window's text in blue, mid-gesture.
+ *
+ * The controller also refuses `selectstart` outright while the band
+ * is up. This rule is the second half: it covers anything already
+ * selectable that the event route misses. */
+body[ data-os-marquee ] {
+	-webkit-user-select: none;
+	user-select: none;
+}
+
+/* Marquee (rubber-band) selection box. Painted by the shared
+ * selection controller on any canvas that opts into it — the
+ * wallpaper, folder windows, every My WordPress list — so this one
+ * declaration dresses all of them.
+ *
+ * `pointer-events: none` is load-bearing: the box tracks under the
+ * cursor for the whole gesture, and without it every hit-test
+ * (drop targets, the controller's own tile lookup) would land on
+ * the box instead of what's beneath it. */
+.os-selection-marquee {
+	position: absolute;
+	z-index: 5;
+	pointer-events: none;
+	border: 1px solid
+		var( --os-tile-focus-ring, var( --wp-admin-theme-color, #2271b1 ) );
+	border-radius: 2px;
+	/* A WASH, not a fill — the whole point of a rubber band is that
+	 * you can see what it is about to catch. `--os-ui-accent-dim` is a
+	 * solid hex (Pulse, one step back), so reaching for it directly
+	 * paints an opaque rectangle over the icons; `--os-ui-accent-soft`
+	 * is that same colour already taken down to a 14% wash, which is
+	 * what every other ambient accent surface in the shell uses. */
+	background: var(
+		--os-selection-marquee-bg,
+		var( --os-ui-accent-soft, rgba( 34, 113, 177, 0.14 ) )
+	);
+}
+
 .os-folder-status-bar {
 	display: flex;
 	justify-content: space-between;
@@ -961,7 +1031,10 @@ button.os-folder-status-bar__segment:hover {
 .os-file-tile__label {
 	font-size: 12px;
 	line-height: 1.2;
-	max-width: 84px;
+	/* Follows the tile rather than restating its width, so a section
+	 * that re-points `--os-tile-w` (image-led sections do) gets a
+	 * label that grows with the tile instead of staying narrow. */
+	max-width: calc( var( --os-tile-w, 88px ) - 4px );
 	text-align: center;
 	overflow: hidden;
 	text-overflow: ellipsis;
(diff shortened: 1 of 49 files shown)


AllTerrainDeveloper:
@claude can you review?

AllTerrainDeveloper:
All six addressed in 7ab895cf, plus one more found in testing. Every fix has the test that would have caught it; 3916 tests green.

**Plus: the marquee scrolled the wallpaper.** Dragging a band to the bottom of the screen slid the desktop under it. Auto-scroll only asked whether the content overflowed its box — and `.os-area` is `overflow: hidden` with icons plus 80px of bottom padding overflowing it constantly, so it scrolled a surface with no scrollbar that nothing else can scroll. It now asks whether the box is a scroll container at all (`overflow-y: auto | scroll | overlay`). Tests both ways: a hidden canvas is never scrolled, a real scroller still is.

---

**1. Bulk-undo broadcasting `untrashed` regardless.** Fixed — the settle results are inspected and only the ids that actually came back are announced. You're right that the single-item paths already got this right by broadcasting inside their success branch; this now matches. Added the other half too: if part of an Undo fails the user is told, rather than left believing it worked. Two tests — a partial restore announces 2 of 3 ids and surfaces a toast, a clean one announces all and says nothing.

**2. `destroy()` clearing the wrong snapshot.** Fixed, and thank you — comparing `scope` as well would still have been wrong: two folder windows open on the *same* folder share both fields. Each controller now holds an identity object and clears the snapshot only if it owns it. The test opens two controllers with identical `surface` **and** `scope` and closes the inactive one.

**3. `multiId` merge calling only the first contributor's `bulk`.** You offered "document the constraint or refuse to merge" — I went with a third option: make the framework route correctly, so there's no constraint to violate. The resolver groups the selection by `bulk` **function identity** and calls each runner once with the items whose own contributor declared it; contributors without a `bulk` still fan out through `onClick`.

That exposed a real trap in my own code: the two Trash entries were `bulk: ( p ) => trashManyWithUndo( p )` written inline at each site — two distinct closures, so a mixed folder + file selection would have become two batches, two toasts, two Undos. They now share one hoisted reference, and *that* is the documented way to opt into a single batch alongside a built-in. Four tests: different runners each get their own subset; a shared reference makes one call with the whole set; no-`bulk` contributors still fan out; a throwing batch doesn't stop the others.

Documented in `javascript-reference.md` and the example.

**4. Stale pre-modal snapshot.** Fixed — the terms are re-read after the modal closes, and only when the patch actually merges categories or tags, so the extra request is paid only where it buys something. Test simulates a collaborator adding a term while the modal sits open and asserts it survives; a second test asserts a status-only edit still makes just one read.

**5. Partial-failure toast dropping the reason.** Fixed: `"1 entry updated · 1 failed — Sorry, you are not allowed to edit this post."`

**6. Unused imports.** Removed. (Worth noting for anyone else: these are side-effect imports, so no linter was ever going to catch them — thanks for reading closely enough to.)

PR #506, opened 2026-08-06
Recent Comments: render decoded commenter names

### LINKED ISSUE #496: Recent Comments widget renders raw HTML entities in commenter names
There is a visual rendering bug in the **Recent Comments** widget where comment author names containing special characters (like ampersands) render as raw HTML entities (e.g. displaying literally as `Q&amp;A Guest` instead of `Q&A Guest`).

Additionally, this causes the dynamic initials-based avatar bubble to grab the first character of the raw string, rendering `&` instead of the actual first letter of the name (e.g. rendering `&` instead of `Q` for `Q&amp;A Guest`).

### Steps to Reproduce
1. In classic WordPress, quick-edit or add a comment where the **Author Name** contains an ampersand (e.g. `Q&A Guest`).
2. Switch to OpenStation (Desktop Mode).
3. Add the **Recent Comments** widget to the desktop.
4. Observe that the commenter's name renders as `Q&amp;A Guest` on the widget, and the avatar circle displays `&` as the initial.

### Expected Behavior
Commenter names should be decoded so they display cleanly (e.g. `Q&A Guest`) and the avatar bubble should use the correct first letter (`Q`).

### Root Cause
In `src/plugins/recent-comments-widget/index.ts`, `c.author_name` is set directly to the DOM elements using `.textContent` without running it through the imported `decodeHTML` utility:

```ts
// src/plugins/recent-comments-widget/index.ts
avatar.textContent = ( c.author_name || '?' ).trim().charAt( 0 ).toUpperCase();
...
author.textContent = c.author_name || 'Anonymous';
```

***

joenermunch:
Fixes #496.

## The bug

The WordPress REST API can return HTML-encoded comment author names. The Recent Comments widget assigned that raw value to `textContent`, so a name such as `Q&amp;A Guest` appeared with the entity still visible.

The avatar initial was derived from the same raw value, which could also select the first character of an entity instead of the commenter's initial.

## The fix

Decode each author name once with the existing `decodeHTML()` utility, then use that value for both the visible name and the avatar initial. The decoded value is still assigned through `textContent`, so it cannot be interpreted as markup.

## Tests

A new widget test mounts the real Recent Comments bundle, returns an encoded author name from the REST stub, and checks both rendered outputs:

- visible name: `Q&A Guest`
- avatar initial: `Q`

`npm run build`, `npm run lint`, `npm run test:js` (3,962 tests), and `npm run typecheck` are all green.

## To verify

1. Add or quick-edit a comment whose author name is `Q&A Guest`.
2. Open OpenStation and add the Recent Comments widget.
3. Confirm the widget shows `Q&A Guest`, not `Q&amp;A Guest`, and the avatar shows `Q`.

### DIFF
--- a/src/plugins/recent-comments-widget/index.ts
+++ b/src/plugins/recent-comments-widget/index.ts
@@ -112,10 +112,11 @@ function render( container: HTMLElement, comments: CommentRow[] | null, error: b
 	for ( const c of comments ) {
 		const row = document.createElement( 'div' );
 		row.className = 'dm-comments__row';
+		const authorName = decodeHTML( c.author_name || '' );
 
 		const avatar = document.createElement( 'div' );
 		avatar.className = 'dm-comments__avatar';
-		avatar.textContent = ( c.author_name || '?' ).trim().charAt( 0 ).toUpperCase();
+		avatar.textContent = ( authorName || '?' ).trim().charAt( 0 ).toUpperCase();
 
 		const body = document.createElement( 'div' );
 		body.className = 'dm-comments__body';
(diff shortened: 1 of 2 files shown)


AllTerrainDeveloper:
Hey @joenermunch. That was nice! I encourage you to participate in the repo, we have a beatiful hall of fame :)

image

PR #509, opened 2026-08-06
Use the theme outline token for dock status indicators

### LINKED ISSUE #508: Dock active indicator is nearly invisible
The active/focused window indicator in the dock is rendered using the surface color instead of the theme's accent color. On the default Station theme, this causes the indicator dot to blend into the dock background, making it difficult to see.

| **1. OpenStation Theme** | **2. Classic Theme** |
|---------------------------|----------------------|
| <p align="center">OpenStation Theme</p> | <p align="center">Classic Theme</p> |
| Active indicator blends into the dock and is difficult to distinguish. | Active indicator is bright and clearly visible against the dock. |

### Affected selectors

* `assets/css/dock.css:636` – `.os-dock[data-os-dock-placement="left"] .os-dock__item--active::before`
* `assets/css/dock.css:676` – `.os-dock[data-os-dock-placement="right"] .os-dock__item--active::before`
* `assets/css/dock.css:911` – `.os-dock[data-os-dock-placement="bottom"] .os-dock__item--active::before`

These rules currently use:

```css
background: var(--os-ui-surface, #fff);
```

### Current behavior

* `--os-ui-surface` is a surface/background token, not an indicator color.
* On the default Station theme it resolves to **Obsidian (`#1a1721`)**, which is nearly the same color as the dock background.
* As a result, the active indicator has very little visual contrast and is effectively invisible.
* The focused indicator inherits the same color; only its dimensions differ.

### Expected behavior

The active indicator should use the theme's accent color (for example, **Pulse `#f252fc`** on the default theme) or a dedicated dock indicator token such as:

```css
background: var(--os-dock-indicator, #fff);
```

This preserves compatibility with Legacy and custom themes while allowing branded themes to provide their own accent color.

### Evidence of inconsistency

* The "all windows minimized" indicator is rendered as a bright white ring (`rgba(255,255,255,0.85)`), making it significantly more visible than the active indicator.
* The dock already exposes an accent-aware token for keyboard focus (`--os-dock-item-outline`), demonstrating the intended theming pattern.

### Verification

Inspect the `::before` pseudo-element of an active dock item in DevTools.

* **Current:** `background: #1a1721`
* **Expected:** Theme accent color (or the appropriate dock indicator token)

## Accessibility impact

* **Fails WCAG 2.1 Success Criterion 1.4.11 (Non-text Contrast):** the indicator does not provide sufficient contrast against the dock background.
* **Weak visual hierarchy:** the minimized state is more prominent than the active state.
* **Reduced discoverability:** users relying on the dock for quick identification of the active window must instead hover items or inspect window titles.

KarunyaChavan:
This PR fixes the near-invisible dock active-window indicator.

The active status dot (and the all-minimized hollow ring) were previously painted from `--os-ui-surface`, which resolves to the dock's own dark glass tone on the default Station palette rather than the theme's intended indicator color. As a result, the active indicator had almost no visual contrast.

Both indicators now read `--os-dock-item-outline`:

* **OpenStation** uses Pulse (`#f252fc`).
* **Custom themes** use the theme's configured outline/ring color.
* **Legacy** continues to render white via its frozen manifest and remains unchanged.

Each indicator also retains its original literal fallback, ensuring an unthemed or stylesheet-free shell renders exactly as before.

A regression test now pins both the active dot and minimized ring (across all three dock placements) to the theme token.

Closes #508 

### Accessibility

* Restores the intended state hierarchy. The active ("currently in use") indicator had become the weakest visual signal on the dock (~1.06:1 against the dock glass, failing WCAG 2.1 SC 1.4.11 Non-text Contrast), while the minimized ring was more prominent.
* The active indicator is once again the strongest state marker.

### Consistency

* Aligns dock indicators with the rest of the shell.
* Focus rings, hover washes, and dock status indicators now all communicate state using the same accent/outline token, reinforcing a consistent visual language.

### Themeability

* Removes hardcoded styling.
* Pale or custom dock themes no longer end up with invisible status indicators.
* Both indicators now follow a single themeable token for the dock surface.

---

### CSS

**`assets/css/dock.css`**

* Active/focused status dot (left, right, and bottom docks) now uses:

  ```css
  var(--os-dock-item-outline, #fff)
  ```
* All-minimized hollow ring (left, right, and bottom docks) now uses:

  ```css
  var(--os-dock-item-outline, rgba(255,255,255,0.85))
  ```

### Tests

**`tests/vitest/dock-and-titlebar-glyph-tokens.test.ts`**

* Adds regression coverage ensuring:

  * the active/focused dot reads `--os-dock-item-outline` with the exact `#fff` fallback.
  * the all-minimized ring reads `--os-dock-item-outline` with the exact `rgba(255,255,255,0.85)` fallback.
  * all three dock placements are covered.

### Documentation

**`docs/desktop-themes.md`**

* Documents the dock status indicators in the dock glyph token table.
* Adds guidance for pale/custom dock themes.

### DIFF
--- a/assets/css/dock.css
+++ b/assets/css/dock.css
@@ -633,7 +633,7 @@
 	width: 4px;
 	height: 4px;
 	border-radius: 50%;
-	background: var( --os-ui-surface, #fff );
+	background: var( --os-dock-item-outline, #fff );
 }
 
 .os-dock[ data-os-dock-placement="left" ]
@@ -659,7 +659,7 @@
 	height: 6px;
 	border-radius: 50%;
 	background: transparent;
-	border: 1px solid rgba( 255, 255, 255, 0.85 );
+	border: 1px solid var( --os-dock-item-outline, rgba( 255, 255, 255, 0.85 ) );
 }
 
 /* Right dock: indicator on the opposite edge (inside-facing). */
@@ -673,7 +673,7 @@
 	width: 4px;
 	height: 4px;
 	border-radius: 50%;
-	background: var( --os-ui-surface, #fff );
+	background: var( --os-dock-item-outline, #fff );
 }
 
 .os-dock[ data-os-dock-placement="right" ]
@@ -689,7 +689,7 @@
 	height: 6px;
 	border-radius: 50%;
 	background: transparent;
-	border: 1px solid rgba( 255, 255, 255, 0.85 );
+	border: 1px solid var( --os-dock-item-outline, rgba( 255, 255, 255, 0.85 ) );
 }
 
 /* "Open another" chip — left: right of tile; right: left of tile. */
@@ -908,7 +908,7 @@
 	width: 4px;
 	height: 4px;
 	border-radius: 50%;
-	background: var( --os-ui-surface, #fff );
+	background: var( --os-dock-item-outline, #fff );
 }
 
 .os-dock[ data-os-dock-placement="bottom" ]
@@ -929,7 +929,7 @@
 	height: 6px;
 	border-radius: 50%;
 	background: transparent;
-	border: 1px solid rgba( 255, 255, 255, 0.85 );
+	border: 1px solid var( --os-dock-item-outline, rgba( 255, 255, 255, 0.85 ) );
 }
 
 /*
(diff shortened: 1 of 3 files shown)


AllTerrainDeveloper (approved):
This is great! Thank you!

Image

PR #521, opened 2026-08-07
Docs: Fill the filter reference gaps, and drop the brand-map filter

mmtr:
## Proposed Changes

* Documents three filters that were fired but never reached `docs/hooks-reference.md`: `openstation_brand_migration_map`, `openstation_track_type_registrants` and `openstation_my_wordpress_woo_coupon_bands`.
* Adds an index table at the head of the WooCommerce section listing all nineteen `openstation_my_wordpress_woo_*` filters and what each one shapes.
* Adds a WooCommerce block to `docs/api-index.md` pointing at the filters, the REST routes, the `openstation_woo_customer` REST field and the Customer window.

Docs only, no code changes.

Because the three filters carry full `@param` docblocks, so they read as public API, but a plugin author had no way to find them.

The rest of the WooCommerce filters were already written up, but only as bare signatures grouped under one heading, with no backticked hook name anywhere. Searching the doc for a hook name returned nothing, and there was no anchor to link to. The index table fixes both without rewriting the prose that was already there.

## Testing Instructions

This is a docs-only change, so the check is that nothing fired is missing from the contract.

1. Check out the branch and run:

```bash
git grep -h -A1 -E "(apply_filters|do_action)\(" HEAD | grep -oE "'(openstation)_[a-z0-9_]+'" | tr -d "'" | sort -u > /tmp/now.txt
git grep -h -A1 -E "(apply_filters|do_action)\(" v0.9.8 | grep -oE "'(openstation|desktop_mode)_[a-z0-9_]+'" | tr -d "'" | sed -E 's/^desktop_mode_/openstation_/' | sort -u > /tmp/old.txt
comm -13 /tmp/old.txt /tmp/now.txt | while read -r h; do grep -q "\`$h\`" docs/hooks-reference.md || echo "MISSING: $h"; done
```

   Make sure it prints nothing. On `trunk` it prints twenty lines.

2. Open `docs/hooks-reference.md` and find the three new entries. Make sure each one matches the code:
   * `openstation_brand_migration_map` against `openstation_migrate_brand_defaults()` in `includes/migrations.php`.
   * `openstation_track_type_registrants` against `openstation_should_track_type_registrants()` in `includes/core/payload.php`.
   * `openstation_my_wordpress_woo_coupon_bands` against `openstation_my_wordpress_woo_coupon_bands()` and `openstation_my_wordpress_woo_coupon_band_id()` in `includes/my-wordpress/integrations/woocommerce.php`.

3. Preview both files on GitHub. Make sure the tables render, and that the three cross-doc links in the new API index rows and the one in the `openstation_track_type_registrants` entry all land on the right heading.

4. Run `npm run build`. Make sure `git status` is clean afterwards.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-521-16dd3279b1d0de2a22f534ffe49c5d9cc7d32805.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/docs/api-index.md
+++ b/docs/api-index.md
@@ -193,6 +193,19 @@ shared-store + registries. Index:
 | `desktop-mode/agents-chat` shared-store key + `desktop-mode-agent-run` window | [`javascript-reference.md`](./javascript-reference.md#ai-agents--client-surface-experimental) | Experimental |
 | `agent` WP Explorer entity kind | `registerEntityKind()` seam | Experimental |
 
+### WooCommerce integration *(Experimental — inert unless WooCommerce is active)*
+
+No `wp.os.woo` namespace. The surface is PHP filters + REST + a
+native window, all hanging off WP Explorer. Index:
+
+| Surface | Where | Status |
+|---|---|---|
+| What the integration renders, and why | [`plugin-compat-layer.md`](./plugin-compat-layer.md#the-site-window-side-woocommerce) | Experimental |
+| `openstation_my_wordpress_woo_*` filters (orders, products, coupons, store, summaries, customers) | [`hooks-reference.md`](./hooks-reference.md#woocommerce-integration--experimental-filters) | Experimental |
+| `desktop-mode/v1/woocommerce/{orders, store, summary/<type>/<id>, customers, customers/<id>}` | `includes/my-wordpress/integrations/` | Experimental |
+| `openstation_woo_customer` REST field on the core `user` resource | [`hooks-reference.md`](./hooks-reference.md#customers) | Experimental |
+| `desktop-mode-woo-customer` native window *(retargetable singleton, `customerId` param)* | [`hooks-reference.md`](./hooks-reference.md#the-customer-window) | Experimental |
+
 ## CustomEvents on `document`
 
 Every event bubbles from `document`. See [`javascript-reference.md`](./javascript-reference.md#1-customevents) for `detail` shapes.
--- a/docs/hooks-reference.md
+++ b/docs/hooks-reference.md
@@ -1034,6 +1034,20 @@ Return `false` to suppress the dialog — useful for managed-host onboarding flo
 
 ---
 
+### `openstation_brand_migration_map` — Experimental
+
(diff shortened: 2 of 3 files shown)


mmtr:
Removed `openstation_brand_migration_map`, both the `apply_filters()` call in `includes/migrations.php` and the reference entry this branch had added.

It should not be a public surface. The rebrand migration runs once, against one release's stored defaults, and a third party rewriting which values get migrated would leave desks in a state no later migration accounts for. The plugin is also pre-1.0, so there is nothing to keep compatible with.

The map is now a plain array. The `empty( $map )` guard went with it, since it existed only to honour a filter returning nothing and is dead code against a literal.

Checked before removing: no test references the filter or `openstation_migrate_brand_defaults()`, and the brand assertions in `migrationActivation.php` are about migration 5's announcement, not migration 4's map. `php -l` and `phpcs -n` are clean on the file. CI runs the full PHPUnit suite against this branch.

That leaves two new entries in this PR rather than three: `openstation_track_type_registrants` and `openstation_my_wordpress_woo_coupon_bands`, plus the WooCommerce index table and the `api-index.md` block.

PR #524, opened 2026-08-07
About tab: show the credits while the particle scene loads

mmtr:
## Proposed Changes

OpenStation Preferences > About now paints its credits on the first frame instead of showing an empty dark panel for about three seconds while the PixiJS scene loads.

The tab renders a poster (the same eyebrow, title, byline and version the scene draws inside the canvas, as plain HTML) with an `<os-spinner>` in the band the logotype will fill. The canvas cross-fades in over it once the scene is live.

Two side effects worth calling out:

- The poster fades out but stays in the DOM. The canvas exposes no text to assistive technology, so the faded poster is the only accessible copy of what the tab says.
- A scene that never loads (no WebGL, blocked bundle) now leaves a readable About tab instead of a blank panel. The spinner goes and so does the dead canvas host.

The tabpanel background is now the scene's own radial backdrop rather than a different linear gradient, so the fade-in doesn't swap the background under the reader.

Because every word on that tab was gated on a graphics bundle. Three network round-trips plus a pixel-sampling pass sit between clicking the tab and the canvas having anything on it, and there was nothing on screen meanwhile, so a first click read as broken.

## Testing Instructions

1. Load `/wp-admin`, open `OpenStation Preferences` from the dock, then the `About` tab.
2. Make sure the panel is never blank. "WordPress OpenStation", "Crafted with curiosity", "an experiment by Automattic", the version line and a spinner must all be there from the first frame.
3. Wait for the particle scene. Make sure the spinner disappears, the canvas fades in rather than popping, and the text lands close to where the poster had it (a small shift is expected, a jump across the panel is not).
4. Move the cursor over the swarm and click it. Make sure the magnetic pull and the click shockwave still work, so the faded poster is not eating pointer events.
5. Hard-reload with the network throttled (DevTools > Network > Slow 3G) and reopen the tab. Make sure the poster holds the screen for the whole wait rather than flashing.
6. Simulate a failure: with DevTools open, block `**/about-scene*.js` (Network > right-click a request > Block request URL), then reopen `OpenStation Preferences` and the `About` tab. Make sure the credits and version stay readable, the spinner is gone, and the crosshair cursor is gone too. The console logs `[desktop-mode/about] scene mount failed`.
7. Resize the OS Settings window narrow and wide while the scene is still loading. Make sure the poster stays centred and the title scales instead of wrapping badly.
8. Turn on `prefers-reduced-motion` (macOS: System Settings > Accessibility > Display > Reduce motion) and reopen the tab. Make sure the canvas still arrives, just without the fade.
9. Switch to another tab and back a few times, and close and reopen the window. Make sure the scene still mounts every time and the poster does not reappear over a live canvas.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-524-ec4027ba76832756cd0494f2a372fc0cc8c5d6ec.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/assets/css/os-settings.css
+++ b/assets/css/os-settings.css
@@ -1235,16 +1235,16 @@
 }
 
 /*
- * The poster — the same copy the scene paints inside the canvas,
+ * The poster. The same copy the scene paints inside the canvas,
  * as plain HTML, on the first frame. Three network round-trips (the
  * PixiJS vendor bundle, the about-scene bundle, the logotype PNG)
  * plus a pixel-sampling pass sit between opening this tab and the
  * canvas having anything on it; this is what the reader looks at
  * meanwhile.
  *
- * Proportions echo `computeLayout()` in about-scene.ts — eyebrow near
+ * Proportions echo `computeLayout()` in about-scene.ts: eyebrow near
  * the top tenth, title under it, then the band the logotype fills,
- * with byline and version stacked below — so the cross-fade lands the
+ * with byline and version stacked below, so the cross-fade lands the
  * text close to where the canvas is about to draw it. Close, not
  * pixel-exact: the scene measures its own text and the fade covers
  * the difference.
@@ -1262,7 +1262,7 @@
 	align-items: center;
 	padding-inline: 24px;
 	text-align: center;
-	/* Never in the way of the swarm — the canvas below owns the
+	/* Never in the way of the swarm. The canvas below owns the
 	 * pointer once it is live. */
 	pointer-events: none;
 	font-family: -apple-system, blinkmacsystemfont, 'Segoe UI', roboto,
@@ -1292,7 +1292,7 @@
 /*
  * Faded, not hidden. `visibility: hidden` or `display: none` would
  * take it out of the accessibility tree, and the canvas exposes no
- * text at all — the poster is the only readable copy of what this
+ * text at all, so the poster is the only readable copy of what this
  * tab says.
  */
 .os-settings__about.is-scene-ready .os-settings__about-poster {
--- a/assets/css/os-settings.css
+++ b/assets/css/os-settings.css
@@ -1157,15 +1157,16 @@
 /*
  * About tab — full-tabpanel PixiJS canvas. Every credit string and
  * the AUTOMATTIC logotype (formed by particles) render inside Pixi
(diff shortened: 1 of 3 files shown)


mmtr:
Closing. The cross-fade reads worse than the blank panel it replaced: the poster's text does not hold the same position as the text the scene paints in WebGL, so the hand-over jumps rather than resolves.

The approach assumed the poster could sit close enough to `computeLayout()`'s marks that the swap would go unnoticed. Alignment was checked at one viewport (800x600) and landed the eyebrow and title on their marks but the byline and version within about 20px, which is plainly enough to see, and the drift will differ at other panel sizes since the two are laid out by different systems (CSS flow versus scene coordinates).

Not reopening this shape. Anything that stands in for the scene has to derive its positions from the same source the scene uses, or not attempt to occupy the same space at all.

The two underlying problems are still open and tracked in DESKMOD-83:

1. The tab paints nothing for roughly 3 seconds while PixiJS, the logotype image and the sampling pass complete.
2. More importantly, the About tab exposes no text to assistive technology at all. Every word on it is `Pixi.Text` inside the canvas, and `about-scene.ts` sets no `aria-*` or `role`. That is independent of the loading state and worth fixing on its own terms.

Issue #517, opened 2026-08-07
Agents: a run can complete "successfully" with an empty final answer — toText() failure is swallowed into text:null, bypassing both the retry and the error surface


juanlentino:
# Agents: a run can complete "successfully" with an empty final answer — `toText()` failure is swallowed into `text: null`, bypassing both the retry and the error surface

## Environment

- OpenStation **v0.9.8**, agents feature flag enabled
- Provider: Anthropic via the Core AI Client (Connectors), default model
- Agent: custom agent, administrator role, 5 allowlisted abilities (third-party plugin abilities + `get-llms-txt`)
- Observed live on a production single-owner site, 2026-08-06

## Symptom

An agent chat run ends with the UI rendering **"The agent finished without a text answer."** — no error, no retry, no diagnostics anywhere (nothing in the PHP error log either, and by the code path below, nothing *can* be logged). The tool-call trace shows successful read-only tool calls; the final synthesis turn simply never surfaces.

Observed **2 out of 3 runs** with the same agent:

| Run | Chat | Ask | Tool calls | Outcome |
|---|---|---|---|---|
| 1 | fresh | "review post X against the site's voice rules" | 5 (all reads) | ✅ full markdown answer + 2 `call_to_actions` buttons |
| 2 | follow-up via a `call_to_actions` button ("propose a revision splitting a long sentence") | 1 | ❌ "finished without a text answer" |
| 3 | **fresh** | explicit "propose a revision to post X splitting sentence Y, stage as revision" | 2 (both reads) | ❌ "finished without a text answer" |

Run 3 being a fresh conversation rules out history composition as the trigger. The pattern that remains: the successful run's answer was a **short review**; both failures asked for a **revision proposal**, whose structured answer (markdown + Gutenberg block markup inside a JSON string) is much longer — consistent with an output-token cap or a response-shape edge the SDK's text extraction can't handle. I have not been able to confirm the provider-side trigger from outside; the swallowed exception below is exactly the evidence that's missing.

## Root cause (code walk, v0.9.8)

**1. The adapter swallows text-extraction failure** — `includes/ai-copilot/client.php`, `desktop_mode_ai_client_generate()`:

```php
$text = null;
if ( empty( $function_calls ) ) {
    try {
        $text = $result->toText();
    } catch ( \Throwable $e ) {
        $text = null;   // exception discarded, no log, no error
    }
}

return array(
    'text'           => $text,   // null returned as a SUCCESSFUL generation
    ...
);
```

[...]

PR #530, opened 2026-08-07
AI: surface an empty final answer as an error instead of a silent success

juanlentino:
Scoped to the adapter change green-lit in #517 (plus the belt-and-braces runner guard).

## What

A final turn with no function calls whose text cannot be extracted — `toText()` throws, or the response carries no text part (observed live: a hard agent task spends the entire `max_tokens` budget inside a thinking block, `stop_reason: "max_tokens"`) — was swallowed into a "successful" generation with `text: null`. The agents runner then reported success with an empty answer, bypassing both the transient retry and every error surface.

`openstation_ai_client_generate()` now returns `WP_Error( 'openstation_ai_empty_answer' )` for that case, with the original extraction detail preserved in the error data.

## Consumers, walked

- **Search agentic loop**: the duplicate local `! is_string( $text )` check (the old `openstation_ai_empty` branch) is removed — the centralized error reaches the loop's existing `is_wp_error` handling at both generate sites. Net behavior change: an empty-string final turn now reports `openstation_ai_empty_answer` instead of falling through to `openstation_ai_result_parse`.
- **Command follow-up**: the deliberate degrade is preserved exactly as requested — this path matches the new code and still falls back to `"Done."` with the `fallback` flag, so a command that DID run never surfaces as a failure, and no `openstation_ai_search_error` fires for it (same as today).
- **Agents runner**: belt-and-braces guard in the toolless-turn branch (trim-empty text ⇒ error, never a success with `''`), plus a humanized mapping (`openstation_agent_empty_answer`) telling the user the model likely spent its output budget reasoning. The new code is deliberately NOT in the transient-retry signature list — the failure is task-shaped, not a flap; happy to add it if you'd rather retry once.
- `stop_reason` isn't reachable from the SDK result surface this adapter touches, so per your note I didn't block on it — the provider's exception message lands in `detail`.

## Tests

Three additions to `Tests_OpenStation_AgentsRunner` via the `openstation_agent_runner_generate` pre-filter seam: a `text: null` final turn is an error (not an empty success), whitespace-only text is the same failure, and the humanize mapping preserves the extraction detail. `php -l` clean on all four files.

Honest caveat: I couldn't run `npm run test:php` locally (no Docker on this machine) — the suite additions mirror the file's existing patterns and CI runs the same scripts.

Evidence for the mechanism (raw provider captures): https://github.com/WordPress/openstation/issues/517

Follow-up issues for the other three threads from the maintainer reply (provider config filter, async runner shape, replayed placeholder) are being filed separately as requested.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-530-6fc7222742f31300814ef7ad2c753fa89e7117a3.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


tests/phpunit/tests/agentsRunner.php
```diff
@@ -1021,4 +1021,72 @@ static function () use ( &$calls ) {
...
+	 * @covers ::openstation_ai_empty_answer_error
+	 */
+	public function test_humanize_maps_empty_answer_to_actionable_message() {
+		$raw = openstation_ai_empty_answer_error( 'The provider response contains no text part.' );
+		$this->assertWPError( $raw );
+		$this->assertSame( 'openstation_ai_empty_answer', $raw->get_error_code() );
+
+		$human = openstation_agent_humanize_generate_error( $raw );
+		$this->assertSame( 'openstation_agent_empty_answer', $human->get_error_code() );
+		$this->assertStringContainsString( 'output budget', $human->get_error_message() );
```
AllTerrainDeveloper: Hey! I'm wondering if this can be replaced with an error number check

AllTerrainDeveloper (approved):
There is a NIT that can be dismissed if there is no way of checking the error number :)

tests/phpunit/tests/agentsRunner.php
juanlentino: Good call — there is one: the humanizer stamps `status => 502` in the error data (`runner.php`, the `openstation_agent_empty_answer` branch). Swapped the copy pin for that plus a non-empty-message check in 557f2a9, so rewording or translating the sentence can never break the test.

PR #538, opened 2026-08-08
Tile contrast, comments-window styling, Overview clicks, and preview refreshes

AllTerrainDeveloper:
Five independent fixes, one commit each — reviewable in isolation.

### `fix(my-wordpress)` — tile text was unreadable on a dark station

`--os-tile-fg-muted` was declared as a bare literal next to the thing it paints. A declaration on a nearer ancestor beats an inherited one, so it pinned every tile's secondary line to near-black and put the palette's value — and every desktop theme's, Legacy included — out of reach in that subtree. It measured **1.11:1** against the brand's dark surface, while the label right beside it followed the theme correctly.

Adds `stylesheet-token-reachability.test.ts` as the guard — the light-DOM counterpart to `component-token-reachability.test.ts`, which already catches this one layer in on a component's `:host`.

### `fix(comments-window)` — thread rows rendered unstyled

The `os-` rename reached the stylesheet but not the JS namespace constant or the PHP root div, so every JS-rendered node carried a class no selector matched. Only the CSS class moves — the window id stays `desktop-mode-comments`, frozen per AGENTS.md.

### `fix(plugins-window)` — inline styles referenced removed tokens

`--wp-desktop-text-muted` / `--wp-desktop-accent` no longer resolve, so each fell through to its literal and ignored the palette.

### `fix(selection)` — switching desktops from Overview did nothing

Starting a marquee on the desktop area during Overview is meaningless (the layers it selects from are hidden) **and** destructive: `capturePointer()` retargets the rest of the gesture, so the synthesized `click` landed on the desktop area instead of the top-bar tile that was pressed. Thumbnails escaped it because Overview's capture-phase handler stops propagation first; top-bar and bare-backdrop presses share this node, where `stopPropagation()` cannot unregister a same-node sibling listener.

### `fix(editor-preview)` — companion refreshed for saves that never happened

Two independent paths:

1. **The 5 s backstop answered `saved` unconditionally.** Core's `save()` returns early when `compareString === lastCompareString` — no request, no `after-autosave`. A silent 5 s is core *declining* to autosave, not a save it forgot to announce. The shell refreshed ~5.4 s after the eye click, late enough to look like a reaction to whatever was clicked next. Now answers `not-dirty`.
2. **The live watch announced round-trips nobody asked for.** The bound events fire for non-edits (TinyMCE adds an undo level on blur, emits `SetContent` on programmatic writes), and core keeps autosaving a post it considers dirty because `getPostData()` calls `editor.save()` as a side effect. Fingerprinting *through* `getPostData()` inherits that side effect and can never catch it — `editor.getContent()` is the stable read, and is read-only by contract.

Also adds `tests/e2e/editor-preview/`, a Puppeteer regression for the bug class jsdom structurally cannot catch — the stubs encode an assumption about WordPress, and twice the tests passed while the bug shipped. **Unwired from CI on purpose:** no npm script, and Puppeteer is not a repo dependency (see its README).

## Verification

| Gate | Result |
|---|---|
| `npm run typecheck` | clean |
| `npm run lint` | clean |
| `npm run lint:php` (`phpcs -n`) | 207/207 clean |
| `npm run test:js` | 325 files, 4007 tests passing |
| `npm run build` | clean |

Frozen-value check: the `desktop-mode-comments` window id is intact at all 8 registration/lookup sites, and the stylesheet has 80 `.os-comments` selectors with zero orphans.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-538-e61bf0e94856bf8e1764b48218cf02f6fddd986f.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
@claude can you review?

PR #542, opened 2026-08-10
Set OpenStation windows free into real OS windows (Electron adapter)

AllTerrainDeveloper:
https://github.com/user-attachments/assets/e39241e0-0e13-4c54-84b4-0f96f1a5d6a2

Any OpenStation window can now leave the desk and become a **real window of your operating system** — its own dock entry, its own Alt-Tab slot, sitting among your native apps.

Pick **⋯ → Send to your Mac** (or *Windows PC* / *Linux desktop* — the label follows the host). Closing the native window brings it back. It works from the app *and* from a plain browser tab.

Send to your Mac in the window's ⋯ menu

## What this adds

**A desktop host, shipped as an extension.** Everything Electron-specific lives in `extensions/openstation-electron-adapter/` — a separate WordPress plugin plus the Electron app it talks to. Deactivate it and OpenStation is exactly the browser experience it was.

**Both window types travel.** An iframe window opens on the chromeless page it was already showing — same session, same scroll, same admin JS. A native window (Files, Games, a plugin canvas) has no URL of its own, so it opens in **solo mode**: the whole shell, painting exactly one window, no desk around it. Either way it is the same window, not a lookalike.

**The browser gets native windows too.** The app runs a loopback agent and hands its coordinates to the site, so a Chrome tab can ask your machine to open a window. Start the app after the page loaded and one ⋯ click picks it up — no refresh.

**New windows go to the desktop.** Launch a game from a freed Games hub and the game opens as its own OS window; the hub keeps showing the hub. Same for any `window.open()` in the app, except off-site URLs and "Open in browser tab", which go to the browser as asked.

**Windows keep their names.** The OS title bar reads *Trash*, *Posts*, *Inkfall* — not the browser tab title.

## Core gains two generic capabilities

Neither mentions Electron, and both stand on their own:

| | |
|---|---|
| **`wp.os.registerWindowAction()`** | Rows in every window's ⋯ menu — the one title-bar surface with no extension point. `label` / `icon` / `isVisible` may be functions of the window, re-read per open, so one row can express a toggle. Paired with `HOOKS.WINDOW_MENU_OPENED` and a menu that repaints while open, a plugin can answer with something async and still land its row under the pointer. |
| **`?openstation_solo=<window-id>`** | Boot the shell painting exactly one window — no dock, taskbar, wallpaper, desk or admin bar, and no session restore. A rendering mode, not an access grant. Useful for embeds and kiosks, not just this. |

Core footprint: **857 lines across 16 files.** Nothing in `window-manager/`.

## Try it

1. Symlink and activate the adapter — **this step is the feature**; without it the app connects and no ⋯ row appears:
   ```bash
   cd path/to/wp-content/plugins
   ln -sfn desktop-mode/extensions/openstation-electron-adapter openstation-electron-adapter
   wp plugin activate openstation-electron-adapter
   ```
2. Run the app:
   ```bash
   cd extensions/openstation-electron-adapter
   npm install && npm start
   ```
3. Enter your site address (`http://localhost:8889`), sign in, and choose **Open here** or **Use my browser** — both connect.
4. Open any window → **⋯ → Send to your Mac**. It appears on your desktop; the row becomes *Bring back into OpenStation*.
5. Close the native window → it docks back into the shell.

Worth trying specifically:

- **A native window** — Trash or Games — for the solo path.
- **From a browser tab** at `/openstation/`, with the app running.
- **Games → play a game** from a freed Games window: the game opens as its own OS window.
- **Start the app after** loading the browser tab, then click ⋯ — the row appears without a refresh.

## Security notes

The local agent is bound to `127.0.0.1`, requires a per-installation bearer token (which forces a CORS preflight, so a hostile page cannot fire a blind request), accepts exactly one origin — the paired site — and re-checks every URL against that site before opening a window. The token reaches the page only while a host is actually live.

## Tests

**155** extension tests (agent gates, connection pacing, both freed-window registries, URL rules, store) and **35** core tests (action registry, menu painting, solo mode, the adapter's REST surface). Full suites green: **4060** vitest, **2123** PHPUnit, phpcs, typecheck, lint, build.

## Status

**Experimental.** `wp.os.electron`, the REST routes and the solo flag may still move. The capability-probe model — a global whose absence means "browser, behave normally" — will not, and neither will the rule that core stays free of Electron.

Docs: [`docs/desktop-host.md`](docs/desktop-host.md) · [`extensions/openstation-electron-adapter/README.md`](extensions/openstation-electron-adapter/README.md)

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-542-186e968a93d87cd03d09161fb5f45e0b32b12c04.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
@claude can you review this?

AllTerrainDeveloper:
@claude can you review once more?

mmtr:
Reviewed the code and ran through the manual testing, both the solo path in a browser and the full app with the adapter activated.

Findings below, split into what came up while testing and what came up reading.

## Found while testing

### Solo mode leaks the desktop icon rail

`solo.css` hides `.os-icons`, which is the Classic grid. The Spatial layout paints into `.os-files-layer` (`src/desktop-files/layer.ts:238`) and that selector is not in the list. The docblock at `src/desktop-icons.ts:210` says it plainly: "the Spatial layout paints `<os-tile>` placements in the files layer, and on a stock install that second one is what the user is actually looking at".

It is masked whenever a window paints, because `body.os-solo .os-window` fills the viewport at inset 0 with no radius. I hit it with `?openstation_solo=nope`, which paints nothing. It is also a flash on every solo boot before the window lands.

Worth fixing rather than shrugging at, because the file's own reasoning for `display: none` over `visibility` is "so nothing keeps a layout box, an animation frame, or a pointer target". The files layer keeps all three, including drag and drop targets. One line:

```css
body.os-solo .os-files-layer,
```

Screenshot 2026-08-10 at 13 18 31


### The shell window shows the raw WordPress tab title

Freed windows get the treatment: `createFreedWindow` prevents `page-title-updated` and runs `screenNameFrom()`, so the OS title bar reads "Posts". `openShellWindow()` has no such handler, so the main window reads `Plugins ‹ WordPress Develop — WordPress`.

The PR says "Windows keep their names", and the main window is the one you look at most.

Screenshot 2026-08-10 at 13 27 24

### macOS Dock tooltip says "Electron"

`scripts/brand-dev-bundle.mjs` rewrites `CFBundleName` and `CFBundleDisplayName` and swaps the icon. The icon lands, the name does not: LaunchServices caches the bundle's display name, so the plist edit alone does not move the Dock tooltip. Either refresh `lsregister` from the script or soften the claim in its docblock, since right now it says it fixes the Dock name and in practice it does not.

Screenshot 2026-08-10 at 13 27 09


### Pink outline on the dock and the side rail, app only

Happens in the app, not on the same site in Chrome, so it is not site state. I could not track it down. `--os-dock-border` defaults to near white at 10%, the only shipped desktop theme sets it to white at 8%, and neither this PR's `solo.css` nor the adapter's `solo-host.css` applies to the shell page. The small pink indicator dots are separate and correct, those are `--os-dock-item-outline`.

Any idea what would differ under Electron here?

Screenshot 2026-08-10 at 13 26 31


## From reading the code

### The ⋯ menu subscription outlives its window

`src/window/menus.ts:87` subscribes to the registry so the menu repaints while open, and only `closeActionsMenu()` unsubscribes. `destroy()` at `src/window/index.ts:3384` tears down `_boundOnDocumentPointerDown` but not this one, so a programmatic `win.close()` with the menu open leaves a live listener holding the window and a detached panel. Click driven closes hide it, because the pointerdown capture handler closes the menu first. One line next to the existing cleanup.

### Off-site auth redirects

After settling, `navigationVerdict` sends any off-site navigation to the browser, so a site using Jetpack or WordPress.com SSO, or an off-site 2FA step, bounces the user out mid login and strands the app on the login screen. Worth a note in `docs/desktop-host.md` at least.

The related edge: if that first chain does land off-site, `store.set( 'siteUrl', landed )` makes that origin the agent's `allowedOrigin` and the navigation guard's allowlist. Narrowing the settle to the same registrable domain would close it without breaking the `example.com` to `www.example.com` case the comment is protecting.

### CI does not run the adapter's JS suite

`ci.yml` has per-extension jobs for popup-siege and feed-buddy, including `git diff --exit-code -- extensions/desktop-mode-feed-buddy/assets/js` to catch a stale committed bundle. The adapter adds none, so its 155 tests, its typecheck, its lint, and the check that `assets/js/electron-adapter.js` still matches `src/` all run only locally. The PHP side is fine, `tests/phpunit/tests/electronAdapterHost.php` loads the extension directly and rides the core suite.

### No `.distignore`

`bin/package-extensions.sh` zips `git ls-files -co` minus `.distignore`, and only popup-siege ships one. Without it the adapter's plugin zip carries `app/`, `tests/`, `scripts/`, the vite, vitest and eslint configs, `build/icon.icns` and a 12,961 line lockfile. WordPress only needs the plugin file, `includes/`, `assets/` and the README.

### `wp.os.electron` skips `registerNamespace()`

`src/index.ts` does `os.electron = api` directly, which bypasses the `RESERVED_NAMESPACE_KEYS` collision check in `src/api/facade.ts:757`. The file docblock says the adapter "uses nothing but OpenStation's public API", and `registerNamespace` is that API. It also means a second plugin claiming `electron` silently wins.

## Smaller

- Raw `fetch` in `src/agent-bridge.ts`. The loopback calls are defensible since they are not site requests, but `fetchPairing()` hits the site REST root and should go through `wp.os.fetch` with `{ silent: true, source: … }`, or carry the documented `eslint-disable` with a reason. feed-buddy's eslintrc has the rule, the adapter's does not.
- Generic solo mode has a silent failure that the forwarder only fixes under a host. An embed or kiosk, which the docs invite, opens a second window and gets an invisible unreachable one with no error. Worth a caveat in `docs/desktop-host.md`, or a console warning from core when a second window opens in solo mode.
- The agent token never rotates. `Store.agentToken()` generates once and persists, and it is mirrored into user meta and printed into every admin page. The capability is bounded to opening a same site window, so impact is low, but the port already changes per launch and rotating the token with it is close to free.
- `LocalAgent.readBody` counts `raw.length` in UTF-16 code units rather than bytes, so `MAX_BODY` is up to three times looser than documented for multi-byte input. It also writes a response after `req.destroy()`.
- `routeNewWindow` matches `desktop_mode_classic=1` as a substring of the whole URL, so a path or fragment containing it routes to the browser. `searchParams` would be exact.
- `certificate-error` offers "Continue anyway" for any localhost URL, not only the configured site, which is narrower than the reasoning in its own docblock.
- `src/window-actions/server-sync.ts` never clears `loadedUrls` on deactivation, so deactivate then reactivate leaves the script cached and the owner swept actions unregistered until a reload. This is inherited verbatim from `title-bar-buttons/server-sync.ts`, so a family wide follow up rather than something to hold this PR for.

AllTerrainDeveloper:
@claude what about a last review?

PR #551, opened 2026-08-11
Make the active window tab a frosted plate that slides

nuriapenya:
The sub-page strip was flat text with an accent underline, on a strip painted the same colour as the focused title bar. With no container and no relationship to the page it navigates, it read as loose text floating in the chrome.

## What changed

- **The active tab is a tab.** It wears the page's own fill (a chromeless admin page is always `#fff`, whatever the admin colour scheme says), filleted into the floor of the track at both bottom corners, so tab and page are one continuous surface.
- **A Pulsemesh rail traces that silhouette.** One continuous line runs the top edge of the page, curves up through the fillet, traces the tab and comes back down — outlining page-plus-tab as a single shape. Every piece samples one mesh laid across the whole strip, so the gradient carries through the corner where rail becomes fillet becomes tab.
- **The surface slides.** It is one element that travels between tabs rather than a fill that switches off on one and on at the next. A fill that switches has to crossfade a dark tab into a light one, and every frame in between is a grey that belongs to neither.
- **Window chrome comes down one step.** A focused title bar and its tab strip are now the same Obsidian, so the only thing lifting out of that surface is the tab. Unfocused sinks to Void. The focus step is the same size; the whole ramp moved.
- **The always-on activity dot is gone** from the title bar. It was an accent ring painted on every window for its whole life to report a state that is true almost all of the time. The phase machinery is untouched and still reference-counts; nothing renders it unless a plugin mounts an `<os-save-status data-os-activity-indicator>` in a title-bar slot.

## How to test

[Open in Playground](https://playground.wordpress.net/?blueprint-url=https://openstation-tabs.pages.dev/blueprint.json), then open **Tools** or **Appearance** from the dock and click through the sub-page tabs.

Worth checking: the rail closes up when no tab is active; a window with no sub-pages (Comments) shows no strip at all; an external sub-tab's detach and close chips stay legible on the light plate; and the plate follows tabs added or removed at runtime.

## Notes

- New theme tokens: `--os-tabs-rail` (set to `none` for no line), `--os-tabs-rail-width`, `--os-tabs-active-frost`, `--os-tabs-active-crown`.
- The plate's face is deliberately opaque with no `backdrop-filter`. A translucent element with a backdrop filter is promoted to its own compositor layer, and the edge of that layer leaves a faint hairline down the plate's sides — which against a bright rail reads as a second, dimmer line just inside the real one.
- The Legacy desktop theme is a frozen snapshot and does not carry the new tokens, so a Legacy window gets the new geometry with the brand's rail. Its own `--os-tabs-bg` / `--os-tabs-color` still apply.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-551-8dfccd0e346ab8858d0b3e12a8a5ed7acce4daa9.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper (approved):
Lovely!

PR #564, opened 2026-08-12
Build a server-registered window's tabs in the window chrome

nuriapenya:
Stacked on #551 — review that one first. Base is `window-tabs-as-tabs`; retarget to `trunk` once #551 lands.

#551 taught the shell's tab strip to serve native windows. This is the other half: the tabs a plugin registers with `openstation_register_window_tab()` move into that strip too, so a window has **one** tab strip in one place, whether it is an admin page in an iframe or a native window.

## What changed

A native window with more than one registered tab used to have its body wrapped in `<os-stack>` + `<os-tabs>` + one `<os-tabpanel>` per tab. The strip was part of the window's content. It is now built by the shell in the window chrome, under the title bar, from the tab metadata the payload already carries — so the template no longer contains one.

```html
<!-- before -->
<div class="os-window__body">
  <os-stack>
    <os-tabs value="main">…</os-tabs>
    <os-tabpanel for="main">…</os-tabpanel>
  </os-stack>
</div>

<!-- after -->
<nav class="os-window__tabs" role="tablist" aria-label="Jorvy sections">…</nav>
<div class="os-window__body">
  <os-stack>
    <os-tabpanel for="main">…</os-tabpanel>
  </os-stack>
</div>
```

`openstation_register_window_tab()` itself is untouched: same signature, same arguments, same `main_tab_padding` and `openstation_native_window_tab_wrap_padding` controls. Panes are still server-rendered and still stamped `hidden` so first paint is correct.

A plugin that only registers tabs and renders their panes needs **no change**, and picks up the roving `tabindex`, the arrow keys, `aria-controls` paired both ways with each pane, and the shell's active-tab surface for free.

## Breaking change

The emitted markup changes, so this breaks anyone who listened for `os-tab-change` on that strip or styled/queried it. `docs/migration-window-tabs.md` covers both, with the replacement event (`os-window-tab-change`, which bubbles from the window element) and the tokens to retint the strip through now that it is shared chrome.

## Notes for review

Two decisions worth a look:

**Labels no longer reach the template HTML.** They travel as payload data and are written with `textContent`. That is a stronger guarantee than the escaping it replaces — there is no HTML parse of the string on either side — so the PHP test that pinned the escaping now pins the absence instead.

**Panes are found by ownership, not depth.** A server-registered window wraps its panes in `<os-stack>`, so "direct child of the body" was never going to hold. A pane belongs to the window's strip unless it sits inside another tab group, which is what a pane holding its own `<os-tabs>` switcher looks like. Without that filter, opening such a window would hide half the inner group on the way past. Covered by a test.

## How to test

1. Check out the branch, `npm install && npm run build`, and open the shell.
2. Open OpenStation Preferences: its tabs are in the chrome, the rail traces the active one, and the pane below switches.
3. Register a probe window to exercise the server path:

```php
add_action( 'init', function () {
    openstation_register_window( 'tabprobe', array(
        'title'    => 'Tab Probe',
        'template' => static fn() => print( '<p>MAIN</p>' ),
    ) );
    openstation_register_window_tab( 'tabprobe', array(
        'value'    => 'acme/stats',
        'label'    => 'Stats',
        'template' => static fn() => print( '<p>STATS</p>' ),
    ) );
} );
```

Open it from the dock. Both tabs appear in the chrome, the namespaced value works, and clicking swaps the pane.

4. Keyboard: Tab into the strip once, arrow between tabs (focus moves, nothing activates), Enter or Space to choose, Home and End for the ends. Tab again leaves the strip for the pane.
5. `npm run test:js`, `npm run test:php`, `npm run lint`, `npm run lint:php`.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-564-0fdca846375a8473014158eee51b0483c3b6c000.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
@claude is there any kind of regression with the routing and how the menu/submenus are built?

AllTerrainDeveloper (approved):
I like it very much!

PR #568, opened 2026-08-12
Confirm dialog: route Enter to the focused button, trap and restore focus

AllTerrainDeveloper:
## Description

Fix `<os-confirm-dialog>` so keyboard interaction matches the focused control and modal focus is properly contained and restored.

- **Enter no longer confirms from Cancel.** The dialog swallowed every Enter and called `_confirm()`, so pressing Enter with Cancel focused ran the destructive branch. Enter is now the dialog's default action only while no control inside it owns the key — a focused button activates itself.
- **Focus is trapped.** Tab and Shift+Tab cycle between the dialog's own controls and cannot reach the page behind the scrim.
- **Focus is restored** to the control that opened the dialog. Captured through a shadow-piercing lookup, since `document.activeElement` stops at the outer `<os-*>` host. Restored both on `open` being removed and on unmount, because `osConfirm()` removes the element while it is still `[open]` and never sees the attribute change.
- **Escape still cancels** with a button focused.
- **Opening moves focus into the dialog** — onto Cancel when `danger` is set, so Enter is never the shortcut that deletes. The old synchronous `.focus()` in both `osConfirm()` copies ran before the first render and always found an empty shadow root; it is gone and the component owns focus.
- The focused button wears the kit's ring on `:focus` rather than `:focus-visible`. Focus arrives programmatically, which the heuristic suppresses when the dialog was opened with the mouse — leaving the pre-selected button silently armed for Enter.

## Testing

Open Recycle Bin → Empty Trash (a `danger: true` prompt):

1. Cancel opens focused and ringed; Enter cancels.
2. Tab moves the ring to *Empty Trash*; Enter there empties.
3. Escape cancels from either button.
4. Tab repeatedly — focus never leaves the dialog.
5. Close the dialog, then press Enter — focus is back on the button that opened it.
6. A non-danger prompt still opens on Confirm, where Enter confirms as before.

`npm run lint`, `npm run typecheck`, `npm run test:js` (4116 tests), `npm run build` all green. Six new tests in `os-confirm-dialog.test.ts` cover initial focus, the danger default, Enter on Cancel, Escape from a button, the Tab wrap, and focus restore.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-568-eb7558c52b6155bf93e179a8f83148095b88a80a.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
Closees 2 points of 559:
```
1. Pressing Enter on Cancel confirms a destructive dialog
Why this matters: A keyboard user can focus Cancel or Keep items, press Enter, and accidentally approve a deletion. Space works correctly; Enter does the opposite of what the focused button says.

To reproduce:

Open any OpenStation confirmation dialog.
Tab to Cancel or Keep items.
Press Enter.
The dialog resolves true and runs the confirmed action.
The dialog’s host-level key handler treats every bubbled Enter as confirmation, including Enter from the Cancel button.

Where to look: src/ui/components/wpd-confirm-dialog/wpd-confirm-dialog.ts, src/wpd-confirm.ts

2. Confirmation dialogs let focus escape and do not restore it
Why this matters: The dialog says it is modal, but keyboard focus can move into the page behind it. Once that happens, Escape may no longer close it. Closing the dialog also drops focus on the document body instead of returning the user to the control they started from.

To reproduce:

Open a confirmation dialog from a keyboard-focusable control.
Tab through both dialog buttons.
Press Tab once more.
Focus leaves the dialog and enters the background page.
Dismiss the dialog and notice that focus is not returned to the opener.
Where to look: src/ui/components/wpd-confirm-dialog/wpd-confirm-dialog.ts, src/wpd-confirm.ts
```

AllTerrainDeveloper:
@claude give a review

PR #571, opened 2026-08-12
A11y: keep actionable toasts alive while attended, trap focus in first-run dialogs

AllTerrainDeveloper:
## What

Two places keyboard focus could be taken away from the user, plus one window whose first-run dialog had gone missing.

**Toasts** (`src/toast.ts`, `src/ui/components/os-toast/os-toast.ts`) — `duration` is now a countdown, not a deadline. `<os-toast>` reports pointer-over / focus-inside as a reflected `held` attribute and an `os-toast-hold` event; `showToast()` pauses on it and resumes with the time left, floored at 1.2s so a nearly-expired toast doesn't vanish the instant the pointer leaves. Dismissing a toast that holds focus (clicking `Undo` removes the element the button lives in) hands focus back to the last element outside the toast stack, instead of dropping it on `<body>`.

**First-run dialogs** — the intro dialogs are hand-built light DOM and had no focus trap (`<os-modal>` / `<os-confirm-dialog>` carry their own, shadow-scoped). New `src/ui/modal-focus.ts` is one shared scope: focus opens inside, Tab wraps at both ends, focus that lands behind the scrim is pulled back, the scope self-unbinds if the dialog leaves the document, and release prefers the caller's context over the captured opener. Posts, Pages, Users and Plugins each pass their window root — these dialogs open themselves as the window paints, so there is no launcher worth returning to.

**Comments** — its intro was deleted in the conversation redesign (#441) while `introSlug` / `introSeen` / `introUrl` stayed wired server-side. Rebuilt on the shared scope, with copy describing the two-pane conversation view rather than the retired moderation table.

## Testing

`npm run typecheck`, `npm run lint`, `npm run test:js` (4181 passing), `npm run build` — all green. New: `tests/vitest/modal-focus.test.ts`, `tests/vitest/intro-dialog-focus.test.ts` (one 6-case table run across all five dialogs), plus hold / focus cases in `tests/vitest/toast.test.ts` and `os-toast.test.ts`.

Manual, on a dev site:

1. **Toast stays while attended** — trigger any toast with an action (e.g. open a sub-tab whose iframe is blocked by `X-Frame-Options`). Hover it past its 4s and it stays; move away and it goes ~1.2s later. Tab to the action button and wait — still there. Press Enter: the toast closes and focus lands back where you were, not on `<body>` (check `document.activeElement`).
2. **Dialogs trap focus** — OpenStation Preferences → Features → *Reset what's-new dialogs*, then open Posts, Pages, Users, Comments and Plugins. Each opens focused on "Got it"; Tab and Shift+Tab cycle only between the two buttons; clicking the desk behind snaps focus back into the dialog; Escape or "Got it" leaves focus inside the window that was being introduced.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-571-0c86331c71194ee2ec5a11bef100429541c0ee41.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
@claude give it a review

AllTerrainDeveloper:
This fixes 6 and 7 from 559

AllTerrainDeveloper:
All three addressed in 0c86331c.

**Concurrent scopes (the real one).** Confirmed, and worse than "focus doesn't settle" — each pull-back is a genuine focus change, so the two `focusin` handlers re-enter each other synchronously until the stack overflows. Went with the scope-stack direction rather than a shared "an intro is already showing" gate, since it fixes the general case for any future light-DOM modal, not just the five intros.

One wrinkle worth flagging: a **module-level stack would not have worked**. `posts-window`, `plugins-window` and `comments-window` are separate Vite IIFEs, each with its own compiled copy of `modal-focus.ts`, so per-bundle stacks would coordinate Posts↔Users and nothing else — exactly the Posts-then-Plugins pairing that triggers this. The stack lives in a `createSharedStore( 'openstation/modal-focus' )` slot per the cross-bundle-state rule in AGENTS.md.

Only the topmost scope traps Tab or chases focus; the ones underneath stay mounted and inert, and reclaim focus when the top releases (release drops its token *before* moving focus, so the scope below is live in time to take it). A scope whose root leaves the document drops its token on the next event, so it can't block the one underneath.

`isTopmost()` fell out of that, and it turned out to answer a second bug the stack alone doesn't: all five dialogs bind Escape on `document`, so with two open, one keypress closed both — dismissing *and marking seen* a dialog the user had not seen yet. Each dialog's `onKey` now gates on it.

**Escape capture inconsistency.** Fixed — Posts now binds (and unbinds) with `true`, matching the other four.

**`release()` treating `null` as "still ours".** The behaviour is deliberate, the comment was wrong. `deepActiveElement()` returning `null` means focus is on `<body>` — which is the failure the scope exists to prevent, and precisely where a backdrop click lands. Only a real, live element outside the dialog means the user moved on deliberately. Rewrote the comment to say that instead of tightening the condition; happy to flip it if you read it the other way.

**Tests.** `modal-focus.test.ts` gains a `two scopes open` block (second takes focus without a fight, only the topmost traps Tab, releasing hands control back down, a removed root stops blocking), and `intro-dialog-focus.test.ts` gains a Posts+Users case asserting one Escape closes one dialog and focus comes forward to the one behind. 4186 passing; typecheck, lint and build green.

Manual repro for the original bug: reset the what's-new dialogs, then open Posts and Users without dismissing the first — before this commit that recursed; now Users is on top, Escape closes Users, Escape again closes Posts.

PR #576, opened 2026-08-12
Window activity: the status ring

AllTerrainDeveloper:
https://github.com/user-attachments/assets/d7188268-a93c-4829-93c1-398e3581fc23

Stacked on #573 — base is that branch, so this diff is the indicator only. Retargets to `trunk` automatically when #573 merges.

Design by @nuriapenya. (Branch name predates the design and is stale — the glow it names was replaced.)

## The ring

The title bar has had no activity indicator since #551 removed the always-on modem dot, and the objection there was right: the dot had to occupy space whether or not anything was happening, so it spent a permanent accent ring reporting `idle`.

The phase is a **status ring** now, in the position the app icon used to hold — and **the app icon is gone**, because it was a copy of the window's own dock tile a few hundred pixels below it. A title bar has room for one mark of that size; this one changes.

| Phase | Ring |
|---|---|
| `idle` | quiet outline in the title bar's own muted glyph colour, lifting on focus like the control glyphs |
| `pending` / `saving` | accent outline, breathing |
| `saved` | accent fill, white check |
| `failed` | open red outline, red bang — persists until the next request starts |

**Only success fills.** The two outcomes differ in fill *and* glyph, not only in hue — colour alone is not a distinction every user can make.

It's an `<os-save-status variant="ring">`: a new documented treatment in the component kit rather than a one-off, so panel headers and the pinned-note indicator can wear it too. It is found through the public `[data-os-activity-indicator]` attribute, and **every** matching element in the title bar is driven rather than the first — the framework's ring claims no private channel, and a window showing two different phases at once would be worse than one showing none.

The `icon` slot stays, and stays empty: a plugin's `appearance.slots.icon` and a desktop theme's `APP:<id>` slot still render there. What's removed is the duplicate-by-default.

## Iframe windows report now

This was the real gap. Native windows route through `wp.os.fetch`, which the shell owns; an admin page inside an iframe does its own jQuery / XHR / `fetch` calls the parent had no way to see.

The chromeless bridge already wrapped both for the devtools panel — but reported **completion only**, which is useless for an indicator whose whole job is the part the user waits through. New `os-iframe-activity` message brackets each request (`start` / `end`) and feeds the same reference-counted `_markActivityStart` / `_markActivitySettled` pair, so a page firing six requests at once settles as one burst, and settles `failed` if any of them did. **Every method counts, `GET` included.**

Two exclusions, both deliberate:

- **Heartbeat never reports.** A poll nobody initiated, every 15 seconds, forever — it would light every open window's ring on a timer and flash a success check for a save that never happened. Same judgement `wp.os.fetch`'s `silent: true` exists for. The action name is read from the request body, since Heartbeat POSTs to `admin-ajax.php` with nothing in the URL.
- **A new document resets the count.** Navigating away takes the pending `end` messages with it, and a counter that only falls when someone reports back would strand the ring lit for the window's whole life. `os-ready` calls `_resetActivity()`.

## Also in here

A fallback-chain collapse in the component the ring is built from: `saved` / `failed` resolved their fill through `var( --os-ui-surface, … )`, and the palette declares `--os-ui-surface` as Obsidian — so the "green" dot has been painting near-black on near-black. They read the semantic colours now. **Visible change** in the pinned-note indicator, the one place that treatment still ships.

The screen-reader live region survives from the previous pass: `Saved` politely, `Not saved. <error>` assertively, in-flight deliberately silent.

## Test

1. **Native path** — OpenStation Preferences → change the wallpaper: the ring breathes, then fills with the check.
2. **Iframe path** — open Posts or Tools and search / filter / sort: the ring moves on the iframe's own AJAX.
3. **Failure** — DevTools → Network → **Offline**, repeat either: open red ring with a bang, and it stays.
4. **Quiet when idle** — leave a window alone for a minute; the ring must not twitch. That's the Heartbeat exclusion.

## Notes

- `build` / `lint` / `typecheck` / `lint:php` clean. `test:js` green (338 files, 4237 tests); `test:php` green (2155) — the chromeless bridge is PHP-emitted, so the PHP suite ran too.
- Tokens: `--os-titlebar-activity-color` / `-saved-color` / `-failed-color` / `-size` / `-idle-color` (+ `-idle-color-focused`). The resting ring follows `--os-titlebar-btn-color` unless overridden, so a retinted title bar takes the ring with it.
- Docs: `javascript-reference.md` (the ring + the new bridge message), `examples/window-activity.md`, `examples/README.md`, `desktop-themes.md`, `components-reference.md`.
- **Open question for review:** the mock uses Pulse for both in-flight and success. An earlier round of feedback asked for an amber/green/red traffic light instead. This implements the mock, with fill and glyph carrying the distinction rather than hue; `--os-titlebar-activity-saved-color` is one line if green wins.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-576-7a5c12ec759e149e4a9d2dfef72ddd1d3c405e9e.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
@nuriapenya — this touches the title bar you reshaped in #551, so flagging it rather than surprising you with it.

You removed the always-on modem dot there, and the reasoning holds: an indicator that occupies space whether or not anything is happening ends up spending a permanent accent ring on `idle`. This doesn't put it back. The phase is a glow behind the window icon, and idle removes the attribute entirely — an idle window matches no selector, so there's nothing to paint and nothing to animate. The icon already says which app this is; lighting it is how a desk lamp reports work.

Two bits that land in your area specifically:

- `.os-window__slot--icon` leaves the `display: contents` group and becomes `inline-flex; flex-shrink: 0`, because the glow needs a box to be positioned against. That reproduces exactly what the icon contributed to the flex row as the flex item itself, so the row should be unchanged — but **icon-to-title spacing is the thing to eyeball**, especially against the frosted tab plate.
- The glow hangs off the slot host rather than the icon so it survives `paintWindowSlots()` replacing the slot's children, and so it works for `` icons.

Colour is a traffic light: amber in flight, green through, red stopped. In-flight is deliberately not Pulse — a window glowing brand-magenta while it saves is decorating rather than reporting, and as part of a three-colour sequence the phase reads before it's parsed. All five values are themeable tokens, and `--os-titlebar-activity-strength: 0` turns the effect off wholesale if a theme wants the title bar quiet.

Quickest look: open OpenStation Preferences and change the wallpaper (amber → green), then DevTools → Network → Offline and change it again (amber → red, and the red stays until the next request).

AllTerrainDeveloper:
@nuriapenya — rebuilt to your mock, so the comment above is out of date. The glow is gone; the ring is in, and so is the icon removal you wanted. What landed:

- **Ring in the icon's place, app icon dropped.** The `icon` slot stays and stays empty, so a theme's `APP:<id>` slot or a plugin's `appearance.slots.icon` still renders if someone deliberately asks for one. `.os-window__slot--icon` is back to `display: contents`, so the layout worry from my earlier comment is moot — nothing about the row changed.
- **Only success fills.** Idle and saving keep the outline open, failure keeps it open too and tints the glyph instead. That's on purpose beyond matching the drawing: filled-vs-open plus check-vs-bang means the two outcomes still separate for someone who can't separate magenta from red.
- **Idle is chrome, not signal.** The resting ring takes `--os-titlebar-btn-color` — the same colour as the minimise/close glyphs, lifting on focus the same way — so a themed title bar takes the ring with it and an idle window doesn't wear an accent mark. Say the word if you'd rather it sat at a fixed value.
- Built as `<os-save-status variant="ring">` rather than a one-off, so the treatment is reusable and documented in the kit.

nuriapenya:
Looks great! 💎

Issue #552, opened 2026-08-11
A quick second change in Extended Options saves the wrong value


nickhamze:
If an administrator changes an Extended Option twice quickly, OpenStation can show the final choice while WordPress saves the first one. Everything looks fine until the page reloads and the setting jumps back.

## How to reproduce

1. Open **Preferences → Features** as an administrator.
2. Change **Enable drag-and-drop in the Media Library**.
3. Change it back before the first save finishes.
4. The checkbox ends where you left it.
5. Reload OpenStation. The first value returns because the second change was never saved.

I reproduced this by briefly delaying the first Extended Options request. The UI ended enabled, while the only request sent to WordPress contained `media_library_enhanced: false`.

## What should happen

The last choice should win. If OpenStation cannot accept another change while saving, the control should be temporarily disabled instead of pretending the second click was saved.

## What is happening

The control changes immediately, but the save function exits when another request is already running. Nothing queues the newer value for a follow-up save.

## Where to look

- `src/settings/sections/extended.ts`
- `includes/extended-options.php`

A good fix would either queue one trailing save with the latest state or lock the control until the current save settles. This still reproduces on current `origin/trunk` at `c670170`.

nickhamze:
## Screenshots

The checkbox shows the administrator's final choice, but the server still has the earlier one:

![Extended Option UI and saved value disagree](https://raw.githubusercontent.com/nickhamze/desktop-mode/issue-evidence-part-2/issue-assets/audit-552-558/552-ui-off-server-on.png)

After reload, the server's earlier value returns to the UI:

![Earlier Extended Option value returns after reload](https://raw.githubusercontent.com/nickhamze/desktop-mode/issue-evidence-part-2/issue-assets/audit-552-558/552-after-reload.png)

nickhamze:
The full 137-setting clarity pass reproduced this again with **Games**: the switch ended On, but WordPress had saved Off and the app disappeared after reload. The existing report and screenshot confirm the same lost-final-change path applies beyond Media Library. PR #581 improves the site-wide wording, but does not claim to fix this save race.\n\n![Games appears on while the saved value is off](https://raw.githubusercontent.com/nickhamze/desktop-mode/issue-evidence-part-2/issue-assets/settings-clarity/clarity-contract-extended-games-rapid-ui-on-panel.png)

PR #581, opened 2026-08-12
Make OpenStation settings easier to understand

### LINKED ISSUE #584: Appearance says preferences stay in one browser, but they follow the account
The Appearance introduction says its changes are “saved to this browser.” They are actually saved to the current WordPress account and follow that person into another browser.

That distinction matters. Someone using a shared account may reasonably change the appearance believing it is local to their device, then unintentionally change OpenStation everywhere that account is used.

## How to reproduce

1. Change the accent as one WordPress user.
2. Wait for the save to finish.
3. Sign in as the same user in a fresh browser session.
4. The new accent follows the account.
5. Sign in as a different user; that account keeps its own preference.

## Suggested wording

> Changes apply immediately and are saved to your WordPress account.

PR #581 makes this wording change as part of the full settings-language pass.

Appearance says browser even though the setting follows the account

### LINKED ISSUE #587: The theme recommendation button promises the wrong changes
The Themes action says **Apply recommended layout and effects**, but themes are free to recommend a different set of preferences.

During the audit, the Legacy theme changed only the accent. OpenStation changed the accent, layout, and dock position. Neither action changed the window or reveal effects promised by the label.

That makes the button hard to trust: it can change things it does not name and leave named things untouched.

## How to reproduce

1. Select the Legacy theme.
2. Set a distinctive accent, layout, and effect combination.
3. Use the recommendation action and compare the values.
4. Repeat with the OpenStation theme.

## Suggested wording

Use a truthful generic label such as **Apply recommended preferences**, or build a confirmation from the actual keys in the theme manifest.

PR #581 changes the button to **Apply recommended preferences** and updates the surrounding explanation.

The recommendation action is labeled as layout and effects

### LINKED ISSUE #589: Native-window choices do not explain that they apply on the next open
The switches for native Posts, Pages, Users, Plugins, and Comments do not replace a window that is already open. They choose which version OpenStation will use the next time the app opens.

That behavior is reasonable, but the current wording and the Features introduction make the change sound immediate. Someone can switch native Posts off, keep looking at the same native window, and reasonably conclude that the setting did nothing.

## How to reproduce

1. Open the native Posts window.
2. Turn **Use the native Posts window** off.
3. The open window remains native.
4. Close Posts and open it again.
5. The classic version is used.

## Suggested wording

> Applies the next time you open Posts. Existing Posts windows keep their current version.

PR #581 adds this timing note to all five native-window choices and changes the labels from “Use” to “Open … in.”

### LINKED ISSUE #590: The Features introduction promises behavior its controls do not share
The Features introduction makes one blanket promise: every change is personal and takes effect immediately without a reload. The controls below it do not all follow that contract.

- Comment AI is site-wide.
- Heartbeat changes require a reload before the running page uses the new rate.
- Native-window choices apply the next time that app opens.

When the introduction is wrong for several important controls, people cannot tell which individual descriptions are safe to trust.

## Suggested wording

> Most changes are personal and immediate. Site-wide or reload-required options are identified below.

Each exception should then say its own scope and timing. PR #581 makes that change and adds the missing notes to the affected controls.

Heartbeat is saved as 15 seconds while the running page still uses 60

### LINKED ISSUE #591: File Association dropdowns are unnamed for screen readers
The File Associations rows look labeled, but the actual dropdowns are unnamed in the accessibility tree. A screen-reader or voice-control user encounters eleven indistinguishable comboboxes and cannot tell whether a control changes Posts, Media, Users, Web links, or another file type.

## How to reproduce

1. Open **Preferences → File Associations**.
2. Inspect the controls with a screen reader or browser accessibility tree.
3. Move through the dropdowns.

## What should happen

Each dropdown should expose a distinct name such as **Default app for Posts**.

## What is happening

The tab sets `aria-label` on the `<os-select>` host. The component builds the real focusable `<select>` inside its shadow root from its `label` property, removes a host `aria-label` when that property is empty, and never forwards this caller’s value to the native control.

## Where I would look

- `src/desktop-files/settings-tab.ts`
- `src/ui/components/os-select/os-select.ts`

PR #581 now includes the small follow-up that sends the label through the component's supported `label` property, so the real native select receives the name. This issue can close with that PR.

File Association rows have visible context but unnamed controls

### LINKED ISSUE #596: Posts and Pages silently share the Author and Date column choices
Author and Date visibility are shared by native Posts and native Pages, but neither menu says so. Hiding Author in Posts immediately hides it in Pages; Date behaves the same way in either direction and the choice survives reload.

This can look like a separate Pages setting changed on its own.

## How to reproduce

1. Open native Posts and native Pages.
2. In Posts, open **Window actions → Show columns**.
3. Hide Author.
4. Inspect Pages and its Show columns menu.
5. Author is hidden and unchecked there too.

## What should change

Either store separate choices for Posts and Pages, or make the shared scope explicit in both menus:

- **Author (shared by Posts and Pages)**
- **Date (shared by Posts and Pages)**

PR #581 takes the second approach and also names each menu section **Show columns in Posts** or **Show columns in Pages**.

Posts and Pages silently share Author and Date visibility

nickhamze:
## Why this change

We reviewed every setting surfaced by the Settings Clarity run and found a recurring problem: the controls often say *what* they change, but not **who the change affects, when it takes effect, what else it changes, or how to reverse it**.

That makes ordinary choices feel riskier than they should. A person should not need to inspect the code—or learn by surprise—that a preference follows their WordPress account, a theme changes other preferences the first time it is selected, a native-window choice applies only when the app is opened again, or a reset reaches every Preferences tab.

The audit covered 137 settings. Of those, 129 are real OpenStation product settings and 8 are deliberately planted test-lab canaries. The canaries remain in the lab; this PR updates the product language only.

## What changes

This pass focuses on the shared language problems that made otherwise straightforward settings hard to predict:

- **Account and site scope are stated plainly.** Appearance and shortcut choices now say they are saved to the current WordPress account. Site-wide options say “for everyone.” Theme uploads say they install a theme for the whole site.
- **Timing is honest.** Native Posts, Pages, Users, Plugins, and Comments choices now say they apply the next time the app opens and do not replace a window that is already open.
- **Reset is no longer vague.** “Reset to defaults” becomes “Reset all OpenStation preferences,” with an explanation that it reaches every Preferences tab, clears the saved custom-wallpaper choice, keeps the Media Library file, and cannot be undone.
- **Theme side effects are disclosed.** The Themes tab explains that the first selection applies the theme’s recommended preferences. The reapply action no longer promises “layout and effects” when a theme may recommend a different set of preferences.
- **Shortcut placement is easier to understand.** “On both” and “Hidden” become “Desktop and dock” and “Nowhere,” and each picker is named for the shortcut it controls.
- **File Associations describe the actual decision.** Every row now reads “Default app for …,” including the accessible control name, and explains that the choice belongs to the account and is used the next time a file opens.
- **Mio’s abstract controls use visible outcomes.** “Hue spread,” “Iridescence,” “Spin the gradient,” and “Restore Mio” become “Colour range,” “Holographic strength,” “Move gradient around ring,” and “Reset Mio’s look.” Zero-value controls say when zero means off.
- **Shared Posts and Pages column choices are disclosed.** The menu identifies whether it belongs to Posts or Pages, while Author and Date say they are shared by both.

Representative examples:

| Before | After | Why |
|---|---|---|
| “saved to this browser” | “saved to your WordPress account” | The preference follows the signed-in user across browsers. |
| “Reset to defaults” | “Reset all OpenStation preferences” | The action reaches far beyond the visible Appearance tab. |
| “Use the native Posts window” | “Open Posts in the native window” | This is a routing choice used on the next open, not a live conversion of the current window. |
| “Apply recommended layout and effects” | “Apply recommended preferences” | Themes can recommend keys outside those two categories. |
| “Show in” | “Where to show [app]” | Each repeated picker now has a clear, distinct question. |

## What this PR does not claim to fix

This is intentionally a language-focused PR. The audit also found behavior bugs—failed-save rollback, rapid-toggle races, Spatial-layout placement, live opener refresh, ribbons that remain visible, and similar problems. Those need code fixes and are being tracked separately. The copy here does not disguise or close them.

## The 16 findings from the audit

Every finding has its own destination now, so this PR does not become a catch-all bug report.

This PR fully resolves the wording or accessible-name problem in:

- Closes #584
- Closes #587
- Closes #589
- Closes #590
- Closes #591
- Closes #596

It improves the explanation, while leaving the remaining interaction decision open, in:

- #582 — preview or confirm the extra preferences applied on first theme selection
- #583 — add confirmation and recovery to the full-preferences reset

The behavior bugs remain separate:

- #552 — rapid Extended Options changes save the wrong value
- #585 — failed shortcut saves leave the desktop and dock out of sync
- #586 — Spatial layout ignores shortcut placement choices
- #588 — late file openers do not refresh the mounted Preferences tab
- #592 — Mio's holographic checkbox and strength control can disagree
- #593 — failed Posts/Pages column saves leave stale UI
- #594 — Mio's idle-wobble checkbox and strength control can disagree
- #595 — post-status ribbons remain after the setting is off

## Checks run

- `npm run lint`
- `npm run typecheck`
- `npm run test:js` — 4,260 tests passed
- `npm run build`

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-581-fe971359d16be77a14d636ed6c86a5f53119e9cd.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
@nickhamze I think we can close this as @nuriapenya took your idea pretty well!

PR #601, opened 2026-08-13
Woo relations: a refund is not a purchase; user tiles: carry their key

### LINKED ISSUE #598: Fatal error on Product edit screen — WC_Order_Refund doesn't have get_order_number()
The "Related entities" feature (WooCommerce integration) throws an uncaught fatal error on the Product edit screen for any product that has at least one refund on record.

Error:

PHP Fatal error: Uncaught Error: Call to undefined method WC_Order_Refund::get_order_number()
in .../desktop-mode/includes/my-wordpress/integrations/woocommerce-relations.php:762
Stack trace:
#0 .../woocommerce-relations.php(1067): openstation_my_wordpress_woo_product_related()
#1 .../class-wp-hook.php(341): openstation_my_wordpress_woo_related_entities()
...

Root cause: openstation_my_wordpress_woo_orders_with_product() queries wp_woocommerce_order_items / wp_woocommerce_order_itemmeta for any order_id with a _product_id/_variation_id meta match. Refunds store their own line items in the same tables under the refund's own post ID (post_type shop_order_refund), so a refunded product's "orders" list can include refund IDs.

openstation_my_wordpress_woo_product_related() (and the equivalent openstation_my_wordpress_woo_coupon_related()) then calls wc_get_order($order_id) and guards with:

```
if ( ! $order instanceof WC_Abstract_Order ) {
    continue;
}
```

But WC_Order_Refund also extends WC_Abstract_Order, so it passes this guard. The very next lines call $order->get_order_number(), which is defined on WC_Order but not on WC_Order_Refund or the abstract base — fatal.

Fix: Tighten the instanceof check from WC_Abstract_Order to WC_Order in both spots (woocommerce-relations.php lines ~735 and ~931). This correctly excludes refund rows from the "recent orders / who bought it" and "used on" lists (which is also more semantically correct — a refund isn't a purchase).

```
// before
if ( ! $order instanceof WC_Abstract_Order ) {
// after
if ( ! $order instanceof WC_Order ) {
```

Impact while unpatched: Because the fatal fires during the admin_footer hook (via the chromeless-bridge script), the Product edit screen partially renders but breaks: the "Add New" button is missing, the description editor gets stuck showing raw HTML with no Visual/Code tab toggle, and Save/Update stops working — for any product with a refund history.

```
Plugin version:    1.0.1
WordPress version: unknown
Viewport:          2367x1393
Platform:          MacIntel
Current URL:       https://oceansev.com/wp-admin/index.php?desktop_mode_portal=1&desktop_mode_portal_intent=1
```

</details>

AllTerrainDeveloper:
Two unrelated bugs, both in My WordPress.

## The Product edit screen fatals on any refunded product

Fixes #598.

Refunds keep their own line items in the same `woocommerce_order_items` tables, under the refund's id, so the lookup that asks those tables *"which orders contain product X"* answers with refund ids too. `WC_Order_Refund` extends `WC_Abstract_Order` and sails through the guard; the `get_order_number()` call two lines later is a `WC_Order` method the abstract base does not declare.

It fires during `admin_footer`, so the screen half-renders: no **Add New**, the description editor stuck on raw HTML with no Visual/Code tabs, Save dead. Same shape in `openstation_my_wordpress_woo_coupon_related()`.

**Not** fixed by tightening the guard to `instanceof WC_Order`, which is what the report suggested — that exact test is already documented one file over (`woocommerce.php:1260`) as having silently emptied the Orders folder on stores with custom order types:

> `WC_Abstract_Order`, not `WC_Order`: that's the type every order class actually extends […] Testing against `WC_Order` silently dropped every row on some stores while `total` still reported hundreds — an empty folder with a confident count on its tile.

So `openstation_my_wordpress_woo_is_purchase()` excludes the one known-hostile subclass and then asks the object whether it declares the accessor these lists call. An exotic order type extending the base without one is dropped rather than fatal, and every normal order type still lists.

Dropping refunds is also the truer answer: *"who bought it"* and *"where was this coupon used"* are questions about purchases, and a refund is the undoing of one.

### One thing beyond the report

The id list is `ORDER BY order_id DESC`, and a refund always outranks the order it refunds, so **refunds sort first**. A `LIMIT 10` on a much-refunded product could come back as ten refunds and — once filtered — an empty "who bought it" group, on exactly the product whose history a merchant most wants to read. The loops now read 40 candidates and stop at 10 kept. Group budgets are unchanged.

## Clicking a person in the WP Explorer does nothing

Reported separately: in **WOO → Customers**, a single click on a customer leaves the right pane empty and doesn't even highlight the tile.

The users canvas keys its selection on `data-entry-id`, and `buildUserTile()` only ever set `data-user-id`. So `keyOf` returned null for every person tile, the controller never recognised it as an item, and the click fell through to the background handler *that clears the selection*. No highlight, no status count, and a preview pane that never paints, because it only paints from a selection.

Nothing Woo-specific about it — Customers renders through the built-in `user` kind, so **the built-in Users list has the same bug**. Regression from #501, which moved these canvases onto the shared selection framework.

The trash and bulk-remove paths query the same attribute, so they were quietly no-oping on users too.

Giving user tiles an entry id does make post 12 and user 12 collide on the `os-my-wordpress-entity-trashed` broadcast, which walks every live window body — trashing a post in one window would have removed a stranger's tile from a Users grid in another. The canvases carry `data-entity-id` now and the pruner checks the section before removing anything.

## Tests

- `tests/vitest/my-wordpress-selection.test.ts` — a `user`-kind section in the harness: the tile carries its key, single click selects and reports "1 selected", the preview pane paints the person, ctrl-click holds two, and a trash broadcast prunes only its own section. All five red before the fix.
- `tests/phpunit/tests/myWordpressWoocommerce.php` — the purchase guard against a stubbed `WC_Abstract_Order` hierarchy (the suite has no WooCommerce): an order passes, a refund doesn't, a bare subclass doesn't, and neither do `null` or a stray object.

Full suites green: `test:js` (4217), `test:php` (2165), `lint`, `lint:php`, `typecheck`, `build`.

## Manual QA

1. **WP Explorer → WOO → Customers** — single click a customer: tile highlights, status bar says "1 selected", right pane shows the dossier plus the customer panel. Ctrl/Cmd-click a second → "2 selected".
2. **WP Explorer → Users** — same behaviour.
3. **Product edit for a product with a refund** — screen renders fully, Add New present, editor has its Visual/Code tabs, Save works, and the orders group lists orders only.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-601-0ac556a03bd363c5fdfaa086e05f759bce715e07.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/my-wordpress/integrations/woocommerce-relations.php
+++ b/includes/my-wordpress/integrations/woocommerce-relations.php
@@ -60,6 +60,25 @@ defined( 'ABSPATH' ) || exit;
 const OPENSTATION_WOO_RELATION_ITEM_CAP = 20;
 
 /**
+ * How many orders the product and coupon groups list.
+ */
+const OPENSTATION_WOO_RELATION_ORDER_CAP = 10;
+
+/**
+ * How many order-item rows to read to fill that list.
+ *
+ * The id lists come out of `woocommerce_order_items`, which holds
+ * refund rows alongside order rows — and refunds sort *first* there,
+ * since the query orders by descending id and a refund is created
+ * after the order it refunds. A `LIMIT 10` on a much-refunded product
+ * could therefore come back as ten refunds and no orders at all, and
+ * the group would render empty on the one product whose history a
+ * merchant most wants to read. Reading a few times the budget and
+ * stopping at the cap costs one bounded query.
+ */
+const OPENSTATION_WOO_RELATION_ORDER_CANDIDATES = 40;
+
+/**
  * Query flag marking a person-URL as a request for a *particular*
  * view of that person rather than for the profile editor.
  *
@@ -128,6 +147,44 @@ function openstation_my_wordpress_woo_current_order() {
  */
 function openstation_my_wordpress_woo_can_read_orders() {
 	return true === openstation_my_wordpress_woo_orders_permission();
+}
+
+/**
+ * Whether an object read back from an order-item row is a purchase.
+ *
+ * Refunds keep their own line items in the same
+ * `woocommerce_order_items` tables, under the refund's id — so a
+ * lookup that asks those tables "which orders contain product X"
+ * answers with refund ids too, for any product that has ever been
+ * refunded. `WC_Order_Refund` extends `WC_Abstract_Order`, so the
+ * usual guard waves it through, and the next line asks it for
+ * `get_order_number()`: a `WC_Order` method the abstract base does
+ * not declare, and therefore a fatal on the product edit screen.
+ *
+ * Dropping refunds is also the truer answer. "Who bought this" and
+ * "where was this coupon used" are questions about purchases, and a
+ * refund is the undoing of one.
+ *
+ * Deliberately *not* `instanceof WC_Order`. The abstract base is the
+ * type every order class actually extends, including HPOS's overrides
+ * and whatever custom order type a store registers — testing against
+ * `WC_Order` has already been tried elsewhere in this integration and
+ * silently emptied lists on stores that use them. So this excludes the
+ * one known-hostile subclass and then asks the object directly for the
+ * accessors these lists call, which keeps an exotic order type that
+ * extends the base without them out of a fatal too.
+ *
+ * @param mixed $order Whatever `wc_get_order()` returned.
+ * @return bool
+ */
+function openstation_my_wordpress_woo_is_purchase( $order ) {
+	if ( ! $order instanceof WC_Abstract_Order ) {
+		return false;
+	}
+	if ( $order instanceof WC_Order_Refund ) {
+		return false;
+	}
+	return method_exists( $order, 'get_order_number' );
 }
 
 /**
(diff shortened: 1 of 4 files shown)


AllTerrainDeveloper:
@claude can you review?

Issue #560, opened 2026-08-11
Full product audit, Part 2: 11 bugs in settings, sharing, navigation, and large-data behavior


nickhamze:
This is the follow-up to #559. The point of this pass was to catch the things users usually find first: the second quick click, the stale browser tab, the narrow screen, the shared folder after reload, and the site with enough content to make a visualization fall over.

I updated to current `origin/trunk` (`c670170`) before the final verification pass, then replayed every candidate from the older checkout. Anything already fixed upstream was removed from this issue.

The bugs below still reproduce on current trunk. The two settings problems are the most urgent because OpenStation tells the user everything is saved while silently keeping the wrong value. Screenshot evidence is attached in the first comment.

## 1. A quick second toggle in Extended Options is silently lost

**Why this matters:** The control ends in one state, the server saves the opposite state, and the user is not told anything went wrong.

**To reproduce:**

1. Open OpenStation Preferences → Features as an administrator.
2. Toggle **Enable drag-and-drop in the Media Library** off and immediately back on while the first save is still running.
3. The checkbox finishes enabled.
4. Reload or read `desktop-mode/v1/extended-options`; the server persisted disabled.

The save function returns early while a request is in flight, but the still-enabled control keeps accepting changes and never queues a trailing save.

**Where to look:** `src/settings/sections/extended.ts`

## 2. Two open sessions can overwrite each other’s unrelated settings

**Why this matters:** A setting that already showed as saved can be silently undone by a stale tab or another device changing something unrelated.

**To reproduce:**

1. Open OpenStation in two browser sessions for the same user.
2. In session A, change the wallpaper and wait for it to save.
3. Without reloading session B, change only the accent color and wait for it to save.
4. Reload. Session B’s stale full-settings snapshot restores the old wallpaper while keeping the new accent.

Each save replaces the complete server-side settings object. There is no revision check, conflict response, field merge, or patch request.

**Where to look:** `src/settings/state.ts`, `includes/os-settings.php`

## 3. Owned folder windows lose their Share button after reload

**Why this matters:** Folder sharing works on the server, but the owner loses the main way to manage it from the folder window.

**To reproduce:**

1. Enable folder sharing and create or share a folder.
2. Reload the owner’s desktop.
3. Open the folder from the wallpaper.
4. The title bar has its normal window and breadcrumb controls, but no **Share folder** action.

The button matcher requires folder ownership from the client folder store. Normal persisted-placement hydration does not populate that store.

[...]

PR #602, opened 2026-08-13
Audit part 2: silent settings loss, folder-sharing visibility, and wallpaper shortcuts

AllTerrainDeveloper:
Part of #560 — fixes items **1, 2, 3, 4 and 8**, plus one bug found while verifying #3/#4 on a live install. The other items in that audit are untouched, so this doesn't close it.

Every fix ships with a regression test at the owning layer, and each test was verified to fail with the fix reverted.

---

## 1. A quick second toggle in Extended Options is silently lost

`save()` bailed on `state.saving` and queued nothing, while the checkboxes stayed live and `<os-checkbox-label>` flipped its own attribute before anything else ran. A second click mid-flight moved the control, moved `state`, and never reached the server — then the success branch repainted from `state` while `ctx.config.extendedOptions` quietly held what the server actually kept.

Track that a control moved during a request and re-run the save when it settles. The trailing request rebuilds its body from `state`, so any number of mid-flight changes — across different controls — coalesce into one POST that ends on what's on screen. It's awaited rather than fired off, so the promise the first call returns settles only once the server agrees with the panel.

I deliberately did **not** disable the checkboxes while saving: it would narrow the window without closing it, and locking a site-wide toggle for a round-trip is worse than coalescing.

## 2. Two open sessions can overwrite each other's unrelated settings

**Client** — diff the live state against `_lastConfirmedState` (the rollback baseline the module already maintained for failed saves) and post only the fields that moved. A session that never touched the wallpaper can't express an opinion about it, because the key isn't in the request. A save with nothing changed now makes no request at all.

The baseline stays *this session's* last-confirmed state rather than the server's current truth — diffing against fresh server state would read another session's change as a local edit and post the stale value straight back.

**Server** — `POST /os-settings` merges the payload over the user's stored settings, so an absent key keeps its stored value instead of resetting to the shipped default. A full payload behaves exactly as before.

Two things worth flagging for review:

- The merge is in the REST handler and **not** in `openstation_save_os_settings()`, whose contract is replace. Migration 1 in `includes/migrations.php` `unset()`s keys and re-saves precisely so the sanitizer backfills the new defaults; a merging saver turns it into a silent no-op — invisible to PHPUnit, which builds fresh meta every run, and visible on real installs as a migration that did nothing.
- Merging is one level deep. A request that sends a map-shaped field (`wallpaperSettings`, `itemVisibility`, `dockOrder`, `dockPromotedPositions`) replaces that whole map, because deep-merging leaves no way to delete an entry.

This is not a conflict protocol — no revision token, no 409. Two sessions editing the *same* field is still last-write-wins, which is the right trade for per-user preferences.

## 3 + 4. Folder sharing: the Share button and the shared badge

Both were the desktop missing a fact, not the server missing a feature.

**The Share button** gates on ownership from the client folders map, and nothing on the normal boot path filled it — placement hydration populates placements; `listFolders()` only ran after a create, a rename or an untrash. Every folder read as owner `0`, the gate compared `0` to a real user id, and after any reload the owner lost the control.

Inlined the viewer's visible folders into the shell config as `filesBootFolders`, mirroring `filesBootPlacements` — same shape and visibility resolution as `GET /folders`, so owned folders, accepted shares and `share_mode='all'` all arrive. `seedBootFolders()` applies them once and deletes the key, so a later re-hydration still fetches fresh state.

**The badge** reads `placement.file.shareSummary`, and a placement's `file` is whatever `OpenStation_Folder_File::serialize()` returns — which carried `ownerId` and `shareMode` but never the summary. It was reading a key that was never on the wire. The tile menu's "Manage sharing…" label read the same missing key, so it always said "Share folder…".

Extracted `openstation_files_folder_share_summary()` and call it from both the folder response shape and the file shape, so the two paths can't drift apart again. The manage check costs a folder read of its own and an unshared folder counts zero recipients for everyone, so it's skipped unless there's something to count — every folder tile serializes through here.

## 8. Spatial desktop shortcuts ignore enabled native windows

Narrower than the report suggests: `desktop-shortcuts-sync.ts` synthesizes the tile correctly and `desktop-layout.ts` only decides which items get synthesized. The defect is in `built-in-openers.ts` — `desktop-mode-shortcut-opener` went straight from the same-origin check to `windowManager.open()`, never consulting the registry that `Dock.openPage`, the constellation router, the link interceptor, session restore, the Related menu and both command-palette bridges all consult.

It now calls `tryNativeUrlRemap()` between the cross-origin check and the window-id derivation, the order `Dock.openPage` uses. Resolution stays at click time rather than being baked into the synthesized placement, so the registry's `enabled` gate reads the live OS Settings snapshot and turning a native window off falls through to the iframe with no re-synthesis.

## Bonus: a tile filed into a folder could disappear

Found while verifying #3/#4 against a live install. Dragging a desktop tile onto a folder icon made it vanish — from the desktop and from inside the folder.

The placement was there the whole time, at the coordinates it held on the wallpaper. The live row read `parent=19 user:23 x=16 y=616` while every other in-folder tile sat at `y=16`. No folder window is 616px tall, and `.os-files-layer` is `position: absolute; inset: 0` with no scroll — so the tile wasn't below a fold, there is no fold. It was unreachable.

The folder-tile drop target had two branches: the `shortcut` branch packs row-major into the destination (documenting why — the folder window is probably not mounted, so landing top-left is what's visible whatever size it turns out to be), while the `desktop-file` branch sent `{ parentId }` alone and let the source coordinates ride along. It now packs the same way.

`reflow()` only ever tested the right edge, which a scrollless layer makes symmetrical. It now tests the bottom too, and runs after every paint rather than only on a width change — each paint writes stored coordinates back onto the tiles, so a stranded tile needs rescuing again after every one. Still visual only, persisting nothing until the user drags or sorts, so a layout that merely doesn't fit the current window size is restored the moment there's room. That's what lets an existing desktop heal itself instead of needing a migration.

---

## Tests

| Suite | Coverage |
|---|---|
| `tests/vitest/extended-options-save.test.ts` | 5 — the off/on race, several mid-flight toggles across two controls coalescing, a failed first attempt whose error is cleared by the trailing one, the no-endpoint no-op |
| `tests/vitest/os-settings-partial-save.test.ts` | 9 — only-changed-field payloads, an untouched field absent from the wire, nested/array fields diffed by value not identity, no-op saves making no request, baseline advance, retry after failure |
| `tests/phpunit/tests/osSettingsRest.php` | 7 — partial merge, omitted key ≠ default, explicit `false` still turning a setting off, full-payload back-compat, and that the saver itself still replaces (the migration contract) |
| `tests/phpunit/tests/desktopFilesShareSummary.php` | 11 — badge for owner *and* recipient, count owner-only, pending ≠ shared, `share_mode='all'`, both shapes agreeing, boot payload including a recipient seeing the real owner |
| `tests/vitest/desktop-files-share-button.test.ts` | 9 — seeding, one-shot behaviour, owner/recipient gate, and one that imports the real entry point so deleting the wiring line fails |
| `tests/vitest/desktop-files-file-into-folder.test.ts` | 3 — the move re-packs instead of carrying `y: 616`, an out-of-bounds tile reflows into view while the stored value is untouched, a tile that fits is left alone |
| `tests/vitest/desktop-shortcut-native-remap.test.ts` | 4 — an enabled remap claims the URL with no iframe alongside, a *disabled* one still falls through, an unclaimed URL is unchanged, `shortcutWindow` untouched |

Full runs: **4247 JS** across 343 files, **2183 PHP**. `npm run build`, `lint`, `typecheck` and `lint:php` all clean.

## Docs

- `docs/architecture.md` — new **Preference persistence** section: partial payloads, why the diff baseline is the session's own last-confirmed state, and the three things this explicitly is not (conflict protocol, live sync, deep merge).
- `docs/folder-sharing.md` — **What the desktop can see about a share**: `shareSummary` semantics and `filesBootFolders`.
- `docs/files-on-desktop.md` — folder-drop re-packing and the `reflow()` contract; the `filesBootFolders` seed alongside `filesBootPlacements`.
- `docs/hooks-reference.md` — the remap registry claimed *every* URL-opening path consulted it, which was false for exactly this surface. Corrected, with wallpaper shortcuts named.

## Verification

Items 1–4 and the folder-tile bug were reproduced and re-tested by @prismiwi2015 on a live install; the `y: 616` row above is from that database. Item 8 is covered by unit tests and is ready for a browser replay (Spatial layout, native Posts enabled, click the Posts wallpaper tile).

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-602-3349d98bb96f5f7ab2ca801b1a68f1d3bc1469e0.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
@claude can you review?

AllTerrainDeveloper:
Both addressed in 275bc74a — thanks, they were cheap to close and both sit right on this PR's theme.

**Non-array payload.** You're right that the test pinned the wipe as correct, which is the wrong thing to enshrine in a PR about not destroying what wasn't sent. A payload that isn't an object says nothing about any field, so the handler now returns early and changes nothing. Still unreachable over real REST traffic thanks to the schema, but the trap is gone for a direct caller. `test_non_array_payload_is_survivable` became `test_non_array_payload_changes_nothing` and now asserts a previously-stored value survives.

**The `_lastConfirmedState` comment.** This turned out to be more than a doc nit, because this PR gave that baseline a second job: it now decides what a save is *allowed to mention*. Priming it from the localStorage cache would mean treating values a previous session never got as far as saving as already-confirmed — and therefore never sending them. A field silently stuck on one machine is a quieter version of the bug the diff exists to fix.

So rather than reword the comment, I moved the priming into `loadState()` on the server-snapshot path and nowhere else, and dropped the unconditional call from the `OsSettings` constructor. Left unprimed, the first save posts the full snapshot and the divergence heals itself; every save after that is a diff again.

The cost is that the first failed save in a cache-sourced boot has no rollback target. That's the better trade — rolling back to unconfirmed cache data was never the intent, and after one successful save the baseline exists and rollback behaves as before.

Two new vitest cases cover the split (server-loaded boot → first save is a diff; cache-loaded boot → first save is the full snapshot), and `docs/architecture.md` now states the priming rule. Verified both new behaviours fail with the change reverted. Full suites still green: 4249 JS, 2183 PHP.

PR #613, opened 2026-08-14
Perf: load window, wallpaper and widget bundles on demand

### LINKED ISSUE #606: Unnecessary libraries loaded
Unnecessary libraries are always loaded, even when the features are disabled, leading to unnecessary bloat. These libraries include:

my-wordpress-woocommerce.min.js
living-tree-wallpaper.min.js
snow-wallpaper.min.js

```
Plugin version:    1.0.1
WordPress version: unknown
Viewport:          2367x1387
Platform:          MacIntel
Current URL:       https://bricks.infinitemonkeys.ca/wp-admin/index.php?desktop_mode_portal=1&desktop_mode_portal_intent=1
```

</details>

AllTerrainDeveloper:
Fixes #606.

The issue named three bundles. They turned out to be symptoms of one shape repeated three times: `openstation_enqueue_native_window_scripts()`, `openstation_enqueue_desktop_wallpaper_scripts()` and `openstation_enqueue_desktop_widget_scripts()` each looped their registry and called `wp_enqueue_script()` unconditionally, on every admin page the shell rendered.

None of it had to happen at boot. The shell reads a window's render callback off `window.openStationNativeWindows[ id ]` at **open** time; a widget's picker row is built entirely from PHP metadata; a wallpaper's swatch is a CSS string already in the payload. The bundles were being paid for and not used.

## Bytes transferred — one admin page load, stock install, admin user

Measured on the production build (`npm run build`), gzip at level 9. Only OpenStation's own JS; Core's admin scripts are the same either way.

### What stopped loading at boot

| Bundle | Loads instead when | Raw | Gzip |
|---|---|---:|---:|
| `my-wordpress.min.js` | WP Explorer opens | 325.1 KB | 87.0 KB |
| `posts-window.min.js` | Posts / Pages / Users / user-edit opens | 321.0 KB | 86.5 KB |
| `plugins-window.min.js` | Plugins opens | 199.8 KB | 52.2 KB |
| `games.min.js` | Games opens | 99.1 KB | 27.8 KB |
| `comments-window.min.js` | Comments opens | 94.4 KB | 26.2 KB |
| `recycle-bin.min.js` | Trash opens | 80.6 KB | 24.4 KB |
| `agent-run-window.min.js` | the agent runner opens | 71.5 KB | 20.9 KB |
| `content-graph.min.js` | Content Graph opens | 54.8 KB | 16.3 KB |
| **Native windows subtotal** | | **1246.3 KB** | **341.4 KB** |
| `living-tree-wallpaper.min.js` | applied, or the picker opens | 56.6 KB | 18.6 KB |
| `snow-wallpaper.min.js` | applied, or the picker opens | 41.4 KB | 12.4 KB |
| `animated-logo-wallpaper.min.js` | applied, or the picker opens | 5.4 KB | 2.6 KB |
| **Wallpapers subtotal** | | **103.4 KB** | **33.7 KB** |
| `widget-drafts.min.js` | the widget mounts | 45.0 KB | 13.0 KB |
| `widget-focus-timer.min.js` | the widget mounts | 40.0 KB | 11.7 KB |
| `widget-notes.min.js` | the widget mounts | 30.8 KB | 9.7 KB |
| `widget-heartbeat.min.js` | the widget mounts | 9.4 KB | 3.8 KB |
| `widget-jazz-quote.min.js` | the widget mounts | 5.5 KB | 2.6 KB |
| `widget-site-views.min.js` | the widget mounts | 5.5 KB | 2.2 KB |
| `widget-post-stats.min.js` | the widget mounts | 4.9 KB | 2.1 KB |
| `widget-recent-comments.min.js` | the widget mounts | 4.1 KB | 1.7 KB |
| `widget-starter.min.js` | the widget mounts | 2.6 KB | 1.2 KB |
| **Widgets subtotal** | | **147.8 KB** | **48.0 KB** |
| **Total deferred** | | **1497.5 KB** | **423.1 KB** |

### What starts loading at boot

| Bundle | Why | Raw | Gzip |
|---|---|---:|---:|
| `shell-overlays.min.js` | it never actually loaded before — see below | 86.1 KB | 22.1 KB |

### Totals

| | Before | After | Δ |
|---|---:|---:|---:|
| Raw | 2188.3 KB | 771.8 KB | **−1416.5 KB (−64.7 %)** |
| Gzip | 627.3 KB | 224.8 KB | **−402.5 KB (−64.2 %)** |

On a store with WooCommerce active, `my-wordpress-woocommerce.min.js` (46.3 KB raw / 13.2 KB gzip) also stops loading at boot and rides WP Explorer instead: **−1462.8 KB raw (−65.5 %)**.

What still loads at boot is `desktop.min.js`, `admin-bar.js` and `window-system.min.js`, unchanged in role — plus `shell-overlays.min.js`, which is new here only in the sense that it now happens at all.

### The overlays correction

An earlier revision of this description claimed `shell-overlays` was "already injected on demand and not affected". That was true of `window-system` and wrong about `shell-overlays`: its loader tested `customElements.get( 'os-confirm-dialog' )` to decide whether the bundle was in the tab, that tag also reached `desktop.min.js` through a leaked import, and so the check answered "already loaded" at boot and the bundle **was never fetched on any page** — on this branch or on trunk.

`<os-context-menu>` is registered by that bundle and by nothing else in the shell, so right-clicking the wallpaper or a desktop icon appended an inert element and opened no menu. It went unnoticed on trunk because `my-wordpress.min.js` was enqueued on every admin page and imports the same component. Deferring that bundle took the menus with it, which is how a standing bug surfaced as a regression here.

Fixed in `cca0e9a3`: readiness is now `window.openStationShellOverlays`, a flag the bundle's entry sets and nothing else can, mirroring `window.openStationWindowSystem`. So the 86.1 KB above is not a new cost so much as one that was being paid dishonestly — trunk was spending 332.9 KB of `my-wordpress.min.js` to get a context menu.

### Cost

`desktop.min.js` **shrinks**: 689 188 → 683 919 bytes raw (−5 269), 203 226 → 201 631 gzipped (−1 595).

The stub and forwarding logic this PR adds cost +1 771 raw / +494 gzipped. Closing the four leaks that hid the overlays bug more than paid that back: `dock-constellation` and `dock-peek` wanted one string, `ITEM_MENU_OPENING_EVENT`, and imported it from a lazy bundle's *entry*, pulling that entry's tree into main; three `desktop-files` modules imported `osConfirm` from the component module rather than the `../os-confirm` wrapper that exists to keep the class out. That is −7 040 bytes of weight trunk was carrying too — a pre-existing bloat fix riding along, not an efficiency this PR earned.

Deferred bytes are not saved bytes — a user who opens WP Explorer still downloads it, once, on click, with the window's own spinner covering the fetch. The saving is on everything they *don't* open, on every page load.

## New public surface

All documented, all with tests.

- **`'scripts' => [ $handle, … ]`** on `openstation_register_window()` — companion bundles loaded in order immediately before the window's own script. For code that extends a window from outside it and must be subscribed before its render callback paints. `my-wordpress-woocommerce` is the in-tree user.
- **`'preload_script' => true`** — opt-out, for a bundle with a genuine boot-time job (a badge poller, a `wp.os` API surface). Nothing in-tree needs it.
- **`wp.os.loadWindowScript( id )`** — load a window's bundle without opening the window, for reaching an API the bundle publishes. Deduped by URL; concurrent callers share one `<script>`.

`wp.os.myWordpress`'s early stub now forwards `openDetail` / `openMedia` / `openUserFootprint` / `trashEntity` through that loader, so the recycle-bin drag-to-trash path keeps working with the bundle unloaded. `registerEntityKind` still queues and drains as before — it is synchronous by contract.

## Compatibility

Delivery is otherwise unchanged: `wp_localize_script` / `wp_add_inline_script` / `wp_set_script_translations` data is harvested off the registered handle into the boot payload and replayed as inline tags around the injected `<script src>`, in `wp_print_scripts` order. The `'config'` arg works on both paths. The `window.openStationNativeWindows` / `openStationWallpapers` / `openStationWidgets` contracts are untouched.

A third-party window bundle that does something at load time *other* than publish its render callback now does it later. `docs/migration-lazy-window-scripts.md` covers that case and the two ways out (split the bundle, or `preload_script`). Per-tab scripts registered via `openstation_register_window_tab()` stay eager — the shell has no lazy path for them.

The native-window enqueue hook moves from `admin_enqueue_scripts:20` to `:5`, because `openstation_enqueue_assets()` builds the payload at 10 and data attached after that would ship a bundle with no config. Same reason for the WooCommerce config attach.

## Tests

25 for the deferral: `nativeWindowLazyScript.php` (11), `native-windows-lazy-script.test.ts` (10 — including "reads the render callback *after* the load" and companion ordering), `wallpapers-lazy.test.ts` (9), `widgets-lazy.test.ts` (5), plus two additions to `myWordpressWoocommerce.php`.

10 more for the overlays bug, one file per half of it:

- `src/shell-overlays/loader.test.ts` (5) — behavioural. `tests/vitest/setup.ts` pre-registers the whole component kit, so these run in exactly the poisoned state that broke production: a registered `os-confirm-dialog` must not stop the loader fetching.
- `tests/vitest/shell-overlays-boundary.test.ts` (5) — structural, so the *cause* can't return either. Pins the readiness flag, forbids `customElements` anywhere in the loader (no tag can prove which bundle loaded), and walks the import graph from `src/desktop.ts` to hold the main bundle's reach into the overlay kit to a named allowlist.

Reintroducing the tag sniff fails four of them; reintroducing the entry import fails one more, naming the offending file.

Green: `typecheck`, `lint`, `lint:php`, `test:js` (4345), `test:php` (2209).

## Manual QA

1. Hard-reload `/wp-admin/` — none of the bundles in the table above should appear in the network log.
2. Open WP Explorer, Posts, Plugins, Comments, Trash, Games — each fetches on click and renders normally.
3. Preferences → widget picker lists everything correctly; enabling one mounts it; an already-enabled widget still appears on boot.
4. Preferences → Appearance: all wallpaper tiles present, live previews animate. Select Living Tree, reload — paints on boot.
5. Right-click a post tile → Move to Trash; drag a post shortcut onto the bin; Users → Edit user (opens the clicked user, not the viewer).
6. Right-click the wallpaper, a desktop icon and a dock icon — each opens its menu. `shell-overlays.min.js` appears in the network log shortly after first paint. Toasts, confirm dialogs and window-chrome buttons were riding on the same accident, so they're worth a glance too.

## Follow-ups, not in this PR

- The WooCommerce config still computes customer/order band counts on every admin page — a full pass over the user base for a bundle that may never load. Pre-existing; worth moving behind REST.
- `registerPalette` still has no server-side registration, so a plugin's palette needs an F5. Same fix shape as commands if it comes up.
- Four overlay components (`os-toast`, `os-button`, `os-window-button`, `os-save-status`) still reach `desktop.min.js` from real render sites in the notes layer, the share-settings modal and the wallpaper settings section. Harmless now that readiness is a flag, but it is weight on every page, and the boundary test's allowlist is where they are named.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-613-c0593fe1b0f24cb943f4d828146fa34813d57453.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
@claude can you review this? is it a hazard for the stability of the plugin?

AllTerrainDeveloper:
@claude you were not accurate, I was able to spot by myself regressions. The first I found is the Context Menu Options in the wallapper and icons, those were missing now. Check the last commit, and be more accurate

PR #615, opened 2026-08-17
Windows: guard focus(), unstick the opening class, add child windows

AllTerrainDeveloper:
Three related pieces of window-integrity work. The first two are fixes for reported bugs, both of which produced the same symptom — a window that is present, painted and unusable. The third is a new API that came out of the same conversation.

## `WindowManager.focus()` — argument guard

`focus()` took a `Window` with no runtime check, and pushed its argument onto `_stack` **before** calling `setZIndex()` across every member. So `focus( 'some-id' )` — the form [`javascript-reference.md`](docs/javascript-reference.md) and the comment in `src/built-in-commands.ts` already show — left a non-`Window` wedged in the stack, and *every later* `focus()` threw on it. Click-to-focus, dock activation and open-reuse were all dead until a reload, with nothing in the console pointing at the original bad call.

It now accepts `Window | string`, resolves ids, and rejects anything else **before** touching the stack. An unknown *id* stays a silent no-op, matching `raise()`: a window closing between the moment its id was captured and the moment focus is requested is a routine race, not a programming error. A non-Window, non-string argument warns.

Two corrections to the report this came from:

- **`raise()` is not affected.** It already takes a `string`, resolves via `getById()`, and returns early on `idx === -1`. The described `splice( -1, 1 )` on a miss cannot happen.
- **`conflict-toast.ts` was broken, but not by this.** It declared its own `declare global` asserting `window.openStation.windowManager.focus( id ) => Window` — a global the shell never defines (the namespace is `wp.os`) with a signature that matches nothing. Optional chaining meant it never threw and never poisoned the stack; "View folder" just silently did nothing. `share-menu-items.ts` carried the identical fabricated global, calling a `close()` that doesn't exist on the manager either, so leaving a shared folder quietly left its window open. Both now go through `wp.os`, and the folder close goes through the `Window`, where `close()` actually lives.

## `.os-window--opening` always comes off

Removal hung on a single once-only, unfiltered `animationend` listener. Two ways that fails:

1. **`animationend` bubbles.** A spinner, shimmer or holo drift finishing anywhere inside the window body consumed the listener and cut the open animation short.
2. **A hidden document never advances the animation**, so the event never arrives at all. The class is not cosmetic — its animation's `from` frame is `opacity: 0; scale(0.92)` — so a window opened in a backgrounded tab (session restore the user tabs away from, an `os-open-requested` from another surface, automation) stays painted invisible.

Now filtered on `animationName` + `target`, paired with `animationcancel`, backed by a timeout deadline, and skipped outright while `document.hidden` — nothing to animate for a document nobody is watching, so the stuck state can't arise at the source.

One correction here too: `.os-window--opening` carries **no `pointer-events: none`** in our CSS, and neither do unfocused windows. The reported "unclickable" framing isn't ours — that's either the third-party plugin's own CSS or the focus-stack poisoning above. The stuck-class hazard is real regardless; it just makes the window *invisible* rather than inert.

## New: child windows

`openChild( parentWindowId, config )` opens a **real** window — own chrome, drag, resize, minimize, taskbar entry — with one rule layered on top: **its owner can never sit above it.** Clicking the owner shakes the child and leaves focus there.

For the shapes a modal dialog gets reached for when what you actually want is a window: a full editor for one row of a list, a wizard beside the page it configures, a diff over the revision it belongs to.

```js
await wp.os.windowManager.openChild( 'edit-post-42', {
    id: 'my-plugin-seo-audit-42',
    url: '#seo-audit-42',
    title: 'SEO audit',
    icon: 'dashicons-chart-line',
    native: true,
    render: ( body ) => { /* … */ },
} );
```

The owner stays **fully usable** — scrollable, draggable, resizable, minimizable. Only its z-order is constrained; that's the reason to use a window rather than a dialog in the first place. The redirect lives inside `focus()`, so click-to-focus, dock, taskbar, alt-tab and open-reuse are covered by construction rather than per call site.

Surface: `openChild`, `ownerOf`, `childrenOf`, `blockingChildOf`, the `os-window-child-blocked` event, the `os.window.child-blocked` action, and `WindowConfig.parentWindowId` as the primitive.

Behaviors, each with a test:

| | |
|---|---|
| **Chains** | A child can own a child; focus goes to the deepest link. |
| **Close** | Closing an owner closes its children. A child with unsaved changes still gets to veto and becomes an ordinary window — better than discarding the user's work. |
| **Minimize** | Cascades down, and restore brings back *only* what the cascade put away. A child the user minimized themselves stays minimized. |
| **Minimized children** | Stop blocking. The user put it away on purpose, so the owner is theirs again. |
| **Session** | Children are excluded from snapshots, so a reload can't restore a child blocking an owner that failed to come back. |

Two subtleties carry comments where they live: the ownership stack pass **exempts minimized windows**, or hoisting an invisible child to the top would make `setFocused( i === length - 1 )` hand focus to a window nobody can see; and `blockingChildOf` carries a **cycle guard**, since a plugin declaring A-owns-B-owns-A would otherwise spin on the first click.

## Testing

28 new tests in `tests/vitest/window-focus-guard-and-opening-class.test.ts` and `tests/vitest/window-child-ownership.test.ts`.

- `npm run test:js` — 4458 pass
- `npm run lint` — clean
- `npm run typecheck` — clean
- `npm run build` — run

No PHP touched.

**Worth a manual pass before merge:** that the opening animation still looks right on an ordinary open, and that a child window's shake-on-blocked-focus reads as helpful rather than annoying. Those two are feel, not logic, and no browser was driven for this.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-615-a6fef11e2402eacbd623e5a3e18db41d805683cf.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/docs/api-index.md
+++ b/docs/api-index.md
@@ -225,6 +225,7 @@ Every event bubbles from `document`. See [`javascript-reference.md`](./javascrip
 | `os-window-content-loaded` | Stable |
 | `os-window-focused` | Stable |
 | `os-window-blurred` | Stable |
+| `os-window-child-blocked` | Stable |
 | `os-window-closing` | Stable |
 | `os-window-closed` | Stable |
 | `os-window-changed` | Experimental |
--- a/docs/event-driven-framework.md
+++ b/docs/event-driven-framework.md
@@ -103,12 +103,14 @@ stateDiagram-v2
 | `os-window-reopened`    | `{ windowId, baseId, wasMinimized, navigated }` — `navigated`: the request carried a URL the window wasn't showing, so the existing iframe navigated to it in place |
 | `os-window-focused`     | `{ windowId }` |
 | `os-window-blurred`     | `{ windowId, focusedTo }` |
+| `os-window-child-blocked` | `{ windowId, childWindowId }` — a focus request for `windowId` went to its child window instead; the owner cannot come to the front while the child is open |
 | `os-window-closing`     | `{ windowId, element }` |
 | `os-window-closed`      | `{ windowId }` |
 | `os-window-changed`     | `{ windowId?: string, reason: 'moved' \| 'resized' \| 'state' \| 'cascade' \| 'tile', state?: WindowState }` — batch-arrange dispatches (`'cascade'` / `'tile'`) omit `windowId`/`state` |
 
 Same payloads on `wp.hooks` actions:
-`HOOKS.WINDOW_OPENED`, `…_FOCUSED`, `…_BLURRED`, `…_CLOSED`,
+`HOOKS.WINDOW_OPENED`, `…_FOCUSED`, `…_BLURRED`, `…_CHILD_BLOCKED`,
+`…_CLOSED`,
 `…_MINIMIZED`, `…_RESTORED`, `…_MAXIMIZED`, `…_UNMAXIMIZED`,
 `…_FULLSCREEN_ENTERED/EXITED`, `…_REOPENED`, plus geometry events
 (`…_RESIZED`, `…_BODY_RESIZED`, `…_BOUNDS_CHANGED`,
--- a/docs/examples/README.md
+++ b/docs/examples/README.md
@@ -23,6 +23,7 @@ defined( 'ABSPATH' ) || exit;
 - [Accept drops on your desktop icon](./tile-drop-handler.md)
 - [Add an action that works on a whole selection](./multi-selection-action.md)
 - [Window lifecycle hooks (one subscriber per state)](./window-lifecycle.md)
+- [Open a child window its owner can't cover](./child-windows.md)
 - [Custom arrange-menu action](./arrange-action.md)
 - [Style a specific admin page inside the iframe](./chromeless-style-override.md)
 - [Window themes — per-window CSS variables](./window-theme.md)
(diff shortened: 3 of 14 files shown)


AllTerrainDeveloper:
@claude can you review?

PR #475, opened 2026-08-03
Rebrand (2/2): Desktop Mode becomes OpenStation

AllTerrainDeveloper:
Renames the product across code, controls, hooks and docs. The visible name is **OpenStation** everywhere, including the WordPress.org listing. Follows #473, which brought the palette, themes and mascot.

## What changed

| Before | After | Scope |
|---|---|---|
| `wp.desktop` | `wp.os` | JS namespace |
| `<wpd-*>` / `.wpd-*` | `<os-*>` / `.os-*` | Component kit |
| `--wpd-*` | `--os-ui-*` | Component tokens |
| `--desktop-mode-*` | `--os-*` | Shell tokens |
| `.desktop-mode-*` | `.os-*` | CSS classes |
| `desktop_mode_*()` | `open_station_*()` | PHP functions + hooks |
| `DESKTOP_MODE_*` | `OPEN_STATION_*` | Constants |
| `Desktop_Mode_*` | `Open_Station_*` | PHP classes |
| `desktopMode*` / `WpDesktop*` | `openStation*` / `OpenStation*` | JS identifiers |

The two token families kept **separate** prefixes on purpose. The shell and the component kit both declare `bg`, `fg-muted` and `border` at *different* values, so collapsing both onto `--os-*` would have merged the desktop wallpaper gradient with the component surface — exactly the failure mode `AGENTS.md` warns about.

## What deliberately did NOT change

The WordPress.org URL stays `wordpress.org/plugins/desktop-mode/`, so three strings are load-bearing and frozen:

- **Directory name** — it *is* the SVN slug (`release.yml` deploys to `BUILD_DIR: ./build/desktop-mode`).
- **`desktop-mode.php`** — `active_plugins` stores `desktop-mode/desktop-mode.php`; renaming it silently deactivates the plugin for every existing install on update.
- **Text domain `desktop-mode`** — language packs resolve against the directory slug, and a mismatch fails Plugin Check.

The plugin **name** is not one of them: WP.org reads that from `readme.txt` and the header, so the listing now reads OpenStation with the URL unchanged.

Persisted state is also untouched — options, user/post meta, transients, cron hooks, post types, REST namespaces, web-storage keys, **custom tables** (`{prefix}desktop_mode_file_placements` + 7 siblings), the `uploads/desktop-mode-files` and `-themes` directories, and **native-window / desktop-icon / file-opener ids** (stored in `dockOrder`, saved sessions, icon placements and file associations).

## One behavior change

[...]

PR #217, opened 2026-05-14
Shared folders, heartbeat widget, and heartbeat-pipeline hardening

AllTerrainDeveloper:
https://github.com/user-attachments/assets/02ab900f-ddee-45fe-998d-de9fa7311e9c

# Shared folders + Heartbeat widget

Two new features for Desktop Mode.

## Shared folders

Owners can share any folder with specific users or whole roles, at
read or read+write capability. Recipients accept, deny, or leave;
their decisions are private per-user (no "first to click decides
for everyone in the role").

**Capability model.** Owner → write. Direct user invite or accepted
role invite → grant on the shares table. Sub-folders inherit the
parent's grant via the ancestor walker. `share_mode='all'` still
works for site-wide folders and defaults to read.

**Shared namespace.** Every recipient sees the same icons inside a
shared folder; only `(x, y)` is per-user. Files the viewer can't
`can_read` (private posts, drafts, …) ship with `accessGated: true`
and render with a lock overlay + permission-explainer preview pane.

**REST surface.**

| Method | Path |
|---|---|
| `GET`    | `/folders/{id}/shares` |
| `POST`   | `/folders/{id}/shares` |
| `PATCH`  | `/folders/{id}/shares/{shareId}` |
| `DELETE` | `/folders/{id}/shares/{shareId}` |
| `POST`   | `/folders/{id}/shares/{shareId}/accept` |
| `POST`   | `/folders/{id}/shares/{shareId}/deny` |
| `POST`   | `/folders/{id}/leave` |
| `GET`    | `/files/users/search` |

Writes optionally accept `If-Match: <updatedAtMs>` and return a
structured 409 with `reason` / `actor` / `current` on conflict.

**UI.**

- Owner: title-bar People button + Share Settings modal
  (`<wpd-modal>`, `<wpd-user-search>`, `<wpd-role-picker>`,
  `<wpd-segmented>` for read / read+write per recipient, "Apply to
  sub-folders" checkbox).
- Recipient: invitation modal on pending invites, "Leave shared
  folder" tile action, shared folder windows append "· Shared" to
  the title.
- Destructive affordances are gated on a server-computed `canTrash`
  flag — read-only recipients don't see "Move to Trash" or "Move
  folder to Trash" and the recycle-bin drop target refuses their
  drags.

**Schema.** v9 adds `wp_desktop_mode_folder_shares` (polymorphic via
`target_type` for future non-folder sharing) and
`wp_desktop_mode_share_user_decisions` (per-user state for
[...]

PR #616, opened 2026-08-17
Files: make the conflict toast's "View folder" actually open the folder

AllTerrainDeveloper:
Follow-up to #615, which left this one open deliberately.

## The bug

The conflict toast's **View folder** button has never worked, since the file was added in #217. It was written against `window.desktopMode.windowManager.focus( id )` — a global the shell has never defined, with a signature matching no method the manager has. The rebrand (#475) find-and-replaced it to `window.openStation` and carried the dead reference forward. Optional chaining meant it threw nothing and did nothing: the button rendered, the user clicked it, and the desktop sat there.

#615 routed it through the real `wp.os` namespace, but without a way to open a folder window from a `parentId` alone it could only offer the button when that folder's window happened to *already* be open. Correct, and never wrong — but the affordance vanished in the common case, which isn't what the original author intended.

## The fix

The missing piece was never the window. It was a way to **name a folder** that the existing machinery already accepts.

A 409 knows two things: a folder id and a folder name. The opener registry wants a `DesktopFile`. `folderFileById()` is that bridge — so the toast dispatches through the ordinary `openFile()` and gets the registered `folder` opener, which means the window it opens is *the same window a tile double-click opens*: breadcrumb route stack, split preview pane, `· Shared` title cue, all of it.

```js
onClick: () => {
    void openFile(
        folderFileById( targetParentId, err.detail.current.parentName ),
    );
},
```

Reuse of an already-open folder window comes free: `openFile()` ends at `windowManager.open()`, which reuses by id. So the focus-if-open case that #615 special-cased is now just the default behavior, with no branch for it.

**What this avoids:** duplicating the window-building logic, and equally, extracting the opener's ~300-line inline render closure into a shared module just to call it from one more place. Neither was necessary — the indirection to do this already existed.

## `folderFileById`

Resolution order, best source first:

1. **The server's own serialized shape**, found among any loaded placement for that folder — real title and icon, including whatever a plugin's `openstation_files_serialize_file` filter did to them. This is the common case: a conflict arises from moving something around these very folders, so a tile for the destination is usually in loaded state.
2. **The folder row** in the store (authoritative name).
3. **The caller's fallback title**, then a generic label.

It's deliberately a **leaf** — store, registry, type constants, nothing else. `conflict-toast` is imported *by* `layer`, so reaching the opener implementation directly would have been an import cycle (`conflict-toast → … → layer → conflict-toast`). Going through the registry inverts the dependency, which is what that registry is for. Verified: no rollup circular-dependency warnings.

The folder icon fallback lives in `built-in-types.ts`, next to the folder type registration — a file whose docblock already explains why the PHP metadata is mirrored there ("a hard-coded fallback keeps pickers usable in the brief gap before the payload arrives"). It's only ever reached when no server payload describes the folder.

Both modules were already in `desktop.min.js`, so this adds no bundle weight.

## Edge case worth noting

Folder `0` stays button-less — it's the desktop itself, already on screen. That also covers the viewer-scoping in the 409 contract, where `current.parentId` degrades to `0` for viewers outside the actor's collaboration scope, so an out-of-scope viewer gets the message with no button rather than a button pointing at a folder they can't see. Documented in `docs/folder-sharing.md`, which already promised this button existed.

## Testing

10 new tests in `tests/vitest/desktop-files-folder-ref.test.ts` — resolution order (including a placement of a *different type* sharing the same ref, which naive ref-matching would open a folder window titled after a post), and the dispatch asserted against a genuinely registered opener rather than a mocked `openFile`.

One trap the tests document: the files store is a `createSharedStore`, so it lives on a global and `vi.resetModules()` does **not** clear it. Without an explicit `__resetFilesStoreForTests()` the fallback assertions read the previous test's placement — which is how I found it.

- `npm run test:js` — 4477 pass
- `npm run lint` / `npm run typecheck` — clean
- `npm run build` — clean, no circular-dependency warnings

No PHP touched.

**Manual check worth doing:** trigger a real 409 (move a placement in a shared folder from a second browser session, then move it from the first) and confirm **View folder** opens the destination with its breadcrumb and preview pane intact. I haven't driven a browser for this.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-616-83fe469f2aa55beb5873db8df377bfaf0d07dfc2.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/docs/folder-sharing.md
+++ b/docs/folder-sharing.md
@@ -286,6 +286,14 @@ display names or folder names.
 The JS client throws `FilesConflictError` and the layer surfaces a
 toast: *"Alice moved this to 'Backlog' just now. [View folder]"*.
 
+**View folder** opens the destination folder's window — the same
+window a double-click on that folder's tile opens, because the action
+dispatches through the registered `folder` opener rather than building
+a window of its own. An already-open folder window is focused instead
+of reopened. The button follows the viewer-scoping above: when
+`current.parentId` degrades to `0` there is no folder to point at, so
+the toast carries the message alone.
+
 Clients that omit the header retain last-write-wins semantics for
 back-compat. The OpenStation shell always sends the header.
 
--- a/src/desktop-files/built-in-types.ts
+++ b/src/desktop-files/built-in-types.ts
@@ -16,6 +16,19 @@
 
 import { registerType } from './registry';
 
+/**
+ * The folder tile's icon, mirroring
+ * `OpenStation_Folder_File::icon()` on the PHP side.
+ *
+ * Same rationale as the labels above: the server sends this on every
+ * serialized placement, so anything rendering a folder the server
+ * described uses **that** value and never this one. This is the
+ * fallback for the one case where a folder has to be addressed
+ * without a placement to describe it — see `folderFileById()` in
+ * `folder-ref.ts`.
+ */
+export const FOLDER_FILE_ICON = 'dashicons-portfolio';
+
 export function registerBuiltInFileTypes(): void {
 	registerType( { type: 'shortcut', label: 'Plugin shortcut', sort: 1 } );
 	registerType( { type: 'folder', label: 'Folder', sort: 5 } );
--- a/src/desktop-files/conflict-toast.ts
+++ b/src/desktop-files/conflict-toast.ts
@@ -10,26 +10,8 @@
 
 import { showToast } from '../toast';
 import { FilesConflictError } from './rest';
-
-/**
- * Structural slice of the public window manager, read off `wp.os`.
- *
- * `window.openStation` — which an earlier version of this module
- * declared and reached for — is not a global the shell ever defines;
- * the public namespace is `wp.os`. The optional chaining meant the
- * mismatch never threw, it just made "View folder" a button that did
- * nothing.
- */
-interface WindowManagerSlice {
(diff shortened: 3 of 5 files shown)


AllTerrainDeveloper:
@claude can you review?

PR #618, opened 2026-08-17
Update the wp.org screenshots for the current shell

nuriapenya:
The published set was captured before the rebrand and the brand palette, so it still shows teal window chrome and window titles the UI has since renamed.

Replaces all seven with nine current captures, adds Spaces and Games, and rewrites the captions to match.

Screenshots publish to wp.org from `.wordpress-org/` on the next stable release tag.


nuriapenya:
@AllTerrainDeveloper Hey Dani, these are the new set of wp.org screenshots. Could you review and merge it, and fold it into the next release, which is the step that actually pushes them to the plugin page? I don't think I can add them myself.

AllTerrainDeveloper (approved):
Nice!

PR #634, opened 2026-08-19
Explorer: render preview actions in every section, pass the entity to onSelect, and let a section own editing

AllTerrainDeveloper:
## What

Fixes the bug cluster found while user-testing the WP Explorer with a third-party plugin (AllTerrain Forms), plus the entity-argument ask that fell out of it:

- **Preview actions render everywhere.** Descriptors from `openstation_my_wordpress_preview_actions` used to render only in the media pane — a plugin scoping actions to its own section got nothing. The resolver/renderer pipeline now lives in a shared module (`src/my-wordpress/preview-actions.ts`) and renders in the post pane, the user pane, all three tile context menus (as entries the `os.my-wordpress.tile-context-menu` filter can still reorder/remove), and custom-kind renderers via a new `host.previewActionRow()` on `EntityRenderHost`.
- **`sections` matches post type slugs.** Auto-registered CPT sections have ids shaped `cpt-<post_type>`, which plugin authors had no way to know. `sections` now matches the section id, the section's declared post type slug, or `*`.
- **The init:99 snapshot trap is closed.** New Experimental filter `openstation_native_window_config` — a native window's `config` blob at emit time, applied at both serialization points (eager enqueue, lazy payload build). The Explorer uses it to re-collect `previewActions`, so a plugin adding its filter on `admin_init` now ships descriptor and script together instead of the script-without-descriptor symptom.
- **`onSelect( ctx )` carries the selected entity** — `ctx.item` (detail record in the pane, list row in menus), plus `itemId`, `postType`, and `surface`. All additive; the documented media contract is unchanged and pinned by a back-compat test. `MediaPreviewAction*` types stay as deprecated aliases.
- **`editAction` on the section descriptor** (`string | false`). A preview-action id replaces "Open in editor" at the pane's primary button, the menu's open entry (bulk fan-out runs `onSelect( ctx )` per item), and tile double-click — for types edited by a native window instead of `post.php`. `false` removes every edit affordance, reroutes double-click to the detail dossier, and suppresses the bulk "Edit…" modal (a *string* keeps that modal: it PATCHes over REST and doesn't involve the classic editor). If the named action didn't ship (capability) or was never wired, the edit affordances hide — the classic URL is known-broken for these sections.
- **Latent fix:** `fetchEntityDetail()` never requested `editUrl` or the section's `listFields`, so the pane's "Open in editor" ignored row-supplied editor URLs (the HPOS escape hatch). Detail requests now carry both.

## Behavior notes (all surfaces involved are Experimental)

- `os.my-wordpress.preview-actions` now fires for every kind — callbacks appending unscoped entries should guard on `ctx.kind` / use `sections`. Docs call this out.
- Tile context menus now include descriptor-derived entries (`sort: 50`, between navigation and destructive built-ins).
- Mid-session caveat documented: an already-loaded Explorer bundle keeps its config until F5 (`scriptL10n` executes only on first script load).

## Docs

`hooks-reference.md` (rewritten `preview_actions` entry with matching rules + timing; new `native_window_config` entry; `editAction` + `listFields` on the entities filter), `javascript-reference.md` (full `PreviewActionContext` shape, `previewActionRow`, descriptor table, cross-notes on the user-actions and tile-menu filters), `native-windows-proposal.md`, and `examples/my-wordpress-media-action.md` (non-media deep-link recipe + form-builder `editAction` walkthrough).

## Tests

- vitest: new `my-wordpress-preview-actions.test.ts` (matching by id/post-type/`*`, MIME fail-closed outside media, mounted post-pane render, `onSelect` ctx shape, menu wrapping, `isVisible`) and `my-wordpress-edit-action.test.ts` (resolution modes, row/menu exclusion, bulk "Edit…" gating, detail `_fields`, mounted pane replacement + dblclick fallback); back-compat ctx assertions in `my-wordpress-media-preview.test.ts`. Full suite 4519 passing.
- PHPUnit: late-registered filter reaches both emit paths and stays scoped to the Explorer window; `openstation_filter_native_window_config` on eager/lazy/non-array; `sections` and `editAction` pass through the collectors verbatim. Full suite green (2225 tests).
- `npm run build`, `lint`, `typecheck`, `lint:php` all clean.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-634-3f9f3571ccf4573ab1dca73ec13d3729704fe289.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/docs/examples/my-wordpress-media-action.md
+++ b/docs/examples/my-wordpress-media-action.md
@@ -2,11 +2,12 @@
 
 **Status: Experimental**
 
-The WP Explorer native window (Posts / Pages / Users / Media)
-exposes a uniform right-pane action surface. Plugins declare a
-descriptor on the **PHP** side (capability + MIME + script handle)
-and wire the JS handler via a `wp.hooks` filter. This recipe walks
-through both halves.
+The WP Explorer native window (Posts / Pages / Users / Media, and
+every plugin-added section) exposes a uniform action surface: one
+descriptor becomes a button in the right pane **and** an entry in the
+tile context menu. Plugins declare the descriptor on the **PHP** side
+(capability + MIME + script handle) and wire the JS handler via a
+`wp.hooks` filter. This recipe walks through both halves.
 
 The scenario: a "Compress this image" button that appears only on
 image items in the Media section, hits a plugin-owned REST route,
@@ -46,10 +47,19 @@ What you got for free:
 - `capability` is enforced before the descriptor ships to the
   bundle, so the button never appears for users who can't run it.
 - `mime` is re-evaluated client-side per item — the button stays
-  hidden on non-image rows.
+  hidden on non-image rows, and a MIME-scoped action never leaks
+  into a non-media section.
 - `sections` scopes the button. Omit it to show the button on every
-  section, or pass `array( 'media', 'posts' )` to opt in to more.
+  section, or pass an array to opt in: entries match a section's
+  **id** (`'media'`, `'posts'`), a section's **post type slug**
+  (`'atf-form'` — handy because an auto-registered CPT section's id
+  is `cpt-<post_type>`), or `'*'`.
 - `script` is auto-enqueued for users who can see the action.
+- The same descriptor also appears in the tile's right-click menu,
+  between the navigation entries and the destructive ones.
+- Timing is forgiving: descriptors are collected when the window
+  config is emitted, so registering the filter on `init`,
+  `admin_init`, or plain plugin bootstrap all work.
 
 ## JS — wire the click handler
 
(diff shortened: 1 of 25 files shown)


AllTerrainDeveloper:
@claude can you review this?

PR #638, opened 2026-08-20
Files: a drag out of the Media Library lands on the desktop

AllTerrainDeveloper:
Two fixes that turn up in the same place — the Media Library — and share a test surface.

## 1. The Explorer's Media grid stacked in one corner

`my-wordpress.css` overrides the canonical `.os-file-tile` chrome (`position: absolute`, a fixed 88×104 box) so media thumbnails can flow inside their CSS Grid. It does that at **equal specificity**, which means source order decides — and source order was left to chance.

`WP_Dependencies::all_deps()` walks the queue in order and pushes each handle's dependencies ahead of it, so a handle enqueued before the shell's own assets drags whatever it depends on along with it. `os-my-wordpress-woocommerce` is exactly that shape: `admin_enqueue_scripts` priority 5, dependent on `desktop-mode-my-wordpress`. On a store it printed `my-wordpress.css` first and `desktop-files.css` second, the overrides lost, and every tile in the Media grid went back to `position: absolute` with no offsets — one icon over another.

Declare the dependency instead, on all three stylesheets that write `.os-file-tile` rules.

`Tests_OpenStation_TileStylesheetOrder` replays the hostile enqueue order and asserts the chrome still prints first. Its last test generalises the rule to stylesheets that don't exist yet: any of the plugin's own CSS whose text (comments stripped) targets `.os-file-tile` must reach `os-files` through its dependency chain. That is what found the notes widget.

## 2. Dragging an image out of the Media Library did nothing on the wallpaper

Into Gutenberg it inserted a block. Onto the desktop it vanished. One cause.

A drag lifted inside an iframe never becomes a `DragManager` session — it belongs to the browser's native HTML5 machinery in the child document, so the shell sees no `pointerdown` and no `dragstart`. To deliver those gestures into another iframe, `iframe-drop-targets.ts` listens for `drop` on `document` in the **capture** phase. It claimed every drop it saw — `preventDefault` + `stopPropagation` + `stopImmediatePropagation` — and only then looked for an iframe under the cursor, returning silently when there wasn't one. Over Gutenberg there is an iframe. Over the wallpaper the gesture went in the bin before a single shell handler could see it.

Now the intercept resolves the iframe first and claims the drop only when it has somewhere to deliver it. With no iframe under the cursor it **declines**: cancels the browser default, tears the session down, leaves propagation alone. Cancelling the default is not optional even then — a media drag carries `text/uri-list`, whose default action on a plain document is to navigate, so a photo dropped on the desktop would otherwise replace the whole shell with the image.

`src/desktop-files/cross-frame-drop.ts` fills that opening: a files canvas — the wallpaper, a folder window's body — now accepts a native drag carrying a WordPress entity and files it as a shortcut, on the canvas or into a closed folder tile, the same as a pointer drag from WP Explorer. The payload comes from `wp.os.dragBridge` when the source announced itself and from the `application/x-wp-media-attachment` DataTransfer record when it didn't; a bridge kind with no registered file type is refused rather than guessed at.

`fileShortcutEntities()` is extracted from the two drop branches that already existed, so all three routes agree on row-major packing, the optimistic store upsert, and `os.files.shortcut-dropped`.

### The OS upload flow is untouched

A drag carrying `Files` is an upload, not a shortcut: the canvas returns before touching the event — no `preventDefault`, no `stopPropagation` — so it bubbles to `src/os-file-drop/` exactly as before. The `Files` check runs *ahead* of the bridge-payload check on purpose, so a stale payload from an earlier gesture can never capture a real upload. The intercept change can't affect Finder drags either: it only runs while a bridge session is in flight, and an OS file drag never starts one.

### Two boundaries that needed care

- Windows and the widget column float above `#os-area` and share its subtree, so a drop on a title bar is not a drop on the wallpaper. The surface check walks up from the event target and stops at the first answer.
- That check is **relative to the canvas host** — a folder window's canvas lives inside a `.wp-window` and must still accept.

## Tests

- `tests/vitest/desktop-files-cross-frame-drop.test.ts` — 20 tests: acceptance, the OS-file exclusion (both `dragover` and `drop`), surfaces that must not claim, folder windows, both payload channels, folder-tile targeting, teardown.
- `tests/vitest/iframe-drop-targets-bridge-drop.test.ts` — 5 tests on the intercept boundary. Verified as a real guard: reverting the source change fails 2 of them.
- `tests/phpunit/tests/tileStylesheetOrder.php` — 4 tests, including the generalised invariant.

4597 JS tests, 2239 PHP tests, `lint` / `typecheck` / `phpcs -n` clean.

Docs updated in the same change: a "Drops arriving from inside a window" section in `docs/files-on-desktop.md`, and "Drops the bridge declines" in `docs/bridge-protocol.md`.

## Verified manually

Media grid lays out; Media Library → wallpaper, → closed folder tile, and → open folder window all file a shortcut; Finder → wallpaper still opens the upload dialog; Media Library → Gutenberg still inserts.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-638-0967cc37c11314fdf3662e09f8e801cb9a47a59d.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/docs/bridge-protocol.md
+++ b/docs/bridge-protocol.md
@@ -414,6 +414,16 @@ Drop-receiver iframes have two ways to consume the payload:
 
 2. **Pull** — any iframe can postMessage `{ type: 'os-drag-payload-request' }` and the parent replies (directly to `event.source`) with `{ type: 'os-drag-payload', payload }`. Useful for iframes that bind their own native `drop` handler and need the rich payload after the browser has stripped the custom MIME from DataTransfer.
 
+### Drops the bridge declines
+
+While an iframe-sourced session is live, the intercept in `src/drag/iframe-drop-targets.ts` listens for `drop` on `document` in the **capture** phase — it has to, because the gesture is native HTML5 and has to be re-routed by hand into whichever iframe the cursor ended over.
+
+That reach stops at the iframe boundary. When there is no iframe window under the cursor — the drop landed on the wallpaper, a folder window's canvas, the dock — the intercept **declines**: it cancels the browser default and tears the session down, but leaves propagation alone so shell-side handlers get their turn. The files canvas uses exactly that opening to file an attachment dragged out of the Media Library as a desktop shortcut (see [`files-on-desktop.md`](files-on-desktop.md#drops-arriving-from-inside-a-window)).
+
+Cancelling the default is not optional even when the shell has nothing to do with the drop: a media drag carries `text/uri-list`, and the default action for that on a plain document is to navigate — a photo dropped on the desktop would otherwise replace the whole shell with the image.
+
+If you register your own drop handling on a shell surface, expect the event in the bubble phase and claim it with `preventDefault()` + `stopPropagation()`.
+
 ### Payload union
 
 ```ts
--- a/docs/files-on-desktop.md
+++ b/docs/files-on-desktop.md
@@ -503,6 +503,59 @@ window opts INTO accepting drops by registering a target on its own
 body — the Trash's `[data-os-recycle-bin-root]` is
 the canonical example.
 
+### Drops arriving from inside a window
+
+A drag lifted *inside* an iframe window — an image in the core Media
+Library — never becomes a `DragManager` session. The gesture belongs
+to the browser's native HTML5 drag machinery in the child document,
+and the shell sees no `pointerdown` and no `dragstart`. What the
+parent document does receive, the moment the pointer crosses out of
+the iframe, is an ordinary `dragover` / `drop` pair.
+
+The files canvas accepts those too. Drop an attachment on the
+wallpaper, on a folder window's canvas, or on a closed folder tile and
+it files as a shortcut exactly as a drag from WP Explorer would —
+same row-major packing, same optimistic store upsert, same
+`os.files.shortcut-dropped` action.
+
+The payload is read from whichever channel carries it:
+
+1. **`wp.os.dragBridge`** — authoritative. The source frame
+   postMessages `os-drag-start` before the pointer leaves it, so the
+   shell already holds a typed payload by the time the drop lands.
+   Survives browsers that strip custom MIME types across a frame
+   boundary. Bridge kinds `attachment`, `post` and `user` map onto the
+   file types of the same name; any other kind is refused rather than
+   guessed at.
+2. **`application/x-wp-media-attachment`** on the `DataTransfer` —
+   fallback, for a source that fills the transfer but never talks to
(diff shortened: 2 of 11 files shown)


AllTerrainDeveloper:
@claude can you review?

AllTerrainDeveloper:
Good catch — gated in `0967cc37`.

One clarification: that log predates this branch. It came in with #260 and has been firing on every `DRAG_EVENTS.START` since; it just happened to surface in review of a file this PR touches. Your read of it is right either way — `onDragStart` runs for **every** DragManager session, and repositioning a desktop icon is by far the most common drag in the shell, so ordinary use narrated itself and dumped the full payload object alongside.

I gated rather than deleted, because there's an in-tree precedent for exactly this shape: `src/recycle-bin/icon-state.ts` keeps its verbose trace behind `localStorage.openStationBinDebug` with the rationale "useful when this thing breaks again". The drag wiring has the same character — cross-frame, hard to reason about from a static read, and this PR is itself the second time it has needed explaining. So the trace now sits behind `localStorage.openStationDragDebug = '1'`, silent in the shipped shell.

Worth noting the module already exposes `window.__openStationIframeDropDebug()`, which reports the same facts on demand with no flag at all — that covers the snapshot case, and the flag covers the "I need to watch it happen" case.

Two tests pin it in `tests/vitest/iframe-drop-targets-bridge-drop.test.ts`: an ordinary drag says nothing, and the flag turns it back on.

PR #640, opened 2026-08-20
Navigation: one model decides where every menu, app and control lives

### LINKED ISSUE #586: Spatial layout ignores Apps & Icons placement choices
The **Apps & Icons** picker says a shortcut can live on the desktop, the dock, both, or nowhere. In Spatial layout, those choices do not reliably change the core WordPress shortcuts.

For example, choosing **Dock only** for Comments leaves the Comments icon on the wallpaper and does not add it to the dock. **Desktop and dock** produces the same wrong result. The preference saves, so the setting looks successful even though the desktop never matches it.

## How to reproduce

1. Switch OpenStation to **Spatial** layout.
2. Open **Preferences → Apps & Icons**.
3. Set Comments to **Dock only**.
4. Look at the wallpaper and dock.
5. Repeat with **Desktop and dock**.

## What should happen

Each placement should mean exactly what it says in every layout:

- Desktop only: wallpaper icon only.
- Dock only: dock icon only.
- Desktop and dock: both.
- Nowhere: neither.

## What happens

The generated core shortcuts in Spatial layout ignore the saved placement. This affects Dashboard, Posts, Media, Pages, Comments, Products, Appearance, Plugins, Users, Tools, and Settings.

## Where I would look

- `src/desktop-layout.ts`
- `src/desktop-icons.ts`
- `src/settings/sections/apps-icons.ts`

This is related to the Spatial routing problem recorded in #560, but it is a separate user-facing failure: this ticket is about whether the shortcuts appear in the places selected in Preferences.

Comments remains on the wallpaper after choosing a dock placement

mmtr:
Fixes #586

## Proposed changes

Every navigable thing (WordPress admin menus, plugin menus, apps, OpenStation's own controls) collapses into one flat list of nav items in `src/nav/`, and `computeNav()` decides where each one shows up. The dock, the sidebar and the desktop render what it returns.

- **Placement is a region, not a rail.** Stored as `rail | desktop | both | hidden`. `rail` resolves to the sidebar for a WordPress menu while the split layout is on, and to the dock for everything else, so switching layout is a re-render rather than a data migration.
- **The dock paints three zones**, with a divider between each pair of non-empty ones: WordPress menus, then apps (plugin menus, app launchers, and any running window with no home), then OpenStation's controls. Zone membership comes from what a thing is, so a tile cannot be dragged into another zone.
- **Every open window has a tile.** An app whose launcher lives on the wallpaper, or is hidden, gets one in the dock while its window is open and loses it when the window closes. Admin pages are excluded (their menu tile already lights up for child pages), and so is any window a system tile's submenu row opens.
- **Apps & Plugins becomes Preferences > Navigation.** Rows grouped by kind, one section each, and the rail option names the rail the item is actually on.
- **Right-click offers one write.** "Hide from dock" / "Hide from sidebar" / "Hide from desktop", or "Keep in dock" for a tile that is only there because its window is open. Exit OpenStation is the one item that cannot be moved or hidden.
- `itemVisibility` / `dockOrder` become `navPlacement` / `navOrder`. Existing arrangements carry over on read, so nobody loses the way they had things.

Before | After
--- | ---
Screenshot 2026-08-20 at 12 52 41 | Screenshot 2026-08-21 at 10 27 00

Each surface answered "where does this live?" for itself, from the registration path that produced it. An app registered twice got two answers.

Games shipped that way. It registers a native window with a dock tile **and** a desktop icon, which the two APIs explicitly allow, so Preferences read its default off the icon and said "On the desktop" while the dock read it off the window and painted a tile anyway. Picking any value in Preferences wrote an override both paths then agreed on, which is why changing the setting appeared to fix it.

The split layout could not be expressed either. `dock` was both a stored value and a rail name, so a WordPress menu on the sidebar was a menu whose stored placement said `dock`, and every consumer had to know that. #586 asks that a placement mean exactly what it says in every layout. The Spatial layout it reproduces on is gone, but that guarantee is now structural rather than something each surface has to remember.

### The Games bug

1. As an administrator, open **Preferences > Features > Extended options** and enable **Games**.
2. Reload the shell.
3. Make sure the Games icon appears on the desktop wallpaper and **not** on the dock. On `trunk` the dock paints a tile too.
4. Open **Preferences > Navigation**. Make sure Games is listed once, under **Plugins & Apps**, set to **On the desktop**.

### Zones and dividers

1. Look at the dock in the unified layout. Make sure it reads: WordPress menus, a divider, then OpenStation's controls (Mio, Overview, System, Trash, Exit).
2. Activate a plugin that registers a top-level admin menu (WooCommerce, Yoast). Make sure its menu appears after the first divider, in the same run as any app launchers, with no divider between them.
3. Make sure the leading divider does not appear when the run before it is empty.

### Running windows

1. Open Games from its desktop icon.
2. Make sure a Games tile appears in the dock's middle run with a running dot, while the wallpaper icon stays.
3. Close the window. Make sure the dock tile disappears.
4. Set an item to **Hidden** in Preferences > Navigation, then open its window from ⌘K. Make sure it still gets a tile while open.
5. Open **OpenStation Preferences** from the System tile. Make sure the dock does **not** grow a second gear beside it, and that the System tile shows the running dot instead.
6. Open a post for editing. Make sure no extra tile appears next to Posts.

### The split layout

1. **Preferences > Appearance > Desktop layout**, pick **Split**.
2. Make sure the sidebar holds WordPress's admin menus and nothing else, and the dock holds apps and controls.
3. With Preferences still open, make sure the sentence under the **Navigation** title now says "on the dock/sidebar", and every WordPress menu row now reads **In the sidebar**.
4. Right-click a sidebar tile. Make sure the menu says **Hide from sidebar**.
5. Switch back to Unified. Make sure the same menus return to the dock in the same order, with nothing reset.

### Right-click

1. Right-click a dock tile. Make sure you get **Hide from dock**, **Also show on desktop**, **Hide everywhere**, **Navigation settings…**.
2. Pick **Hide from dock** on a dock-only tile. Make sure it disappears rather than moving to the wallpaper.
3. Pick **Also show on desktop** on another, then right-click its dock tile again and **Hide from dock**. Make sure it stays on the wallpaper.
4. Open Games, then right-click its dock tile. Make sure the first entry is **Keep in dock**, and that picking it makes the tile stay after the window closes.
5. Right-click the Exit tile. Make sure no menu opens.
6. Right-click a wallpaper icon. Make sure you get **Hide from desktop**.

### Drag to reorder

1. Drag a WordPress menu tile to a new position in the dock. Make sure it lands there and survives a reload.
2. Drag a tile from the middle run onto the WordPress run. Make sure it snaps back and nothing is reordered.
3. Drag a control (Mio, Overview) within the trailing run. Make sure it reorders and Exit stays last.
4. In the split layout, drag a sidebar tile. Make sure the order holds, and that switching back to Unified keeps it.

### Preferences > Navigation

1. Make sure there are three sections, each in its own box: **WordPress Core**, **Plugins & Apps**, **OpenStation**.
2. Make sure WordPress Core lists menus in Core's order (Dashboard, Posts, Media, Pages, ...), not alphabetically.
3. Make sure Exit OpenStation is not listed.
4. Change a row to **On both**. Make sure the item appears on the dock and the wallpaper immediately, with no reload.
5. Change it to **Hidden**. Make sure it disappears from both.
6. Drag a promoted item around the wallpaper, set it to **Hidden**, then back to **On the desktop**. Make sure it returns to where you dragged it.

### Carrying over an existing arrangement

Existing users keep what they had. To check without an upgrade, write the old shape onto your user meta:

```bash
wp eval '
$m = get_user_meta( 1, "desktop_mode_os_settings", true );
unset( $m["navPlacement"], $m["navOrder"] );
$m["itemVisibility"] = array( "os-mio-toggle" => "hidden", "menu-posts" => "both" );
$m["dockOrder"] = array( "menu-settings", "menu-posts" );
update_user_meta( 1, "desktop_mode_os_settings", $m );
'
```

1. Reload the shell. Make sure Mio is gone from the dock, Posts is on both the dock and the wallpaper, and Settings leads the WordPress run.
2. Change any placement so the settings save. Make sure `navPlacement` and `navOrder` are now on the meta and the old keys are gone.

### Plugin surface

1. With a plugin that registers a dock rail renderer, make sure the rail still paints and updates on plugin activation. Renderers that do not implement `setZones()` are driven through `replaceItems` as before.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-640-43a5368fbab13d8c307fc2c8a55de5a257e0d6b3.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper (approved):
Okay, those changes are wonderful.

PR #641, opened 2026-08-20
perf: take the shell off the editor's typing path

AllTerrainDeveloper:
## Why

Editing a Gutenberg post inside an OpenStation window on production (WP.com) froze constantly. A read-only investigation of the prod server showed a healthy backend — clean PHP error log, lean options/cron tables, 200–300 ms TTFB — so the freezes are made in the browser, by work the plugin does exactly while the user types. This PR removes the three taxes that scale with typing and with open windows. No feature behavior changes.

### 1. Palette-gate the iframe command harvester (`src/commands/iframe-bridge.ts`)

The parent subscribed the focused window's command stream **permanently**. That keeps the chromeless bridge's React harvester mounted, re-rendering all of Gutenberg's command-loader hooks on every `wp.data` store tick — every keystroke — to feed a Cmd+K palette that isn't open. (The bridge's own teardown comment already called the always-on mount wasted CPU.)

Now the bridge subscribes when a palette opens and unsubscribes after it closes:

- New `os-palette-opened` / `os-palette-closed` CustomEvents, dispatched by the palette registry (Cmd+K cycle, `openPalette()`) and by the AI Assistant overlay's own entry points (Escape, ×, programmatic open/close). Consumers treat them as idempotent signals.
- The unsubscribe is sent on a **250 ms grace delay**: picking a command closes the palette *before* `run()` posts `os-commands-invoke`, and the harvester's callback cache dies with the unsubscribe — an immediate teardown would silently no-op loader commands like "Duplicate block".
- A plain close does **not** evict the harvested commands: the last list stays registered so reopening paints instantly while the fresh harvest streams in. Focus changes and window closes still evict stale owners exactly as before.
- Incidental fix: a minimized streaming window previously kept streaming (its lists were dropped on the floor); it now gets a real unsubscribe.

### 2. Coalesce the admin-bar offset neutralizer into idle time (`includes/render/chromeless-bridge.php`)

The body-wide MutationObserver called `getComputedStyle()` — a forced synchronous style recalc — on every added node **and every descendant**, inside the observer microtask (i.e. before the next paint). Gutenberg mounts/unmounts hundreds-of-node subtrees while typing (block toolbar, popovers, autocompleters), so this was textbook layout thrash on the hottest path.

The observer callback now only enqueues added **element** nodes (rich-text edits insert text nodes by the dozen — skipped at the gate) and schedules one `requestIdleCallback` flush (500 ms timeout backstop; `setTimeout` fallback). Nodes detached before the flush are skipped. The one-time full walk at DOMContentLoaded is unchanged, so nothing flashes on load; a late-mounted plugin header is corrected a frame or two later, imperceptible for an element that only just appeared.

### 3. Slow Core Heartbeat in background windows (new `src/window-activity-notifier.ts` + bridge handler)

Every iframe window is a complete wp-admin page whose editor heartbeat fires every 15 s even when the window is backgrounded — Core's backoff only reacts to the *tab* being hidden, which a background desktop window never is. Each of those admin-ajax ticks also boots the full plugin stack server-side.

The parent now posts `os-window-active` on window focus/blur (re-seeded on `os-bridge-ready`, so a background window that navigates doesn't come back fast). The chromeless bridge stretches `wp.heartbeat.interval()` to its 120 s maximum while inactive and restores the saved cadence on focus. Guard rails: never touches a sub-15 s cadence (something urgent like an auth-check retry), and only restores if the interval is still the 120 s it set — page code that re-tuned Heartbeat wins. Post locks stay safe: Core's lock window is 150 s.

## Deliberately not in this PR

- Pausing unfocus filters / backdrop blurs during typing — risks visible flicker regressions; needs design input.
- Background-window sleeping (unload + snapshot) — a feature, not a fix.

## Docs

- `docs/bridge-protocol.md` — new "Background heartbeat throttle" section.
- `docs/javascript-reference.md` — `os-palette-opened`/`os-palette-closed` events, updated `os-commands-subscribe`/`os-commands-unsubscribe` lifecycles, new `os-window-active` message, `registerPalette` guidance for plugin palettes.
- `docs/architecture.md` — harvester parent-side lifecycle rewritten.

## Verification

- `npm run test:js` — 4,722 tests green, including 9 new ones (`commands-iframe-bridge-palette-gating.test.ts` covering the gate, grace delay, reopen-within-grace, focus switching, and eviction; `window-activity-notifier.test.ts` covering focus/blur posts and the `os-bridge-ready` re-seed).
- `npm run test:php` — full suite, 2,297 tests green (4 pre-existing skips).
- `npm run lint`, `npm run typecheck`, `npm run lint:php`, `php -l`, `npm run build` — all clean.

Manual check on a slow machine: open a post in a window, type — with the palette closed there should be no React commits from the harvester host and no violet "Recalculate style" slices under a MutationObserver callback in a Performance recording. Background window heartbeats drop to one every 2 minutes (watch `admin-ajax.php?action=heartbeat` in the Network panel).

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-641-616ead821a68cb538488d45d2c355482facf2ddc.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
@claude can you review this? focus on regresion

PR #654, opened 2026-08-23
Components: the missing-import demo shows the code it talks about

AllTerrainDeveloper:
OS Settings → Components → **Missing-import warner — live demo** (Developer mode on).

The section's whole subject is a pair of bogus `<os-*>` tags, and it showed neither. The live copies have to be real elements for the warner to fire, and a real element is exactly what cannot be read — so the body box sat empty under a sentence pointing "below" at nothing.

## What changed

**The snippet is visible.** The demo markup is now also a string (`WARNER_DEMO_SNIPPET`), rendered in a `<pre><code>` beside the sentence. Two separate things on purpose: text to read, elements to trigger the warner. The live copies are unchanged — still clipped, still real.

**The sentence moved into the section body.** It has to sit *beside* the snippet, and `<os-section>` renders `description=` in its shadow tree, where nothing in the slot can share a row with it. So the section is now heading → box containing *sentence | snippet*, rather than heading → sentence → empty box.

**Layout.** One column, two equal halves at `@container (min-width: 1350px)`. The container is the OS Settings **window** (`.os-window--native:has( .os-settings )`, already `container-type: inline-size`), not the viewport — this panel is a window inside the shell, so window width is the only width that moves here. Neither track carries a fixed width; `minmax( 0, 1fr )` keeps a long code line scrolling inside the `<pre>` instead of pushing the grid past the section box.

**One code block, not three.** The `<code>` is now `display: block`. Left inline it inherits wp-admin's own `code` chip — a wash meant for a word mid-sentence — and an inline box wrapped over several lines paints that wash once per *line box*, so the blank lines in the snippet cut it into three chips.

**Third case dropped.** `<os-totally-made-up-thing>` hit the same unknown-tag branch as case 1 with a different spelling. The warner has exactly two answers — a name it can suggest a fix for, and one it cannot — so the demo has two tags. The console banner and the sentence say two.

## Testing

`npm run build`, `lint`, `typecheck` green; `test:js` 4806 passing. Manually: OS Settings → Features → Developer mode on, open Components, drag the window across 1350px.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-654-85265023ad4f5522613d6d1f55adc4d0300906a6.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/assets/css/os-settings.css
+++ b/assets/css/os-settings.css
@@ -1022,6 +1022,90 @@
 }
 
 /*
+ * Missing-import warner demo — the sentence and the snippet it talks
+ * about, in the section box.
+ *
+ * The snippet is shown because the section's whole subject is three
+ * tags and it used to show none of them: the live copies have to be
+ * real elements for the warner to fire, and a real element is exactly
+ * what cannot be read, so the box sat empty under a sentence pointing
+ * at nothing.
+ *
+ * One column by default; two equal halves once the window is wide
+ * enough for the prose to still have a paragraph's worth of room
+ * beside the code.
+ *
+ * Both tracks are `1fr` — the split is the box in half, and neither
+ * side carries a fixed width. A hard cap on the prose would leave a
+ * gutter of nothing between two columns that both had more to say.
+ * `minmax( 0, … )` on each so a long line of code shrinks the track to
+ * its share and scrolls inside the `<pre>` rather than pushing the
+ * grid wider than the section box.
+ *
+ * The container is the OS Settings WINDOW (`container-type` further
+ * down this file), not the viewport: this panel is a window inside the
+ * shell, so window width is the only width that moves here.
+ */
+.os-settings__help-warner {
+	display: grid;
+	gap: 16px;
+	align-items: start;
+}
+
+@container (min-width: 1350px) {
+	.os-settings__help-warner {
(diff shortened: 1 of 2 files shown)


AllTerrainDeveloper:
image

PR #655, opened 2026-08-23
Desktop icons: one reading order per canvas, and a bounded scan

AllTerrainDeveloper:
## The symptom

The wallpaper icons arrive **as a row across the top** on some page loads and as columns on others — same install, same icons, no setting changed.

## What was actually wrong

Three allocators disagreed about the desktop grid, and whichever ran last decided what the user saw.

**`FilesLayer.reflow()` is the visible half.** It rescues tiles whose stored coordinates fall outside the canvas — the layer is `position: absolute; inset: 0` and does not scroll, so a tile past an edge isn't below a fold, it's gone. But it repacked **row-major** whatever surface it was rescuing, so a single out-of-range tile flipped the entire wallpaper. It is deliberately not persisted, so it re-derived that verdict on every boot and every resize: the desktop flipped between rows and columns as the window height crossed the threshold.

**The server is why a tile was out of range at all.** `openstation_files_auto_place_orphans()` packed a column *999 cells deep* for a viewport it never asked about, so any desktop with more root items than fit in one visible column always had one stored below the bottom edge. It was also still on the **96 × 110** pitch the client left behind when the grid was unified at **108 × 120**, so its coordinates landed off-grid — and from the seventh row down two server rows aliased onto one client cell and had to be displaced on sight.

## The fix

- **An order is a property of the canvas, not of the operation.** `orderForFolder()` answers once — columns on the wallpaper, rows in a folder window — and every path that allocates a cell goes through it: drops, sorts, collision displacement on repaint, and the rescue. `nextFreeCell()` / `packCells()` replace the three hand-rolled allocator loops that each carried their own answer.
- **A scan is bounded by the canvas, and an unmeasurable canvas is not a one-cell canvas.** A host that isn't laid out yet reports `0`; the old `floor( ( 0 - 16 ) / 120 )` clamped up to `1`, and a column-major scan one row tall *is* a row. `GRID_FALLBACK_ROWS` / `GRID_FALLBACK_COLS` are what a caller with nothing to measure gets — the server always, the client before first layout. Small on purpose: wrapping a cell early costs a column, wrapping a cell late loses a tile.
- **`includes/desktop-files/grid.php`** is the PHP mirror of `src/desktop-files/grid.ts` — same pitch, same fallbacks, same order. `store.php` and `shares-store.php` had a copy of the maths each; both now consume it.
- **The rescue measures the tile box, not the cell box.** A cell carries the gap that separates it from the next one, and a tile in the last row has nothing after it — asking whether its gap fits condemned a fully-visible tile and repacked the canvas around it.

## Not changed, deliberately

A stored coordinate stays a **pixel** at render time. Resolving it through `pointToCell()` would land old-pitch rows on today's grid for free — and would also take away the off-grid manual positioning `buildOccupiedSet()` documents and the `tiles position from x/y` test pins. Coordinates written on the old pitch heal on the next drag, sort, or rescue; moving them wholesale needs a migration, not a render-time reinterpretation. Happy to write that migration as a follow-up if we want the tighter spacing gone immediately.

## Tests

- `tests/vitest/grid-order.test.ts` (new, 11 tests) — order per canvas, bounded scans, and the zero-height-host degeneracy that turned the column into a row.
- `tests/phpunit/tests/desktopFilesGrid.php` (new, 9 tests) — **parses the TypeScript**, so a number moving in one language and not the other fails a test that names the other one. Includes "no slot is ever invented past the assumed canvas", which is the server-side defect in one assertion.
- Full suites green: **4817 vitest**, **2381 PHPUnit**, `lint`, `lint:php`, `typecheck`, `build`.

## Docs

`docs/files-on-desktop.md` gains a **Reading order** section (the table, the two rules, and why the fallbacks are small), and the two places that described row-major packing as universal now point at it.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-655-527840a801b8180167c491d3a3d49d3d1ed7cb10.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/assets/css/os-settings.css
+++ b/assets/css/os-settings.css
@@ -1022,6 +1022,90 @@
 }
 
 /*
+ * Missing-import warner demo — the sentence and the snippet it talks
+ * about, in the section box.
+ *
+ * The snippet is shown because the section's whole subject is three
+ * tags and it used to show none of them: the live copies have to be
+ * real elements for the warner to fire, and a real element is exactly
+ * what cannot be read, so the box sat empty under a sentence pointing
+ * at nothing.
+ *
+ * One column by default; two equal halves once the window is wide
+ * enough for the prose to still have a paragraph's worth of room
+ * beside the code.
+ *
+ * Both tracks are `1fr` — the split is the box in half, and neither
+ * side carries a fixed width. A hard cap on the prose would leave a
+ * gutter of nothing between two columns that both had more to say.
+ * `minmax( 0, … )` on each so a long line of code shrinks the track to
+ * its share and scrolls inside the `<pre>` rather than pushing the
+ * grid wider than the section box.
+ *
+ * The container is the OS Settings WINDOW (`container-type` further
+ * down this file), not the viewport: this panel is a window inside the
+ * shell, so window width is the only width that moves here.
+ */
+.os-settings__help-warner {
+	display: grid;
+	gap: 16px;
+	align-items: start;
+}
+
+@container (min-width: 1350px) {
+	.os-settings__help-warner {
+		grid-template-columns: minmax( 0, 1fr ) minmax( 0, 1fr );
+	}
+}
+
+.os-settings__help-warner-text {
+	margin: 0;
+	font-size: 14px;
+	line-height: 1.55;
+	color: var( --os-ui-fg-muted, #646970 );
+}
+
+/*
+ * `overflow-x: auto` rather than wrapping: the first tag is 56
+ * characters of deliberately absurd component name, and breaking a tag
+ * across lines turns the one thing the reader is here to recognise
+ * into two things. The section box clips (`::part( body )` sets
+ * `overflow: hidden`), so the scroll has to live on the `<pre>`.
+ */
+.os-settings__help-warner-code {
+	margin: 0;
+	padding: 12px 14px;
+	overflow-x: auto;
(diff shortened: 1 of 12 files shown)


AllTerrainDeveloper:
@claude can you review?

PR #658, opened 2026-08-24
Fix: Plugin icons still missing

### LINKED ISSUE #611: Plugin icons still missing
Image

This has been around since a very early pre-release version. Is there any way to fix this issue of plugins missing their icons? About 50% of plugins I use are missing their icons. It's not that they don't have icons—the icon is there in the Add Plugins screen.

```
Plugin version:    1.1.0
WordPress version: unknown
Viewport:          2367x1387
Platform:          MacIntel
Current URL:       https://bricks.infinitemonkeys.ca/wp-admin/plugins.php?plugin_status=all&paged=1&s&desktop_mode_portal=1&desktop_mode_portal_intent=1
```

</details>

hbhalodia:
Fixes #611

### Problem

Some of installed plugins paint the placeholder instead of their icon, while the same icon shows fine on **Add Plugins**.

### Cause

The icon URL was built rather than looked up: `https://ps.w.org/<folder>/assets/icon.svg`, with the card walking `svg → png → gif` on error. That guess is wrong twice over.

- **Format** — plugins also upload JPEG, which was in neither the guess nor the chain. Gutenberg is example  as it only have JPEG, so every candidate 404s.
- **Slug** — the folder is not always the directory slug. `hello.php` is listed as `hello-dolly`.

Add Plugins is unaffected because it doesn't guess: it reads the `icons` map wp.org returns, which WordPress already caches locally in the `update_plugins` transient.

### Fix

That map is now consulted before guessing — **local file → `icons` (`svg` → `2x` → `1x`, core's own ladder) → guessed SVN URL** — and the guessed chain gains the JPEG variants for a cold transient. `icons['default']` (wp.org's geopattern for plugins with no art) is skipped, so those keep the window's placeholder.

`openstation_wporg_slug` was reading the same transient by hand; both now share one reader.

### Testing

On a site with Gutenberg or Hello Dolly active, open the Plugins window — both were iconless before, both resolve now.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-658-52be6567cc91a59aa44118719425746fe45b92f7.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


tests/phpunit/tests/pluginsWindowRegistration.php
```diff
@@ -633,6 +633,103 @@ public function test_icon_url_single_file_uses_textdomain_fallback() {
...
+	 *
+	 * @param string $plugin_file Plugin file the entry describes.
+	 * @param array  $icons       The entry's `icons` map.
+	 * @param string $slug        Directory slug; defaults to the folder.
+	 */
+	private function prime_update_transient( $plugin_file, $icons, $slug = '' ) {
+		set_site_transient(
+			'update_plugins',
+			(object) array(
+				'response'  => array(),
```
mmtr: This helper needs `'last_checked' => time()`, like every other seeded transient in this file.

Without it, `maybe_refresh_update_transient()` doesn't hit the 12h short-circuit, so `wp_update_plugins()` runs and the real wp.org response overwrites the seeded entry (`gutenberg/gutenberg.php` isn't actually installed). `field_icon_url()` never primed before this PR, which is why this is only biting now.

The three new tests pass in a full run only because `static $primed` was already set by `test_update_available_field_reports_false_when_no_update` earlier in the class. In isolation:

```
npm run test:php -- --filter='test_icon_url_prefers_directory_icons_over_guess'
```

```
-'https://ps.w.org/gutenberg/assets/icon-256x256.jpg?rev=1776042'
+'https://ps.w.org/gutenberg/assets/icon.svg'
```

`test_icon_url_survives_a_slug_that_is_not_the_folder` fails the same way. `test_icon_url_ignores_the_wporg_default_geopattern` passes for the wrong reason: the transient gets wiped, so the guessed URL it asserts is what you'd get with no entry at all. It isn't proving `default` is skipped.

Adding `'last_checked' => time()` makes all three pass in isolation. Verified locally.

includes/plugins-window/rest-fields.php
```diff
@@ -548,6 +574,37 @@ function openstation_plugins_window_field_icon_url( $row ) {
...
+		if ( empty( $icons[ $size ] ) || ! is_string( $icons[ $size ] ) ) {
+			continue;
+		}
+		$url = esc_url_raw( $icons[ $size ] );
+		if ( '' !== $url ) {
+			return $url;
+		}
+	}
+
+	return null;
```
mmtr, line 611: When wp.org gave us an `icons` map with only `default`, we know for a fact the plugin uploaded no art. Falling through to the guess means the card then walks all seven candidates (404 each) before painting the dashicon.

Returning `null` here when `$entry['icons']` is a non-empty array with no `svg`/`2x`/`1x` gets the placeholder straight away. That also matches what the hooks-reference line already promises ("`default` is skipped, so an iconless plugin still gets the window's placeholder").

includes/plugins-window/rest-fields.php
```diff
@@ -548,6 +574,37 @@ function openstation_plugins_window_field_icon_url( $row ) {
 	);
 }
 
+/**
+ * Pick a card icon out of the `icons` map wp.org returned.
+ *
+ * `svg` → `2x` → `1x`, the ladder core's Add Plugins cards use (see
+ * `WP_Plugin_Install_List_Table::display_rows()`); matching it is what
+ * makes the two screens agree. `default` — wp.org's geopattern for
```
mmtr, line 588: Small thing: this matches core's Add Plugins ladder, but not the Add Plugins tab in this same window. `pickIcon()` in `src/plugins-window/card.ts` ranks `icons.default` above `2x`/`1x` and never skips it, so the same plugin can show a geopattern under Add Plugins and a dashicon under Installed.

Either align the two or say here that they diverge on purpose.

docs/hooks-reference.md
```diff
@@ -3018,7 +3018,9 @@ apply_filters( 'openstation_plugins_window_icon_url', string|null $url, string $
...
+2. **The directory's own `icons` map** — whatever wp.org returned for this plugin, cached in the `update_plugins` transient, read `svg` → `2x` → `1x` exactly as core's "Add Plugins" cards read it. Used verbatim, `?rev=` cache-buster included. `default` (wp.org's generated geopattern for plugins that uploaded no art) is skipped, so an iconless plugin still gets the window's placeholder.
+3. **Guessed wp.org SVN asset** — `https://ps.w.org/<slug>/assets/icon.svg`, for when that metadata isn't cached. `<slug>` is the directory slug when known, else the folder name, else the textdomain. Guessing is last because both halves are unknowable: the format (Gutenberg and UpdraftPlus ship JPEG only) and the slug (`hello.php` is listed as `hello-dolly`). The JS card walks a candidate chain (SVG → 256 PNG/JPEG/GIF → 128 PNG/JPEG/GIF) on `<img>` error for URLs of exactly this shape, then drops to the placeholder. Local, `icons`-map and custom URLs are one-shot, then placeholder.
+
```
mmtr: Nit: extra blank line here.

src/plugins-window/icon-fallback.ts
```diff
@@ -34,15 +34,16 @@ function buildCandidates( initialUrl: string ): string[] {
...
-	// over GIF within each size — PNG is faster to decode + render
-	// for a 32-px card cell, GIF only matters when that's the only
-	// format the plugin shipped (Elementor, a handful of others).
+	// 256-px variants before the 128 set, and within each size the
+	// likeliest format first — PNG also being the fastest to decode
+	// for a 32-px cell.
 	return [
 		initialUrl,
 		base + 'icon-256x256.png',
+		base + 'icon-256x256.jpg',
```
mmtr, line 43: wp.org accepts `icon-*.jpeg` too, not just `.jpg`. Low stakes now that the map is the primary path, but the "the format is unknowable" reasoning applies here as well.

mmtr (changes requested):
Thanks for digging into this one, @hbhalodia. The diagnosis is spot on: guessing a URL when wp.org already handed us the `icons` map was the bug, and reading both `response` and `no_update` through one shared helper is the right shape. Precedence is correct, the map URL goes through `esc_url_raw`, and keeping `WP_ORG_ASSET_RE` anchored on `icon\.svg$` so the `?rev=` URLs stay one-shot is a nice detail.

I ran the suite on the branch: `test:php` (55/55 in that class), `test:js`, `typecheck`, `lint` and `lint:php` are all green.

One blocking thing on the tests, plus a few smaller notes inline. 🙇

tests/phpunit/tests/pluginsWindowRegistration.php
hbhalodia: Thanks @mmtr for the detailed review. I have checked and this fails when only a single test is run via `--filter`. I would update the transient the way other tests are using in the files.

This is resolved in - https://github.com/WordPress/openstation/pull/658/commits/69aa0257fb00fcb96096a2323b8dc6a438d4d2c5

includes/plugins-window/rest-fields.php
hbhalodia: Thanks, I have updated to check if we have non empty icon array, do not make the default as null and skip that, so it won't waste resources in guessing the URLs.

Fixed in - https://github.com/WordPress/openstation/pull/658/commits/64549b07797d254851b54e26035bdb0d1123ebda

includes/plugins-window/rest-fields.php
hbhalodia: I guess, we should align this 2 instead of saying that they diverge to make it consistent accross the screens. I would update PR to make sure default is added after `1x` in order.

Fixed in - https://github.com/WordPress/openstation/pull/658/commits/6e6d8069372fbc7ef0ace3a45508dfac1f377dce

docs/hooks-reference.md
hbhalodia: Thanks, this is removed.

Fixed in - https://github.com/WordPress/openstation/pull/658/commits/da830ff0e513719c409747d90b518cec13c85a38

src/plugins-window/icon-fallback.ts
hbhalodia: Thanks, have added the `jpeg` to the list as well.

Fixed in - https://github.com/WordPress/openstation/pull/658/commits/f1d1a9de03d906d7112a716c7b88d31926a3ff41

hbhalodia:
Thanks @mmtr For the review, The feedbacks are now addressed. Can you please re-review?

mmtr (approved):
All good now, thank you!

PR #656, opened 2026-08-23
Trash: localize config before the lazy payload harvest

AllTerrainDeveloper:
Since bundles went lazy (#613), openstation_enqueue_assets() harvests every native window's wp_localize_script data into the shell payload at admin_enqueue_scripts priority 10. The recycle bin and the games window still attached their config at priority 30 — fine for as long as their bundles were enqueued eagerly and WordPress printed the data at print time, but on the lazy path the payload had already shipped without it. The visible failure: the Trash window opens, lists nothing, and logs "openStationRecycleBinConfig is missing".

Both localizers move to priority 5, alongside the native-window data attach, and a new test pins the contract for every module that localizes config onto a lazy window handle: attach before the harvest, or your bundle boots blind.

Claude-Session: https://claude.ai/code/session_0119U4sRRWGcTQdwYreTwpAp

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-656-6088a33ed0485f3fb7239ff1a1518cf0210214fe.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
  Open WordPress Playground Preview
</a>

PR #659, opened 2026-08-24
Menu-refresh probe: attach script data before the payload harvest

AllTerrainDeveloper:
## What broke

Since #613, WP Explorer's WooCommerce integration rides the Explorer window as a lazy **companion script**, delivered through the native-windows payload together with its `openStationWooConfig` inline blob (attached to the handle at `admin_enqueue_scripts` ≤ 5, harvested into the payload at 10 — the contract #656 pinned).

That chain holds for the boot payload and the admin-footer bridge — but not for the **menu-refresh probe**. `wp.os.refreshMenu()` (auto-fired on plugin activation/deactivation and on any `os-menu-signature` drift) spawns a hidden iframe whose handler short-circuits `admin.php` on **`admin_init` @ 99 — before Core ever fires `admin_enqueue_scripts`**. Nothing was ever attached, so the probe's payload shipped every window's companion with empty `scriptBefore` / `scriptL10n` arrays. The shell refreshes its native-window index from every payload it receives, so one menu refresh silently downgraded entries the boot payload had delivered complete.

The visible casualty: after any menu refresh, the first lazy open of WP Explorer loaded the Woo companion with no config — **order bands gone, merchant preview panels dark**, nothing in the console to say why. The recycle-bin and games configs were stripped from probe payloads the same way. Pre-#613 none of this mattered because the bundles and their config were printed eagerly at boot.

## The fix

- **`openstation_menu_refresh_probe_payload()`** (new, `chromeless-bridge.php`): the probe now replays `admin_enqueue_scripts` before building its payload — inside a discarded output buffer (an echoing callback must not beat the short-circuit's `header()` calls), with the current screen set first, because `admin.php` only calls `set_current_screen()` *after* `admin_init` and Core's own block-editor enqueue callbacks read `get_current_screen()->id` unguarded. Guarded by `did_action()`, so any request where Core fired the hook is a pure pass-through. Fixes Woo, recycle bin, games, and any third-party window following the documented attach-at-≤5 contract in one move — no new hook, no parallel contract.
- **Woo assets cache-bust by `filemtime`** like the Explorer assets they ride with. They're fetched lazily by URL, so a `ver` that only moves on release let a browser's cached copy outlive builds within one — stale-companion contract drift with no error pointing at it.

## Scope / risk

The replay lives only in the probe request — a throwaway hidden iframe that exits before any print pipeline runs. Enqueue callbacks it fires are the same ones every real admin page already runs; anything they echo is buffered and discarded. Worst case (a callback fataling in the unusual context) breaks only the probe, degrading to `refreshMenu()`'s pre-existing timeout path.

## Verification

- Regression test pins that the probe payload carries handle-attached inline data (`Tests_OpenStation_NativeWindowLazyScript`).
- Full PHPUnit suite green (2384 tests), `phpcs -n` clean, `npm run build` clean.
- Live before/after on a real store: the unpatched probe response contains neither `openStationWooConfig` nor `openStationRecycleBinConfig`; the patched one carries both, and the Explorer's Woo bands/previews survive a menu refresh.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-659-a77b9f538cc97ce5d8a4dbe264f58dd668333183.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/my-wordpress/integrations/woocommerce.php
+++ b/includes/my-wordpress/integrations/woocommerce.php
@@ -1900,23 +1900,32 @@ add_action( 'rest_api_init', 'openstation_my_wordpress_woo_register_routes' );
 /**
  * Register the integration bundle.
  *
+ * Cache-busted by `filemtime`, like the window bundle it rides along
+ * with (see `openstation_my_wordpress_register_assets()`). The bundle
+ * is fetched lazily by URL, so a `ver` that only moves on release
+ * would let a browser's cached copy outlive builds within one — a
+ * stale companion against a fresh WP Explorer bundle is a contract
+ * drift no error message points at.
+ *
  * @return void
  */
 function openstation_my_wordpress_woo_register_assets() {
+	$js_path = OPENSTATION_DIR . 'assets/js/my-wordpress-woocommerce' . openstation_asset_suffix() . '.js';
 	wp_register_script(
 		'os-my-wordpress-woocommerce',
-		OPENSTATION_URL . 'assets/js/my-wordpress-woocommerce' . ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG ? '' : '.min' ) . '.js',
+		OPENSTATION_URL . 'assets/js/my-wordpress-woocommerce' . openstation_asset_suffix() . '.js',
 		array( 'wp-hooks' ),
-		OPENSTATION_VERSION,
+		file_exists( $js_path ) ? (string) filemtime( $js_path ) : OPENSTATION_VERSION,
 		true
 	);
 	wp_set_script_translations( 'os-my-wordpress-woocommerce', 'desktop-mode' );
 
+	$css_path = OPENSTATION_DIR . 'assets/css/my-wordpress-woocommerce.css';
 	wp_register_style(
 		'os-my-wordpress-woocommerce',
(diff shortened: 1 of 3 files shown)


AllTerrainDeveloper:
@claude can you review?

PR #660, opened 2026-08-24
Windows: companion stylesheets, and the Woo CSS stops riding every page

AllTerrainDeveloper:
## What

`openstation_register_window()` gains a `'styles' => [ $handle, … ]` arg — the styles-side mirror of the `scripts` companion list — and the WooCommerce Explorer integration becomes its first user, dropping the one eager enqueue it had left.

## Why

A window takes one `style` handle, injected when the window registers. With no companion slot on the styles side, the Woo integration's stylesheet had to be enqueued eagerly — which stamped it into **every admin document on a store**: the shell boot page *and* every chromeless iframe, where every selector in it is scoped to surfaces (Explorer panels, tile ribbons, Customer window chrome) that only ever render in the parent shell. Pure parse weight, per document, on every page a merchant opens in a window.

## KB improvements (WooCommerce store)

`my-wordpress-woocommerce.css` = 12.8 KB raw / 4.3 KB gzipped · `my-wordpress-woocommerce.min.js` = 47.4 KB raw / 13.6 KB gzipped

| Document | Before | After | Saved |
|---|---|---|---|
| Shell boot page | CSS linked + parsed (12.8 KB raw / 4.3 KB gz) | nothing until first open | **12.8 KB raw / 4.3 KB gz** |
| Every chromeless iframe (each page opened in a window, each in-window navigation) | CSS linked + parsed per document | nothing, ever | **12.8 KB raw / 4.3 KB gz per document** |
| First open of WP Explorer / Customer window | JS only (47.4 KB raw / 13.6 KB gz — deferred since #613) | JS + CSS (60.2 KB raw / 17.9 KB gz), fetched in parallel, once per session for both windows | — (cost moved here, not added) |
| Admin session that never opens either window | CSS on every document | config blob in the boot payload only (~1–2 KB) | everything above |
| Site without WooCommerce | 0 | 0 | — |

The steady-state per-document cost of the whole integration on a store is now the `openStationWooConfig` blob riding the boot payload.

## How

- **Registry** — `styles` normalized like `scripts` (deduped, empties stripped). With `preload_script`, companion styles are enqueued at boot alongside the scripts, keeping "preload means everything".
- **Payload** — each handle resolves through `openstation_resolve_style_payload()` into `companionStyles[]` (`styleUrl` / `styleHandle` / `styleInline`); unregistered handles drop silently, same as script companions.
- **Shell** — companions inject at the top of the first-open script chain (the `<link>` fetches in parallel with the bundles), once per URL, skipping links the server printed. Deliberately different timing from `style`, which still lands at registration so mid-session activations paint.
- **Cascade** — the subtle bit `tileStylesheetOrder` exists for: the Woo sheet overrides `my-wordpress.css` at *equal specificity*, an order the enqueue path guaranteed via a `wp_register_style` dependency. Append-to-head injection preserves it — the Explorer's own sheet is already in `<head>` when the companion lands, so the companion wins by source order. Pinned by a vitest case.
- **Woo integration** — eager enqueue deleted; the sheet rides WP Explorer via the `window_args` filter and the Customer window via `styles` (moved off `style`, which would have re-injected it at every shell boot through the sync path). Whichever window opens first pays it once for both.

## Tests & docs

- 5 new vitest cases: sync doesn't inject; first open injects after the window's own sheet; shared-URL dedupe across windows; preload path; a server-printed link is detected, not duplicated.
- 3 new PHPUnit cases: payload default, resolution with inline data + silent drops, preload enqueue.
- Full suites green: 4847 vitest / 2387 PHPUnit, typecheck, eslint, phpcs.
- Docs updated in the same change: `hooks-reference.md` (registration args), `architecture.md` (delivery model), `migration-lazy-window-scripts.md` (companion recipe), `javascript-reference.md` (`loadWindowScript` note).

Verified live against a real store: iframe documents and the shell page carry zero Woo CSS links; both window entries in the boot payload carry `companionStyles`; bands, ribbons, and panel chrome render identically, with the CSS/JS arriving in the Network tab at first open.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-660-e977e14c23e544b681f660ac7722df3d7e79876f.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
@claude can you review? is this harmless?

PR #664, opened 2026-08-24
Perf: defer the command palette's Gutenberg runtime (3.6 MB raw / 1.1 MB gz) to first ⌘K

AllTerrainDeveloper:
## What

The ⌘K palette's WordPress baseline commands live in the `core/commands` store, which needs `wp-commands` + `wp-core-commands` — and their dependency closure is the **entire Gutenberg runtime: 52 bundles, 3.59 MB raw / 1.10 MB gzipped** (production builds; `block-editor` 427 KB gz and `components` 271 KB gz alone). Until now every shell boot paid all of it — on WP 7.0 via Core's *own* `admin_enqueue_scripts` default, on 6.9 via our eager call — for a palette most sessions never open, powering a Core UI the shell suppresses unconditionally anyway.

This PR turns the chain into an ordered manifest replayed on the **first palette invocation**.

### Opening OpenStation (shell boot, vanilla WP site)

| | trunk | this PR | saved |
|---|---|---|---|
| Command-palette runtime at boot | 52 bundles — **3.59 MB raw / 1.10 MB gz** — plus 3 stylesheets, fetched, parsed, and React-booted on every admin boot | a JSON manifest inside the config blob (~15 KB) | **~3.6 MB raw / ~1.1 MB gz + 52 requests** per boot |
| Transfer time | | | ~5.5 s on Fast-3G (200 KB/s) · ~0.7 s on 4G (1.5 MB/s) |
| Parse/execute | React + Gutenberg runtime evaluated at boot | nothing until first use | every boot, every device |

### First ⌘K of a session (the cost's new home)

| | trunk | this PR |
|---|---|---|
| Palette opens | instantly, baseline pre-loaded | **instantly** — shell commands, slash commands and AI available immediately; the WP baseline (Add new post, Manage plugins, …) streams in when the chain lands (`os-command-palette-ready` → harvester mounts) |
| Chain load | already paid at boot | preload-warmed in parallel, executed in dependency order; second ⌘K onward: already resident |

### Honest caveats

- **Plugin-heavy sites keep third-party weight**: on the dev store, `wp-abilities` and WooCommerce's `wc-entities` independently pull `block-editor`/`components` onto the dashboard. The client's same-path DOM sniff detects those and skips re-executing them (re-running `wp-data` would wipe every registered store) — the win there is bounded to what only we pulled. A vanilla site gets the full 1.1 MB gz.
- **Chromeless iframes and classic-mode pages are untouched**: iframes keep Core's enqueue because the iframe-side command harvest depends on the store; classic pages are Core's own UI where Core's palette is the right one.
- Pre-6.9 sites: manifest is `null`, everything no-ops, behavior unchanged.

## How

- **`openstation_defer_core_command_palette()`** — unhooks WP 7.0's boot-time default on shell pages only (priority 0, ahead of Core's 10). Harmless no-op on 6.9.
- **`openstation_build_command_palette_assets_payload()`** — lets Core's `wp_enqueue_command_palette_assets()` run exactly as it would (menu-command serialization + the `initializeCommandPalette` inline included), then **unwinds** it: queue snapshot → diff → restore, with the ordered dependency chain resolved on clones so the request's live `$to_do` is never touched. Each handle is harvested into the same url/before/after/l10n/translations shape the native-window lazy loader uses; src-less aggregators are kept for their inline data.
- **`src/commands/palette-assets.ts`** — wired into the assistant stub's `_ensure()` (the single ⌘K funnel): `<link rel="preload">` warms every missing URL in parallel, then handles execute strictly in dependency order via `loadVendorScript` with inline data replayed around each. Single-flight with retry-after-failure. Fires `os-command-palette-ready`; the shell harvester (idle `install()` now a graceful no-op) re-installs on it.
- `window.__openStationMenuCommands` stays a boot-time inline — built in PHP, a few KB, and the harvester's classification needs it before `desktop.min.js` boots.

## Tests & verification

- `commandPaletteAssets.php` (5): the unwind leaves boot queues and `$to_do` byte-identical; the manifest is dependency-ordered and carries both roots; `initializeCommandPalette` rides the `wp-core-commands` entry; the style chain rides; the unhook is shell-only (chromeless keeps Core's default).
- `command-palette-assets.test.ts` (6): ordered replay, already-on-page skip, src-less inline slots, single flight, retry after failure, no-manifest no-op.
- Suites: 2401 PHPUnit / 4861 vitest green, typecheck, eslint, `phpcs -n`, build.
- Live-verified: the boot page's manifest carries all 49 script handles in correct order ending `…wp-core-data → wp-router → wp-core-commands`, with the init inline; `commands`/`core-commands`/`router` gone from the boot script list; manifest URLs serve 200. Manual pass on :8889 confirmed ⌘K opens instantly and the WP baseline appears and executes.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-664-81b2bf755ff3bd1c9838a357073e671c13c08b29.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


epeicher (approved):
Thanks @AllTerrainDeveloper for improving this! I have tested it and reviewed it, and I have observed similar numbers for the improvement when the Gutenberg plugin is not active.

Changes LGTM!

epeicher:
These are the numbers I have measured 
CleanShot 2026-08-24 at 17 27 29@2x

PR #661, opened 2026-08-24
Perf: defer every window stylesheet to first open; fix shared-bundle configs; real theming tokens in Users/Profile

AllTerrainDeveloper:
## What

Three fixes travelling together because your test pass surfaced each while verifying the last:

1. **Every built-in window's stylesheet is deferred to its first open** — completing for CSS what #613 did for scripts, on the `styles` companion mechanism #660 shipped. Twelve windows migrate (`'style' => X` → `'styles' => [ X ]`, eager enqueues deleted); the per-game sheets ride the Games window via `openstation_games_window_args`; and the three on-demand surfaces that are *not* native windows (Preferences panel, AI assistant, bug report) get the same deferral through a new `openStationConfig.deferredStyles` map + `ensureDeferredStyle()` at each open path. Deliberate exception, pinned by a test: **recycle-bin.css stays eager** — it styles the dock tile's drag-over drop target, a boot-time surface.
2. **Shared bundles deliver their siblings' configs.** Posts / Pages / Users / Profile share `os-posts-window.js`; the shell fetches a URL once, so a config that only travelled with its own entry was dropped for every sibling after the first (`[desktop-mode-pages] config blob is missing`), and the Users window's embedded Profile form — which reads the *user-edit* config — could never see it under any open order. Configs now group **by script handle** server-side (own window first), and the client replays each entry's inline data on that entry's own first open even when the script tag was deduped.
3. **Users / Profile theming.** Seventeen `var()` reads named phantom tokens (`--wp-admin-theme-bg`, `-bg-elevated`, `-border`, `-fg-muted` — WordPress defines none of them), so the identity card and colour-scheme tiles sat on light fallback literals with theme-following text on top: light-on-light on every dark desktop theme. All moved to the palette's `--os-ui-*` tokens with pre-brand literals as fallbacks, and a new sweep test bans the phantom family for good.

## KB & time — measured against v1.1.2 (the wp.org release)

CSS sizes from the v1.1.2 tree; current-state lists verified live against a running store.

### Scenario 1 — opening OpenStation (shell boot)

| | v1.1.2 | this PR | saved |
|---|---|---|---|
[...]

PR #668, opened 2026-08-24
Perf: native-window script data ships once per handle, not once per window (boot document −8.9%)

AllTerrainDeveloper:
## What

The boot payload's fattest key was repetition. `nativeWindows[]` inlined each entry's *resolved* script data — localize blobs, inline snippets, translations, and the synthesized `openStationWindowConfig` set — but script data is a property of the **bundle**, not the window. Posts, Pages, Users and Profile all ride `os-posts-window`, so the same blobs serialized four times over; the shared-config grouping from #661 (correctly) made that four copies of the whole four-window config set. `scriptL10n` alone was ~100 KB of payload, most of it duplicates.

The payload now ships two halves:

- **Slim entries** that reference bundles by handle — `scriptHandle`, `companionScripts` as an ordered handle list, `tabs[].scriptHandle`.
- **`nativeWindowScriptData`** — a handle-keyed map holding each bundle's resolved data exactly once (`url` / `before` / `after` / `l10n` with the handle's whole config set appended once / `translations`).

`openstation_collect_native_windows_payload()` builds both from one walk; the shell joins them on receipt via `hydrateServerEntries()` (`src/native-windows.ts`), wired at both consumption points — boot sync and the live menu-refresh `applyPayload`. The hydrator deliberately tolerates the **old inline format** (entries carrying resolved fields, companion objects instead of handle strings), so a bridge payload crossing versions mid-deploy degrades to previous behavior instead of breaking.

## Comparison — measured with `bin/boot-cost.mjs` (#667), same instance, same configuration

| | before (trunk) | after (this PR) | delta |
|---|---|---|---|
| Boot document | 676.3 KB | 616.0 KB | **−60.3 KB · −8.9%** |
| …of which inline JS | 492.1 KB | 431.8 KB | **−12.3%** |
| Document gzipped (wire) | 126.4 KB | 124.0 KB | −2.4 KB · −1.9% |
| Asset requests | 190 | 190 | 0 |
| Asset bytes (raw / gz) | 16.00 MB / 3.29 MB | 16.00 MB / 3.29 MB | ±0.0% |

The honest read of that table: gzip was already soaking up the repetition on the wire, so the real win is **60 KB less JSON parsed on the main thread at every boot** — and the identical shrink in every payload rebuild the shell performs live (the chromeless bridge emission on plugin changes, and the menu-refresh probe, both of which serialize this structure per request). Asset rows are byte-identical by design: this is a structural change to the payload, not to what loads.

Payload-structure verification against a live store: 14 handles in the map, zero `scriptL10n` keys left inside `nativeWindows[]`, companions shipped as handle strings, and the shared-handle `l10n` carries each sibling config exactly once.

## Correctness notes

- The #661 shared-bundle guarantees are *strengthened*, not re-implemented: whichever entry loads a bundle now hands the loader the handle's full data set by construction, so siblings and cross-window consumers (the Users window's embedded Profile form reading the user-edit config) see their config on any open order.
- Config assignments for handles that resolve to nothing drop, exactly as before. The dead `elseif` that synthesized config onto a payload field no path ever delivered (declarative windows) is gone with the field it wrote to.
- `wp.os.debug.window()` resolves its reported URL through the map, with the old inline field as fallback.

## Tests & gates

- PHP: the lazy-script suite reads harvested data from the map and pins the dedupe (each shared config **exactly once**); the menu-refresh-probe test reads `nativeWindowScriptData`.
- JS: three new hydrator cases — the join across entry/companions/tabs, old-format passthrough, and a shared handle delivering every sibling config on one load; the menu-refresh-apply forwarding test follows the hydrated shape.
- Suites: 2401 PHPUnit / 4869 vitest green, typecheck, eslint, `phpcs -n`, build.
- Docs: `architecture.md` payload-harvest section rewritten around the map; `AGENTS.md` payload key list updated.
- Manual regression pass on :8889 covered the historical breaking orders: Posts→Pages, Users→Profile tab, Edit-user window, WP Explorer→Woo→Customer window, tabbed windows, and a plugin activate/deactivate live-refresh cycle.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-668-e27e4300f2621db2929c09dd006b6b8ee30da939.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
@claude can you review?

PR #669, opened 2026-08-24
Perf: the boot config ships references, not copies (boot document −12.8%)

AllTerrainDeveloper:
## What

Part 2 of the boot-payload diet (part 1 was #668). Three cuts, each removing data the boot page already delivers another way:

1. **Native-window templates** — every registered window's template is server-printed as a real `<template>` tag (`admin_footer` @ 20, before footer scripts; `ensureTemplate()` adopts it by id), so the payload's `templateHtml` copy was ~27 KB of the same markup twice. The boot config now ships it empty. **Bridge and probe payloads keep theirs** — they're the mid-session activation path, where no template was printed.
2. **The menu-command list ships once** — Core's captured `initializeCommandPalette` inline embedded ~20 KB of `menu_commands` that the boot page already carries as `window.__openStationMenuCommands` (the shell harvester's lookup, and the richer copy: its URL derivation routes legacy file-path slugs through `menu_page_url()` where Core's regex takes them literally). The palette manifest strips the embedded copy and synthesizes the same call against the global — which prints at boot, while the manifest replay waits for the first ⌘K.
3. **The theme library slims at boot** — `serverDesktopThemes` shipped each theme's compiled `cssText` **and** its token map (~40 KB for Legacy alone) when nothing reads either at boot: the *active* theme's stylesheet is server-enqueued (and `bootAlreadyApplied()` detects the stamp), and an inactive theme's CSS matters only at the moment it's picked. Boot entries now carry `cssDeferred: true`; that moment fetches full entries from the new **`GET desktop-mode/v1/desktop-themes`** — same builder, same `openstation_desktop_themes` filter, so deliberately *not* a second source of truth — via `ensureFullDesktopThemes()`, wired into the apply path: the chrome flips instantly, the stylesheet lands when the fetch does, and a stale pick discards the late arrival. Read gate is the shell's own (the payload shipped identical data to every desktop user); bridge/probe payloads stay full.

## Measured — `bin/boot-cost.mjs`, same instance, same configuration

| | before (trunk) | after (this PR) | delta |
|---|---|---|---|
| Boot document | 616.0 KB | 537.1 KB | **−78.8 KB · −12.8%** |
| Document gzipped (wire) | 124.0 KB | 112.7 KB | **−11.3 KB · −9.1%** |
| …of which inline JS | 431.8 KB | 335.0 KB | **−96.8 KB · −22.4%** |
| Asset requests | 190 | 190 | 0 |
| Asset bytes (raw / gz) | 16.00 MB / 3.29 MB | 16.01 MB / 3.29 MB | ±0.01% |

Cumulative across the diet series (#668 + this): the boot document is down **692 KB → 537 KB (−22.4%)**, with no behavioral surface removed — every byte still arrives, once, on the path that actually needs it.

Live verification on a running store: `templateHtml` empty in the boot blob while the `<template>` tags and the bridge payload keep the markup; `"menu_commands":[` absent from the manifest with the synthesized init present; the Legacy entry slim at boot (`cssDeferred: true`, tokens gone) while `GET /desktop-themes` serves the full entry (18.2 KB `cssText`, 465 tokens).

## Correctness notes

- Theme boot is untouched: the active theme's CSS arrives exactly as before (server-enqueued), and `bootAlreadyApplied()` short-circuits the client path — the deferral only engages on a mid-session pick.
- The apply path re-checks the active id when the fetch lands, so picking something else mid-flight never injects a stale stylesheet (pinned by test).
- One doctrine reversal, done on purpose: `desktopThemesRest.php` pinned "there is deliberately no GET route — the library rides the payload". The library still rides the payload for *metadata*; the route exists precisely so the payload doesn't have to carry compiled CSS for themes nobody is wearing. The test now pins the route serving the library, and the route header documents why it isn't a second source of truth.
- `DesktopThemeEntry` gains `cssDeferred` (documented); `cssText`/`tokens` are unchanged in shape and still full on bridge payloads, upload responses, and the GET route.

## Tests & docs

- New `bootPayloadDiet.php` (4): boot blob slim for templates and themes, printer + bridge payload keep full copies, the GET route serves full entries and refuses users without the shell.
- `commandPaletteAssets.php` (+1): the menu list appears once — synthesized init reads the global, embedded copy stripped.
- `desktop-themes.test.ts` (+3): `cssDeferred` normalization, deferred-apply flow (instant chrome, CSS on fetch), stale-pick guard.
- Suites: 2406 PHPUnit / 4872 vitest green, typecheck, eslint, `phpcs -n`, build.
- Docs: `architecture.md` (the ships-references model), `desktop-themes.md` (delivery + the route), `javascript-reference.md` (`cssDeferred` row).
- Manual regression on :8889: boot in Legacy identical, System-default ↔ Legacy switching with the on-demand fetch, window opens + a plugin activate/deactivate cycle, ⌘K baseline commands navigating correctly.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-669-dc4e09f4dbc96f4433973bf935fa1c8eae552fc1.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
@claude can you review?

PR #674, opened 2026-08-25
Fix: Tile icons in the wallpaper can't display correctly some characters  

### LINKED ISSUE #663: Tile icons in the wallpaper can't display correctly some characters
Image

hbhalodia:
Fixes #663

### Problem

Tile labels show HTML entities instead of characters — `Alder &#038; Oak` rather than `Alder & Oak`.

### Cause

`get_the_title()` runs through `wptexturize()`, which encodes exactly the characters titles are full of: `&` → `&#038;`, apostrophes → `&#8217;`, quotes → `&#8220;`, ellipses → `&#8230;`. The file types stripped tags from that string but never decoded it, and `<os-tile>` paints the label with `textContent`, which renders an entity as itself.

### Fix

`openstation_plain_text_title()` decodes entities and strips tags, applied once in `OpenStation_File::serialize()` rather than in each `title()`. Every built-in file type is covered, along with any plugin's own, and the serialized `title` becomes a single contract: plain text. The recycle-bin row builder gets the same treatment.

Decoding runs **before** the tag strip, never after — `&lt;script&gt;` decodes into a real tag, and stripping second is what removes it.

## Screenshots or screencast 

|Before|After|
|-|-|
| Screenshot 2026-08-25 at 2 32 22 PM | Screenshot 2026-08-25 at 2 31 10 PM |

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-674-ab559492d94d46d635dcfd28707331cb613d05d7.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/helpers.php
```diff
@@ -220,6 +220,30 @@ function openstation_site_title() {
...
+function openstation_plain_text_title( $rendered ) {
+	$decoded = html_entity_decode(
+		(string) $rendered,
+		ENT_QUOTES,
+		get_bloginfo( 'charset' )
+	);
+
+	return trim( wp_strip_all_tags( $decoded ) );
+}
+
```
AllTerrainDeveloper, line 246: We already have a helper for exactly this, it was introduced by @Pranjal1423 in: https://github.com/WordPress/openstation/blob/ffdcf3267cfcdc824ac317ae91e03be36dca7e60/src/utils.ts#L465

@claude in your opinion, is it better for this particular case, doing it in the server side or in client side? We are focussing all our efforts into making OpenStation as fast as possible, for low tier hostings.

AllTerrainDeveloper:
@claude can you reply to my question above?

> We already have a helper for exactly this, it was introduced by @Pranjal1423
@claude in your opinion, is it better for this particular case, doing it in the server side or in client side? We are focussing all our efforts into making OpenStation as fast as possible, for low tier hostings.

AllTerrainDeveloper (approved):
All good then!

PR #676, opened 2026-08-25
Perf: shared admin-asset cache + hover window prewarming (opt-in), and the fixes that make the SW run on WordPress.com

AllTerrainDeveloper:
Two opt-in performance features (both **default off**, behind OpenStation Preferences → Features → Beta features) plus two service-worker bug fixes that ride along — one of which makes the SW run on WordPress.com hosting for the first time ever. Everything below was measured live on **openstation.blog** (WordPress.com production hosting, real network) with in-page instrumentation; the full protocol and raw run data live in the branch-side engineering report.

> **Scope of the numbers in this description.** Every measurement here is **feature-off vs feature-on within the same shell** — i.e. what this PR changes. It is *not* a comparison against classic wp-admin. That separate comparison was run afterwards and is in [this comment](https://github.com/WordPress/openstation/pull/676#issuecomment-5408745744), including the part that flatters us least: **the shell is ~2–2.8 s slower to cold-start** than a classic admin page (it loads its own document and bundles, then a full wp-admin screen inside a window), and it wins from the first revisit onward (1.7 ms window focus vs 1,632–2,822 ms per classic navigation). An [earlier comment](https://github.com/WordPress/openstation/pull/676#issuecomment-5408645169) stated that comparison **without** the boot cost in it and is superseded — read the correction, not the original. Nothing in that thread changes the deltas below, which are measured on top of whichever baseline you pick.

## What ships

1. **Shared admin-asset cache** (`adminAssetCacheEnabled`, Experimental). The root-scope SW serves versioned admin assets — Core CSS/JS, the `load-scripts.php`/`load-styles.php` concat blobs, plugin/theme assets carrying `?ver=` — from one origin-wide Cache Storage bucket shared by the shell and every window's chromeless iframe. Core paths are exact-URL cache-first (the `ver` param embeds the WP version — the same immutability contract Core expresses by serving the loaders with a one-year `Cache-Control`); plugin/theme assets are stale-while-revalidate so unversioned file edits self-heal on the next load; uploads, HTML, REST, AJAX and unversioned URLs are never touched. Bucket capped at ~500 entries (FIFO), version-keyed, flushed on every SW release. Policy is pure + unit-tested (`src/pwa/sw-policy.ts`). The per-user toggle is the filter default; `openstation_pwa_admin_asset_cache` can force or veto site-wide.
2. **Hover-intent window prewarming** (`windowPrewarmEnabled`, Experimental). A 180 ms mouse dwell on a dock tile builds that page's window hidden — the iframe loads while the user decides — and the real click adopts it: reveal, stack, focus, `os-window-opened` fired exactly then. The speculative window lives in a single slot **outside** the window stack, so sessions, taskbar, dock peek and Alt-Tab never see it; unclaimed prewarms self-destruct after 45 s with the close callback detached (no phantom events). Excluded by design: native windows, cross-origin URLs, native-remapped URLs.
3. **Bug fix — the SW never ran on WordPress.com.** The platform's nginx short-circuits `.js`-looking paths to the filesystem, so the virtual `/openstation/sw.js` route 404'd before WordPress executed — silently, on every WP.com install. The same bytes are now also served at the extensionless `/?openstation_sw=1` (path `/` grants root scope natively), and registration retries with `PwaConfig.swFallbackUrl` on failure. Verified live: openstation.blog registered through the fallback on its own, first SW execution ever on this host class.
4. **Bug fix — stale own SW blocked its successor forever.** The foreign-SW guard compared full script URLs, so a browser holding a pre-portal-move registration (`/desktop-mode/sw.js`, which now serves HTML and can never self-update) was stuck on a dead worker, observed live. Own legacy script URLs are now recognized and replaced by registering the current URL at the same scope; genuinely foreign SWs still block.

## Measurements (openstation.blog, production)

| Measurement | Without | With | Factor |
|---|---|---|---|
| Asset network requests per window open (cache on, warm bucket) | 117–143 | **0** | — |
| Per-asset fetch, browser-cold | 176–260 ms | **0.8–2.3 ms** | ~100× |
| 79 KB `load-styles.php` concat blob (PHP-generated per miss) | 277 ms median (254–521 ms) | **4.3 ms** | ~65× |
| `load-scripts.php` blobs | 188–219 ms median (peaks 504 ms) | **3–4.4 ms** | ~50× |
| Cold window open, asset network time (modeled from measured per-request costs, 117 assets / 6 connections) | ~3.7 s | **< 0.5 s** | ~8× |
| Window open → content loaded, prewarm hit | 2,000 ms | **3.3 ms** | ~600× |
| Document TTFB per window (SW-exempt by design: nonces) | 681–1,736 ms | unchanged — this is what prewarming hides | — |
| Boot cost with toggles off (`bin/boot-cost.mjs`, baseline vs branch) | 190 requests / 518.4 KB document | **+0 requests / +0.01% / +0.05% gz assets** | — |

Integrity: SW-served bodies were byte-identical to network fetches in every comparison run (core assets, plugin assets, and concat blobs). Cross-window sharing observed directly: a first-ever `tools.php` open started with 38/114 assets already bucketed by a different page's window, and `plupload.min.js` returned in 2 ms on a nominally cold fetch because another page had cached it.

## Worst case: parity — and why that trade is worth it

For users who never opt in, the worst case is **nothing**: measured +0 requests and +0.01% document on boot, zero behavior change. For opted-in users, the worst case is **parity, not regression**: on a warm browser against well-configured hosting, steady-state reopens measured 1,589 ms (cache on) vs 1,657 ms (off) — inside run-to-run noise, because the browser's own HTTP/memory cache already absorbs a warm steady state and parse/execute dominates. No measured scenario got slower than baseline. The bounded costs that do exist: a one-time ~400 ms populate pass on the first open after enabling (1,979 vs 1,589 ms, then gone); one speculative server-side page render per hover-intent (single slot, 45 s TTL) — the explicit trade the prewarm toggle buys; and a plugin that edits files without bumping its version serves one stale load before stale-while-revalidate self-heals, which is no worse than any far-future-cache host already treats such plugins.

The upside side of that trade is every case where the browser cache *doesn't* help — first visit of the day, evicted disk cache, hard reload, high-latency links, offline — where the measured per-request delta is two orders of magnitude, and window opens go from seconds to milliseconds on a prewarm hit. Parity in the best-case scenario, large wins in every degraded scenario, zero cost when dark: that is the definition of a worthwhile opt-in.

One thing this PR explicitly does **not** fix, now that it is measured: the shell's **~6.3 s cold boot** (vs 2.6–4.4 s for a classic admin page). Prewarming hides per-window document cost *after* boot; it does nothing for boot itself, and the asset cache is cold on a first-ever visit by definition. That number is the obvious next target and is called out in the report rather than left implied.

## Testing

- 14 new vitest cases: `sw-policy` classification matrix, SW-register fallback + legacy-recognition, window prewarm lifecycle (hidden mount, adoption, URL-mismatch discard, silent teardown, single slot, TTL). Full suite: **4,916 passed**.
- 11 new PHPUnit tests: filter/default/coercion for both settings, config preamble emission, fallback endpoint routing. Full suite green.
- `npm run lint`, `lint:php`, `typecheck`, `build` all green; `bin/boot-cost.mjs` before/after diff attached above.
- Deployed and exercised live on openstation.blog throughout development (both toggles, SW fallback registration, concat path via a temporary server-side shim since `page-optimize` normally replaces core concat).

## Docs

`docs/pwa.md` (caching policy table, shared-cache section, WP.com hosting note, fallback URL), `docs/hooks-reference.md` (new filter + helpers), `docs/javascript-reference.md` (`windowManager.prewarm()`/`discardPrewarmed()`, both `getOsSettings()` keys) — all updated in the same commits as the code.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-676-37a104d00b14a45136143cf7e46654c88e9be52f.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
### Follow-up measurement: OpenStation vs classic wp-admin, head to head

Measured on openstation.blog after the PR was opened — same host, same browser session, warm cache on both sides. Classic admin reached via the plugin's own `?desktop_mode_classic=1` escape hatch (vanilla wp-admin, shell suppressed), so this is like-for-like on identical hosting.

| Action | Classic wp-admin | OpenStation |
|---|---|---|
| Load Settings (`options-general.php`) | TTFB 843 ms · DCL 1,742 ms · **load 2,822 ms** · 146 subresources | first open ~1,589–1,979 ms; **3.3 ms** on a prewarm hit |
| Load Tools (`tools.php`) | TTFB 965 ms · DCL 1,323 ms · **load 1,632 ms** · 147 subresources | as above |
| **Return to a screen already visited** | full navigation again — **1,632–2,822 ms**, every time | **1.7 ms** median, 5.6 ms max (window focus) |

The last row needs no feature flags and is the structural argument for the shell itself: classic wp-admin discards the document on every click, so revisiting a screen costs what visiting it cost the first time (server re-render + ~146 subresources re-evaluated). OpenStation keeps windows alive, so a revisit is a focus call — a ~1,000× difference on the most common interaction in the admin, true today with both toggles off.

This PR's two features attack what remains, i.e. *first* opens of *new* screens: the shared asset cache takes the network out of the asset bill (~3.7 s → <0.5 s cold), hover prewarming takes the document wait out of the click (2,000 ms → 3.3 ms). Full protocol and raw data in the branch-side engineering report (Part 4).

AllTerrainDeveloper:
### Correction to the head-to-head numbers above: the shell's cold start belongs in the comparison

The previous comment compared classic **cold document loads** against OpenStation **window opens measured inside an already-booted shell** — an asymmetry, since the shell's own boot cost sat outside the table. Measured properly on openstation.blog, fresh navigations both sides:

| Action | Classic wp-admin | OpenStation |
|---|---|---|
| **Cold start** — admin from nothing (dashboard) | TTFB 1,221–2,230 ms · **load 2,619–4,364 ms** · 209–216 subresources | shell + bundles + dashboard window · **load 6,253–6,279 ms** · 228–229 subresources |
| Return to a screen already visited | **1,632–2,822 ms**, every time | **1.7 ms** median (5.6 ms max) |

**OpenStation is ~2–2.8 s slower to first usable screen.** That is structural, not a regression in this PR: the shell loads its own document and bundles and *then* loads a full wp-admin screen inside a window — two documents where classic loads one.

Break-even on that boot, using measured inputs (classic ≈3.5 s cold dashboard / 2.8 s Settings / 1.6 s Tools; OpenStation 6.3 s boot, ~1.6 s new-screen open, 1.7 ms revisit):

| Interaction | Classic cum. | OpenStation cum. | Ahead |
|---|---|---|---|
| open the admin | 3.5 s | 6.3 s | classic by 2.8 s |
| + Settings | 6.3 s | 7.9 s | classic by 1.6 s |
| + Tools | 8.0 s | 9.5 s | classic by 1.5 s |
| + back to Settings | 10.8 s | 9.5 s | **OpenStation by 1.3 s (crossover)** |
| + back to Dashboard | 14.3 s | 9.5 s | OpenStation by 4.8 s |
| + Settings again | 17.1 s | 9.5 s | OpenStation by 7.6 s |

Crossover is the **first revisit**; prewarming pulls it one interaction earlier. Honest scope: the shell wins any session with at least one revisit, and loses a single-screen visit-and-leave. None of this changes what *this PR* does — the two features and the two SW fixes are all measured deltas on top of whichever baseline you pick — but the comparison should not have been stated without the boot cost in it. Report Part 4 and the blog draft have both been corrected, and reducing that boot is the obvious next target.

AllTerrainDeveloper:
### Added: windows no longer load the admin bar they never draw

A window suppresses the admin bar entirely (`show_admin_bar` false + the `wp_admin_bar_render` removal), so `#wpadminbar` never reaches the DOM — but WordPress and the host kept enqueueing its scripts and styles, which loaded, parsed and executed against markup that does not exist. Once per window, every window.

Measured on openstation.blog, one Settings window, **before**:

| Asset | Size |
|---|---|
| `admin-bar.min.css` | 20.9 KB |
| `os-admin-bar.js` (ours) | 17.7 KB |
| `admin-bar.min.js` | 3.4 KB |
| notes `admin-bar-v2.js` | 3.1 KB (cross-origin) |
| notes `admin-bar-v2.css` | 2.4 KB (cross-origin) |
| `wpcom-admin-bar.js` | 0.6 KB |
| `notes-common-lite.min.js` | 0.5 KB (cross-origin) |
| `a8c-faux-inline-help.js` | 0.2 KB |
| **Total** | **48.8 KB + 3 cross-origin round trips** |

**After: all eight gone** (verified live — `adminBarAssetsStillLoading: []`), scripts 109 → 101, stylesheets 19 → 18, `s0.wp.com` requests 5 → 2, page renders and its form works, zero console errors. The shell is untouched: it still has `#wpadminbar`, both scripts, the stylesheet and the OpenStation toggle node.

Two passes were needed, and the second one is the interesting part. Dequeuing at `admin_enqueue_scripts` removed only half the family: WordPress.com's `wpcom-notes-*` handles enqueue *after* that hook, and core's `admin-bar` stylesheet then rode back in as their dependency — `WP_Dependencies::all_deps()` pulls a dequeued-but-registered handle back when something still queued declares it. `print_scripts_array` / `print_styles_array` run inside `do_items()` after every enqueue, dequeue and dependency walk, so they are the last word.

Handles are **dequeued, never deregistered**, so a script that genuinely declares one as a dependency still resolves it. The whole family ships in the defaults because leaving one member queued undoes the trim. Extensible via `openstation_chromeless_trimmed_scripts` / `..._styles` and the `openstation_chromeless_trimmed_assets` action.

Honest sizing: 48.8 KB out of ~5 MB of JavaScript that this (Gutenberg-plugin-heavy) site loads per window, so the wall-clock effect is inside this host's run-to-run noise — the win is the bytes, the parse/execute work, and the three cross-origin round trips, on every window a user ever opens. 10 new PHPUnit tests cover the defaults, both passes, the shell no-op, registration survival, and the filters.

AllTerrainDeveloper:
### Emoji polyfill dropped in windows too — plus a correction

**Correction first.** My earlier note labelled the emoji script "never used in admin". That was wrong, and our own measurement disproves it: `wp-emoji-release.min.js` (22 KB) genuinely loaded inside a window on current Chrome. The inline detection script tests the browser against the newest Unicode emoji set and pulls in Twemoji whenever anything is missing — browsers routinely lag the newest emoji, so it fires in practice. It is a real polyfill, not dead code.

**Dropping it in a window is still right**, and Core sets the precedent itself: `wp-admin/edit-form-blocks.php:42` runs `remove_action( 'admin_print_scripts', 'print_emoji_detection_script' )` on the block-editor screen. The only loss inside a window is that a very new emoji in admin content renders with the OS glyph instead of a Twemoji image — and the shell hosting these windows already requires service workers, custom elements, ES2020 and `:has()`. A browser clearing that bar does not need help drawing emoji.

Removal runs on `admin_init`, because `wp_enqueue_emoji_styles` rides `admin_enqueue_scripts` at the default priority and has already fired by the time the handle trim runs at `PHP_INT_MAX` — the same hook and reasoning as the admin-bar suppression in `helpers.php`. Filterable via `openstation_chromeless_trim_emoji` (default `true`).

**Verified live on openstation.blog** — window vs shell, same page load:

| | Window | Shell |
|---|---|---|
| `wp-emoji-release.min.js` | **gone** | still loads ✅ |
| `_wpemojiSettings` (inline detection) | **undefined** | defined ✅ |
| admin-bar assets | **none** | all present ✅ |
| `#wpadminbar` | not rendered | present ✅ |

Page renders, its form works, zero console errors.

**Cumulative effect of this trim work, per window:** scripts 109 → **100**, stylesheets 19 → **18**, JavaScript 5,063 KB → **4,892 KB**, `s0.wp.com` round trips 5 → **2**. That is ~71 KB and nine assets that no longer download, parse or execute in every window a user opens — while the shell keeps everything it actually draws. 13 PHPUnit tests now cover the module.

AllTerrainDeveloper:
### The chromeless bridge is a bundle now — the window document halved

Investigating "what else can we trim from a window?" turned up the biggest item in the page, and it was ours.

`includes/render/chromeless-bridge.php` held the bridge in a PHP nowdoc and printed it **inline into every chromeless window**: ~125 KB of unminified JavaScript, comments and all, the largest single block in the document. A document is the one asset no cache can touch, so that cost was paid in full on every window open, forever.

The code moves verbatim to `src/chromeless-bridge.js` and builds like every other bundle (**34.5 KB minified, 10.5 KB gzipped**). The four values that genuinely vary per request — menu payload, menu signature, content identity, soft-reload rules — are attached as a small inline `before` block on `window.__osChromelessData`, replacing the `str_replace` pass over placeholder comments. `in_footer` + `before` reproduces the old emission exactly: data defined, then the bridge running, same point in the document, no `defer`/`async` that would move it after the page's own footer scripts.

**Measured live on openstation.blog, same Settings window:**

| | Before | After |
|---|---|---|
| Window document, transferred | 76.4 KB | **37.2 KB** |
| Window document, uncompressed | 339 KB | **213.9 KB** |
| Inline scripts in the document | 285 KB | **160 KB** |
| Largest inline block | 125 KB (ours) | 66 KB (Jetpack's) |
| PHP file | 3,990 lines | **460 lines** |

Bridge verified functioning live: bundle served from cache (`fromNetwork: false`), `window.__osChromelessData` present with all four keys, `fetch` and `XMLHttpRequest` wrapped, tab strip built (11 tabs), window title reported, zero console errors.

**Re-ran the Settings → Writing → Reading × 3 journey** (the one that previously came out a tie): total **20.7 s → 19.4 s**, median per screen **2.77 s → 2.30 s** (−17%). It is now level with classic wp-admin's 19.4 s rather than behind it. Honest framing: halving the document moved the total by ~6%, because the bytes were never the expensive part — waiting for the server is. Every screen is simply delivered in half the bytes now.

**One hardening that came out of the test suite:** `WP_Dependencies::enqueue()` silently no-ops on an unregistered handle, and plugins that rebuild `WP_Scripts` wholesale are a documented shape (that is what `asset-guard.php` exists for). The bridge is the one script a window cannot do without, so it re-registers if the handle has gone missing. This was a genuine latent fragility, surfaced by a full-suite run where another suite's fixtures had rebuilt `WP_Scripts`.

Tests that assert on bridge behaviour now read the source that builds into the bundle rather than slicing it out of the nowdoc — 5 test classes updated. Full suites green: **4,916 JS / 2,435 PHP**, phpcs, typecheck, lint, build.

AllTerrainDeveloper:
@claude can you review this?

Issue #171, opened 2026-05-12
We updated a plugin from updates page then links opening in a new tab


demironmanx:
[Go to Plugins page](https://website.com/wp-admin/plugins.php) | [Go to WordPress Updates page](https://website.com/wp-admin/update-core.php)

This is causing screen on screen.

epeicher:
I have reproduced this by navigating to  Dashboard -> Updates -> Update to latest 7.1 nightly

<img width="3336" height="2742" alt="Image" src="https://github.com/user-attachments/assets/05ec35f0-6ab3-4c35-acc1-ac3bb45ab974" />

PR #678, opened 2026-08-25
Perf: the shell asks the service worker for a screen before you click it

AllTerrainDeveloper:
The shared asset cache from #676 took the network out of a window's **assets**. It could never touch the **document** — admin HTML carries nonces and per-request screen state, so it is uncacheable by construction. And on a live install the document is the majority of a window open: measured at **~2.1 s of a ~3.8 s tab click**, against ~1.7 s for everything the browser then does with it.

That cost does not have to be paid *after* the click.

## The gap this closes

Two halves existed and were never connected:

- The **shell** already knows every URL a window can reach (dock items and submenu tabs come straight from the menu payload) and already reads hover intent.
- The **service worker** already receives the `fetch` event for every iframe navigation — and deliberately declines to answer it.

Now a hover on a submenu tab posts `os-speculate-doc` to the worker, which fetches that screen **once** and holds it; the iframe's own navigation is then answered from those bytes.

**This is explicitly not "keep the tab alive."** Nothing rendered is retained: no DOM, no live iframe, no memory beyond a response body dropped after 30 s. The page is still built fresh — it is simply built while the user is still deciding.

## Measured on openstation.blog (production WordPress.com hosting, real network)

| Tab click | Time from click to screen ready |
|---|---|
| No hover (control) | **1,997 ms** / **3,605 ms** |
| ~3 s hover | **961 ms** / **1,000 ms** / **1,002 ms** / **1,014 ms** |
| 600 ms hover (partial head start) | **1,014 ms** (from a 3,605 ms control) |

Roughly **2–3.5× faster**, and strikingly consistent once the document is in hand. The residual ~1 s is the browser parsing and executing that page's JavaScript — the half no prefetch can remove, and exactly what the earlier breakdown predicted:

| Where a tab click's time went (measured before this change) | | |
|---|---|---|
| Server building the document | 2,119 ms | 56% |
| Browser running the page's JavaScript | 1,671 ms | 44% |
| Downloading assets | 0 ms | 0% (already handled by #676) |

This PR removes the first row from the user's wait. The second row is untouched and honestly out of reach without retaining the rendered page, which we deliberately are not doing.

## The bug the first measurement caught

The first implementation held only the **settled** response. Two runs disagreed wildly — 5 ms in one, 1,001 ms in another — and the reason mattered: a click landing while the fetch was still in flight found nothing waiting and **started a second request for the same screen**. Full price for the user, double load for the server.

The store now holds the **promise**, so a mid-flight click awaits the request already running and inherits whatever head start the hover bought. That is why 600 ms of hover was enough to turn a 3,605 ms click into 1,014 ms.

## Why answering an iframe navigation is safe here

The fetch handler otherwise refuses to answer iframe navigations, and that guard stays. The hazard it exists for (#171) is the worker **re-fetching** an iframe request: Chrome then sends `Sec-Fetch-Dest: empty`, the server's chromeless detection falls through, and the entire desktop renders inside a window.

A speculative document is never re-fetched. It is fetched once, ahead of time, from a URL the shell built with `openstation_chromeless=1` present — and the server checks that query flag **before** it consults Sec-Fetch. So the held bytes are already correctly chromeless, and the fallback path (speculative fetch failed → fetch normally) is safe for exactly the same reason.

## Safety

- **Speculation never acts.** Any URL carrying `action`, `action2`, `_wpnonce`, `nonce` or `delete_all` is refused, so a hover can never activate a plugin, empty a trash, or apply an update.
- Same-origin, `/wp-admin/` only, and the chromeless flag must be present.
- **Single-use, exact-URL match, 30 s TTL**, capped at 6 held documents with oldest-first eviction.
- Redirected or non-200 responses are never held — a redirect means the server wanted the user somewhere else, and replaying the destination under the original URL would hide that.
- Gated on the existing **hover-prewarm** opt-in (`windowPrewarmEnabled`), delivered to the worker through the same `self.__OS_SW_CONFIG` preamble the asset cache uses. Off by default; zero behaviour change until a user opts in.
- Pointer-only (`pointerType === 'mouse'`) — touch has no hover, so there is no intent to read.

## Testing

`npm run typecheck`, `lint`, `lint:php`, `build` green. Full suites: **4,972 JS**, **2,495 PHP**. New coverage for the config flag's independent parsing in `sw-policy.test.ts`. Verified live on openstation.blog throughout: correct screens rendered (`Writing Settings`, `Media Settings`, `Discussion Settings`, `Permalink Settings` all confirmed by heading), forms intact, zero console errors.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-678-e41f84a016660bab3a6099e21ebe861ec7c64d1e.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/docs/pwa.md
+++ b/docs/pwa.md
@@ -79,6 +79,7 @@ OpenStation and you want OpenStation to take over the install path.
 | `/wp-content/plugins/desktop-mode/assets/**.js` | Network-first with `cache: 'reload'` + cache fallback | JS bundles change per deploy — a fresh deploy reaches online users on the next load, with no stale-revalidate window where a freshly-pushed fix is invisible. The cache still serves offline users. |
 | **Opt-in** — versioned Core statics (`/wp-admin/**`, `/wp-includes/**` with `?ver=`) and `load-scripts.php` / `load-styles.php` | Exact-URL cache-first (`os-admin` bucket) | The `ver` query embeds the WordPress version, so bytes behind a URL only change when the URL changes — the same contract Core expresses by serving the loader endpoints with a one-year `Cache-Control`. A warm window-open costs zero HTTP requests for these. Requires the `openstation_pwa_admin_asset_cache` filter (default off). |
 | **Opt-in** — versioned plugin/theme statics (`/wp-content/plugins|themes/**` with `?ver=`) | Stale-while-revalidate (`os-admin` bucket) | Same `ver` contract in principle, but authors edit files without bumping versions often enough that cache-first would pin stale bytes; SWR serves instantly and self-heals on the next load. Uploads are excluded (quota, thumbnail regeneration keeps the URL). |
+| **Opt-in** — an iframe navigation to a document the shell asked for early | Served from the held response (never re-fetched) | The document is the one thing that can never be cached — admin HTML carries nonces — and it is the majority of a window open. Speculation does not make it cacheable; it moves the wait to before the click. See [Speculative documents](#speculative-documents-opt-in). |
 | Navigation requests under our scope | Network-first with offline fallback | wp-admin HTML carries nonces and per-request screen state; caching it would desynchronise the user. The fallback is a tiny inline placeholder so an offline user sees something coherent. |
 | REST / AJAX / non-asset GETs / unversioned asset URLs | Pass-through (no SW handling) | Same reason as navigation — auth-bound dynamic content must hit the network; an asset URL without a `ver` cache-buster carries no immutability contract. |
 | `install`-time precache | A handful of CSS files, the three critical-path JS bundles (`desktop.min.js`, `window-system.min.js`, `shell-overlays.min.js`), and the plugin logo | Just enough to render the offline shell skeleton. Anything else is picked up at runtime by the caching paths above. |
@@ -134,6 +135,46 @@ The cache is keyed by version (`os-static-<v>`,
 `os-runtime-<v>`). The `activate` step deletes any cache whose
 key doesn't carry the current version, so a deploy doesn't accumulate
 stale buckets.
+
+## Speculative documents (opt-in)
+
+The shared asset cache removes the network from a window's **assets**. It can never touch the **document**: admin HTML carries nonces and per-request screen state, so it is uncacheable by construction — and it is the majority of a window open (measured at ~2.1 s of a ~3.8 s tab click on production hosting).
+
+That wait does not have to happen *after* the click. The shell knows every URL a window can reach; the worker sees every iframe navigation. Connecting them lets a document be fetched **while the user is still deciding**, and the navigation that follows is answered from those bytes.
+
+This is not "keep the window alive": nothing rendered is retained. No DOM, no live iframe, no memory beyond a response body dropped after 30 seconds. The page is still built fresh — just early.
+
+Two triggers, one mechanism:
+
+| Trigger | What happens |
+|---|---|
+| **Hovering a submenu tab** | The shell posts `os-speculate-doc`; the worker fetches that screen and holds it. |
+| **Booting the shell** | The worker replays the previous session's restore list the moment the shell's own navigation arrives — *before* the server has finished building the shell document, so the two renders overlap instead of running back to back. The list is persisted from `os-remember-session`, which the session saver posts on every save. |
+
+Measured on production hosting: window-document TTFB **1,353 ms → ~1 ms**, whole shell boot **6,492 ms → ~4,700 ms**, a hovered tab click **~3,600 ms → ~1,000 ms**.
+
+### Rules
+
+- **Speculation never acts.** A URL carrying `action`, `action2`, `_wpnonce`, `nonce` or `delete_all` is refused, so a hover can never activate a plugin or empty a trash. Same-origin, `/wp-admin/` only, and the `openstation_chromeless` flag must be present.
+- **Held documents are single-use and expire after 30 s**, capped at 6 with oldest-first eviction. A document is a moment-in-time view carrying nonces; replaying one twice would show a superseded page.
+- **The store holds the in-flight promise, not the settled response**, so a navigation landing mid-fetch joins the request already running instead of starting a second one for the same screen.
+- **Answering an iframe navigation is safe only because the response is never re-fetched.** The worker otherwise refuses iframe navigations: re-fetching one makes Chrome send `Sec-Fetch-Dest: empty`, the server's chromeless detection falls through, and the whole desktop renders inside a window. A speculative document is fetched once, ahead of time, from a URL carrying the chromeless flag — which the server reads *before* it consults Sec-Fetch.
+- The speculative fetch is a plain same-origin GET and does not forward `Referer` or `Accept-Language` from the navigation it stands in for. Admin screens do not branch on either (locale comes from the user's profile, server-side).
+- Gated on the **hover-prewarm** opt-in (`windowPrewarmEnabled`), delivered to the worker as `windowPrewarm` in the `self.__OS_SW_CONFIG` preamble. Off by default.
+
+### Worker message surface
+
+Both messages are posted to `navigator.serviceWorker.controller` and are ignored unless the opt-in is on.
+
+```js
+// Fetch this screen now; hold it for the navigation that follows.
+{ type: 'os-speculate-doc', url: '<absolute same-origin chromeless URL>' }
+
+// Remember these screens for the NEXT boot's replay.
+{ type: 'os-remember-session', urls: [ '<absolute>', … ] }
+```
+
+The shell-side helpers are `speculateDocument( url )` and `rememberRestoreTargets( urls )` in `src/pwa/speculate.ts`. Policy lives in `src/pwa/sw-policy.ts` (`isSpeculatableDocument`) and the store in `src/pwa/speculative-store.ts`; both are pure and unit-tested.
 
 ## PHP surface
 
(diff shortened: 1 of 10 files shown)


AllTerrainDeveloper:
@claude can you review?

AllTerrainDeveloper:
### Extended: the worker now starts your windows' documents while the server is still building the shell

Same collaboration, applied to boot instead of hover — and this is the bigger half.

A shell boot is **two independent server renders run strictly one after the other.** Measured on openstation.blog before this change:

```
    0ms ─── shell document requested
 3172ms ─── shell HTML arrives (3.2s of server time)
 3876ms ─── only NOW is a window document requested
 6491ms ─── everything ready
```

The window's document does not depend on a single byte of the shell's. It simply cannot be *asked for* until the shell's HTML has arrived and its JavaScript has run and built an iframe. And the shell can never fix this from inside itself: by the time its code exists, the 3.2 seconds are already spent.

The service worker can. It is woken by the shell's *own* navigation — before the server has finished building it — so it can start the window fetches right then, in parallel. It only needs to know what to fetch, which is what the session saver now tells it: every save posts the restore list, the worker persists it to a cache, and the next boot replays it the moment the shell navigation arrives.

**Measured, same site, same session:**

| | Before | After |
|---|---|---|
| Window document TTFB | 1,353 ms / 1,247 ms | **9 ms / 5 ms** → **1 ms / 1 ms** |
| First window ready | 5,861 ms | **3,323 ms / 3,309 ms** |
| Whole shell boot (`loadEventEnd`) | 6,492 ms | **4,736 ms / 4,700 ms** |

**~1.8 s off every boot, and the first window usable 2.5 s sooner.** The window documents now arrive effectively instantly, because they were fetched alongside the shell rather than after it.

#### The bug the measurement caught

The first live run was suspicious: one window's TTFB halved (1,353 → 687 ms) while the other sat unchanged at 1,246 ms. The cause was a wrong assumption of mine — `replayedThisBoot` was a boolean, but **a service worker outlives any single page load**; it stays resident across navigations and can be reused for hours. So the replay fired on the first shell load that worker ever saw and never again. Replaced with a short time-based throttle, which is all it ever needed since `beginSpeculation()` already de-duplicates by URL. The numbers above are after that fix.

Reuses the speculative store wholesale, including the in-flight promise semantics, so a window whose iframe appears mid-fetch waits on the request already running rather than starting a second. Same safety rules throughout: same-origin, `/wp-admin/` only, chromeless flag required, nothing carrying `action`/`_wpnonce`/`delete_all`, capped and short-lived, gated on the same opt-in — and fire-and-forget, so a boot never waits on speculation. Zero console errors across the runs.

AllTerrainDeveloper:
### Review fixes pushed (`b339274e`)

**1. The `asked` set — you were right, and it was worse than a missed optimisation.** It silently disabled the feature after first use. Never cleared in production, while the worker's copy is single-use and 30 s TTL'd — so once a tab had been speculated once, whether the click consumed it or it simply timed out, every later hover over that tab fell through to a plain navigation with no speculation at all. Since users bounce between the same few screens constantly, the repeat visits were exactly what the feature should accelerate and exactly what it abandoned.

Replaced with a **1 s throttle**: long enough to damp a pointer crossing a tab's icon and label, far shorter than the worker's hold, so a genuine return visit always gets through. The worker stays the real de-duplicator (`beginSpeculation()` ignores a URL it already holds), which is what the original comment claimed and the set contradicted.

**2. `docs/pwa.md`** now carries the whole surface: a row in the caching-policy table, a "Speculative documents (opt-in)" section covering both triggers, the safety rules, the measured numbers, and the worker message shapes (`os-speculate-doc`, `os-remember-session`) plus the `windowPrewarm` preamble flag.

**3. Real coverage for the load-bearing parts.** To make them testable rather than test around them:

- `isSpeculatableDocument` moved to `sw-policy.ts` (pure, already the home of the other policy predicates).
- The store moved to a new `speculative-store.ts` — a small class with an injectable clock.

15 new cases in `tests/vitest/speculative-documents.test.ts` covering: the `action`/`action2`/`_wpnonce`/`nonce`/`delete_all` denylist that keeps a hover from ever performing a state change; the chromeless-flag requirement; **promise-vs-settled-response** semantics (the 5 ms / 1,001 ms bug — a claim taken mid-flight still resolves to the response); single-use taking; TTL expiry; oldest-first eviction past the cap; and shell-side origin rejection, no-controller no-op, and the throttle-then-allow behaviour that pins issue 1 so it cannot regress silently.

**Minor notes, both now documented in code rather than left implicit:**

- `pointerover` over `pointerenter` is deliberate — `pointerenter` does not bubble so it cannot be delegated, and tabs come and go too often to bind per tab. The extra events are absorbed by the throttle and the worker's de-duplication.
- The speculative fetch not forwarding `Referer` / `Accept-Language` is a considered tradeoff: admin screens do not branch on either (locale is server-side from the user profile), and fabricating headers we never received would be guessing. If a screen ever did vary by them, the fix is to exclude that screen, not to invent headers.

Full suites green after the refactor: **4,987 JS** (was 4,972), **2,495 PHP**, typecheck, lint, phpcs, build. Re-verified live on openstation.blog that the extraction changed no behaviour — boot still shows window-document TTFB of **1 ms / 2 ms** and a **4,850 ms** shell load, with zero console errors.

AllTerrainDeveloper:
### Verified live: the repeat-visit fix works, and the "missing tabs" was my test harness, not a bug

**On the tab strip.** I chased this as a possible pre-existing bug and it is not one — the product is correct and I was measuring wrong. Restored windows carry their submenu tabs properly (`options_general_php` restores with `submenu.length: 10` and **11 tabs** rendered), because `src/boot/session.ts` resolves the owning dock entry and passes `submenu: dockEntry?.submenu`.

The empty strips I kept hitting came from my own harness calling `windowManager.open({ id, url, title })` **without `submenu`** — something the dock never does (`dock.ts` always passes `item.submenu`). A window opened with no submenu correctly has no submenu tabs. Opening it the way the dock does gives 11 tabs immediately. No fix needed; nothing to change.

**On the actual fix (issue 1), measured on openstation.blog** using a dock-shaped open:

| Visit | Hover | Click → ready |
|---|---|---|
| Writing, 1st visit | 2.5 s | **999 ms** |
| Reading (control) | none | 1,999 ms |
| **Writing, 2nd visit** | 2.5 s | **1,002 ms** |
| Permalinks (control) | none | 2,000 ms |
| **Writing, 3rd visit** | 2.5 s | **995 ms** |

Before the fix, the 2nd and 3rd visits to Writing would have been silent no-ops falling through to a plain ~2 s navigation, because the URL was already in the permanent `asked` set. They now behave identically to the first visit. That is precisely the "users bounce between the same few submenu tabs repeatedly" case you called out, and it is worth roughly **2× on every visit, not just the first**.

Controls in the same session confirm the baseline hasn't moved: un-hovered clicks still land at ~2.0 s.

AllTerrainDeveloper:
@claude would you mind one second review?

AllTerrainDeveloper:
### Both fixed (`e41f84a0`)

**1. The opt-in gate.** You're right that this contradicted the stated design, and the contradiction was load-bearing — "off by default; zero behaviour change until a user opts in" appears in both the PR body and `docs/pwa.md`, and it simply wasn't true for this call. `rememberRestoreTargets()` now goes through a shared `noteRestoreTargets()` helper that reads `windowPrewarmEnabled` exactly as `speculateTabDocument()` and the dock's equivalent do.

I also gated the worker's `os-remember-session` handler on `CONFIG.windowPrewarm`, so an opted-out browser cannot be made to write caches by a stray message from anywhere. Belt and braces, but the claim is about what an opted-out user pays, and it should hold regardless of who is posting.

**2. The unload path.** Fixed, and I agree it's the case that matters most. `flushImmediately()` now calls the same helper before sending its beacon. Its snapshot is the most accurate one in the system — taken as the tab goes away, after the last window the user closed or opened — so replaying whichever debounced save happened to land last meant the next boot could speculate a document for a window that no longer exists and miss one opened seconds earlier. Exactly the "close it and come straight back" moment the feature is aimed at.

Both paths now share one helper, so they cannot drift apart again.

**3. Docs tightened** as you suggested, now that #2 is addressed: the list is posted from **both** save paths (debounced and `pagehide` beacon), with a note on why the unload one matters; and the opt-in line now says it is checked on both sides, with a user who never touches the setting not paying so much as a message.

**Coverage:** a new test drives the real session saver with the toggle off and asserts no `os-remember-session` is posted, so "off by default" can't silently regress the way the `asked` set did. Suites green: **4,988 JS**, **2,495 PHP**, typecheck, lint, phpcs, build. Deployed to openstation.blog.

PR #680, opened 2026-08-25
Fix: a spinner from one tab painted over another tab's finished page

AllTerrainDeveloper:
## The bug

A window's loading overlay is a **single element on the window body**, armed when the **primary** iframe starts a navigation. But the body hosts *every* tab's iframe, and `switchToTab()` only toggles their `display` — nothing reconciled the overlay with the tab actually on screen.

So:

1. Click a submenu tab → the primary iframe starts loading, the spinner paints.
2. Switch to a kept-alive external tab whose content has been ready for minutes.
3. Sit under a spinner belonging to a *different* tab until a fade timer happens to end.

Reproduced live on openstation.blog before the fix — body `--loading` already cleared, overlay still carrying its visible class, sitting on top of a finished page:

| Moment | overlay visible |
|---|---|
| Primary tab mid-navigation | true *(correct)* |
| Switched to a fully-loaded external tab | **true** ← the bug |
| ~3 s later, when the fade timer ended | false |

## The fix

External tabs never drive this overlay — they load in their own iframe, hidden, and their readiness is their own business. So the rule is simply: **show the spinner only while the primary tab is the one on screen**, and restore it on the way back if that iframe is still loading.

`syncLoadingOverlayToTab()` deliberately **leaves the body's loading class alone** — that class is the primary iframe's own state, and the handler that clears it needs to find it there. Only the painted spinner is suppressed, so nothing is faked into looking "finished". Coming back re-arms through `scheduleLoadingOverlayShow()`, which re-reads the load-start stamp: a navigation already past the 120 ms grace period paints at once, a fresh one still waits it out.

Verified live after the fix, same scenario:

| Moment | overlay visible |
|---|---|
| Primary tab mid-navigation | true *(unchanged — correct feedback)* |
| Switched to a loaded external tab | **false** |
| Still on that tab 1.5 s later | **false** |
| Switched back while the primary is still loading | **true** *(restored)* |
| After the primary settled | false |

## Not a bug: the two kinds of tab

Worth writing down, because the report's premise is a reasonable misreading of the UI:

- **Submenu tabs** (General / Writing / Reading in a Settings window) share **one** iframe and genuinely navigate it — the window I inspected had 11 tabs and exactly 1 iframe. They are *not* kept in memory, and a loader there is honest.
- **External tabs** (a link that opened as a tab inside the window) each get **their own** iframe and stay alive. Switching between those is a `display` toggle, and correctly showed no loader.

## Testing

6 new cases in `tests/vitest/tab-switch-loading-overlay.test.ts`: suppression on switching away, the body class being left intact, restoration on returning to a still-loading primary, no spinner conjured for a settled primary, the grace period still honoured on return, and a no-op on a window without an overlay yet.

One of them caught a real subtlety while I was writing it — an early fixture omitted the load-start stamp, so the restore path silently deferred behind the 120 ms delay instead of painting. The fixture now mirrors what `createWindowElement` actually builds.

Full suites green: **4,994 JS**, typecheck, lint, build. Deployed to openstation.blog and verified there.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-680-d229c80cdf4b723bc65c08ed93835c9f1928cd38.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/src/window/loading.ts
+++ b/src/window/loading.ts
@@ -39,6 +39,7 @@ import {
 	LOADING_HANDOFF_BODY_CLASS,
 	LOADING_STARTED_ATTR,
 	removeLoadingOverlay,
+	scheduleLoadingOverlayShow,
 	stampLoadingStart,
 } from './dom';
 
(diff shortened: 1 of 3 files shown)


AllTerrainDeveloper:
@claude can you review?

PR #682, opened 2026-08-26
Perf: windows stop paying for a command palette they never show

AllTerrainDeveloper:
> **Numbers corrected in `bca279bc`.** The first revision of this PR claimed a Settings window went from 137 requests / 14.49 MB to 80 / 3.36 on a WooCommerce-heavy site. That build was misclassifying Core's block-editor stack as palette contributors and dropping it — the saving was substantially collateral damage. The measurements below are from the corrected walk. Earlier browser-timing figures are withdrawn for the same reason.

## The finding

Opening **Settings** in a window pulls a contiguous Gutenberg dependency chain — react, react-dom, `components.js`, `block-editor.js`, `core-data`, `blocks`, `sync` — terminating at `core-commands.js`.

All of it exists to power **Core's ⌘K command palette**, which the shell replaces.

The shell handles the keystroke in the parent frame and only asks a window for its commands when the palette is actually opened — `onPaletteOpened()` in `src/commands/iframe-bridge.ts` is the only caller of `os-commands-subscribe`. **The harvest was already strictly on demand; the runtime that answers it was loaded eagerly, in every window**, on the chance the user might one day press ⌘K while that window held focus. An iframe is its own JavaScript realm, so an HTTP cache hit buys nothing — every window parses and executes all of it again.

The shell already defers this (#664). Windows were excluded on the reasoning that the iframe runtime feeds the harvest — true, but it doesn't require paying at boot.

## Why it has to be a family trim

Unhooking Core's enqueue alone recovers almost nothing. Astra's `command-palette.js` and WooCommerce's `command-palette.js` / `command-palette-analytics.js` each declare `wp-commands`, so `all_deps()` pulls the chain straight back in on their behalf. That is the same lesson the admin-bar trim in this file already records, so this follows the same shape.

## Why it must also stop at Core's own packages

This is the correction in `bca279bc`, and it is the subtle part of the change.

`wp-block-editor` declares `wp-commands` **directly**:

```
wp-block-editor => …, wp-blocks, wp-commands, wp-components, …
wp-editor       => …, wp-block-editor, wp-commands, …
```

It does so because the editor *registers* commands into the palette store — the dependency runs the **opposite way** from a palette extension, which *is* the palette. A plain closure walk cannot tell those apart, so the first revision convicted the entire block-editor stack and dropped it, taking every plugin block-registration script above it along. Measured on `customize.php`: the block-widgets panel lost `wp-block-editor`, and Contact Form 7's and MailPoet's block scripts went with it.

Two rules fix it:

1. **Core packages are never trimmed as dependents.** They are libraries; `WP_Dependencies` already knows whether one is needed — if something queued requires it, it resolves and stays; if the palette was its only consumer, it falls out on its own.
2. **The walk never travels *through* a Core package to reach a root.** Reaching the palette via `wp-block-editor` says something about the editor, not about the handle under test. A genuine palette extension names the palette in its own chain, which is exactly what Astra's and WooCommerce's scripts do.

## Exemptions

`post.php`, `post-new.php`, `site-editor.php` and `widgets.php` are exempt: they load the Gutenberg chain for their own reasons, so the palette rides along for the cost of `commands.js` + `core-commands.js`, and those are the screens whose stores hold commands worth harvesting ("Duplicate block", pattern commands).

`customize.php` **was** exempt and no longer is. It was added on the assumption that the block-widgets panel made it an editor screen; measured, the exemption bought nothing. With the corrected walk the Customizer keeps its own chain either way, so the trim removes only the palette there. Verified on that screen: `wp-block-editor`, `wp-blocks`, `wp-components` and all three plugin block scripts survive; only `commands`, `core-commands` and the two palette extensions go.

---

# Measurements

`npm run perf:boot-cost` — deterministic, measures what the server printed rather than what a cache happened to hold. **The size of the win depends entirely on whether anything else on the site pulls the Gutenberg runtime**, so both shapes are reported.

## Clean install — WordPress + OpenStation only

The palette is the sole consumer, so the whole chain goes. `options-general.php` in a window:

| | Requests | Raw | Gzipped |
|---|---|---|---|
| Before | 85 | 4.06 MB | 1.26 MB |
| After | **43** | **0.83 MB** | **0.25 MB** |
| | **−49%** | **−80%** | **−80%** |

## A plugin-heavy install — WooCommerce, ACF, Astra, Contact Form 7, MailPoet

WooCommerce enqueues its data packages on every admin screen and those pull the Gutenberg runtime **independently of the palette**, so the chain stays and the remaining win is modest. Nothing breaks, which is the point of the correction.

| Screen | Before | After | |
|---|---|---|---|
| `options-general.php` | 137 req / 14.49 MB | 127 req / 14.14 MB | −10 req |
| `tools.php` | 109 req / 13.19 MB | 99 req / 12.83 MB | −10 req |
| `edit.php` | 115 req / 13.32 MB | 105 req / 12.96 MB | −10 req |
| `customize.php` | 340 req / 32.57 MB | 332 req / 32.36 MB | −8 req |

What is removed on such a site is Core's palette runtime, the two third-party palette extensions, the 19.6 KB inline `initializeCommandPalette( … )` blob, and the `$menu` / `$submenu` walk (a `current_user_can()` and `menu_page_url()` per entry) that PHP runs to build it — per window.

## Verified on a production install

Checked on a live WordPress.com site: `core-commands.js`, `commands.js` and `block-editor.js` are absent from a Settings window. `components.min.js` is still loaded there — by a genuine non-palette consumer, not the palette — which is the dependency walk behaving as intended. No A/B was run against production, since that would mean editing a live site.

---

## New hooks

Documented in `docs/hooks-reference.md`:

- `openstation_chromeless_trim_command_palette` — whether to trim (default `true` off block-editor screens)
- `openstation_command_palette_root_handles` — what counts as a palette root
- `openstation_command_palette_family` — the resolved family; the escape hatch for a bundle that registers commands *and* draws its own screen
- `openstation_chromeless_trimmed_command_palette` (action)

## Testing

- Full PHPUnit: **2516 tests, 0 failures**; 19 in `openStationChromelessCommandPalette.php`, including regression tests for the Core-package rule (a core handle declaring `wp-commands` is never trimmed; a plugin block script reaching the palette only via `wp-block-editor` is never trimmed; a palette extension behind its own script still is) and both directions of the style-side print-list pass
- `npm run lint:php` (phpcs -n): clean
- `npm run build`: no drift

**Manual check requested:** open Settings / Tools / Users in windows and confirm they render and behave normally; press ⌘K with a window focused to confirm the shell palette still works; open a post in a window to confirm block commands still reach the palette; and open the Customizer in a window to confirm the block-widgets panel still loads.

## Follow-up, not in this PR

The same family still anchors the **shell boot**. Probed and reverted: trimming it on shell pages took boot from 190 req / 15.74 MB to 142 / 5.19 on the plugin-heavy site — but that probe predates the Core-package correction, so treat the figure as an upper bound until re-measured. Left out because it touches the palette-replay/harvester architecture rather than the established chromeless-trim pattern.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-682-bca279bc2d010d6c91168fa1e0bc311ff97933e4.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/docs/hooks-reference.md
+++ b/docs/hooks-reference.md
@@ -1803,6 +1803,56 @@ add_action( 'openstation_chromeless_trimmed_assets', static function () {
 
 ---
 
+### `openstation_chromeless_trim_command_palette` — Experimental
+
+Whether Core's command-palette runtime is dropped inside a window. Default `true` on every screen except block-editor screens.
+
+The shell owns ⌘K: the keystroke is handled in the parent frame, and the parent only asks a window for its commands when the palette is actually opened (`os-commands-subscribe`). The runtime that answers is nevertheless enqueued by Core on every admin page, so each window paid for it eagerly on the chance the user might one day press ⌘K while that window held focus. Measured on a live install opening **Settings** in a window, that was 43 files, **10.66 MB raw / 1.94 MB gzipped — 73.6% of everything the window downloaded** — parsed and executed again in each window's own JavaScript realm, where an HTTP cache hit buys nothing.
+
+`post.php`, `post-new.php`, the site editor and the widgets screen are exempt: they load the same Gutenberg chain for their own reasons, so the palette rides along for the cost of `commands.js` + `core-commands.js`, and those are exactly the screens whose stores hold commands worth harvesting ("Duplicate block", pattern commands). Everywhere else the store holds only the WordPress baseline, which the shell already publishes from its own lazily-loaded runtime.
+
+```php
+// Keep Core's palette (and its Gutenberg chain) inside windows.
+add_filter( 'openstation_chromeless_trim_command_palette', '__return_false' );
+```
+
+### `openstation_command_palette_root_handles` — Experimental
+
+The handles treated as command-palette roots. Defaults: `wp-commands` (the `core/commands` store package) and `wp-core-commands` (WordPress's baseline command set). A queued script whose dependency closure reaches one of these is considered a palette contributor.
+
+```php
+apply_filters( 'openstation_command_palette_root_handles', string[] $handles );
+```
+
+### `openstation_command_palette_family` — Experimental
+
+The full set of handles dropped by the palette trim: the roots plus every scanned handle that reaches one of them, directly or transitively.
+
+This is a family trim for the same reason the admin-bar trim above is: dropping the roots alone saves nothing while a single dependent survives, because `WP_Dependencies::all_deps()` pulls the whole chain back in on that dependent's behalf. Measured with only Core's enqueue deferred, a Settings window still pulled 14.28 MB of the original 14.49 MB — Astra's `command-palette.js` and WooCommerce's `command-palette.js` / `command-palette-analytics.js` each declare `wp-commands` and were still queued.
+
+Note that handles pulled in only *as dependencies of* a palette contributor need no special handling: they stop being requested once their last dependent leaves, and stay if anything else still needs them. WooCommerce's `wc-settings` (a 96 KB inline blob) disappears from a Settings window for exactly that reason, and remains on a WooCommerce screen that uses it.
+
+Use this filter as the escape hatch for a script that registers commands *and* renders part of its own admin screen — a distinction the dependency walk cannot make.
+
+```php
+// Keep a bundle that registers commands but also draws its own screen.
+add_filter( 'openstation_command_palette_family', static function ( array $family ) {
+    return array_values( array_diff( $family, array( 'my-plugin-admin-app' ) ) );
+} );
+```
+
+### `openstation_chromeless_trimmed_command_palette` — Experimental (action)
+
+Fires after the palette family is dropped in a window.
+
+```php
+add_action( 'openstation_chromeless_trimmed_command_palette', static function () {
+    wp_dequeue_script( 'my-plugin-palette-extras' );
+} );
+```
+
+---
+
 ### `openstation_native_window_allowed_html` — Experimental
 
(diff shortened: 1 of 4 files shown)


AllTerrainDeveloper:
@claude can you review this pr?

AllTerrainDeveloper:
@claude can you check once more?

PR #683, opened 2026-08-26
Perf: the command palette stops costing every window and every boot

AllTerrainDeveloper:
WordPress enqueues Core's ⌘K command palette on **every** admin page. Behind `wp-commands` / `wp-core-commands` sits the whole Gutenberg chain — react, react-dom, `components.js` (3.7 MB), `block-editor.js` (3.7 MB), `core-data`, `blocks`, `sync`, forty-odd bundles.

OpenStation replaces that palette with its own. The shell already deferred Core's runtime to the first ⌘K; windows were explicitly excluded, and the deferral the shell *did* have was being quietly defeated. This fixes both ends.

## What was actually happening

**Windows paid for a palette they never show.** `src/commands/iframe-bridge.ts` only asks a window for its commands when the palette is *opened* (`onPaletteOpened()` → `os-commands-subscribe`). The harvest was already strictly on demand — but the runtime that answers it was loaded eagerly, in every window, forever.

**Unhooking Core's enqueue alone recovered almost nothing** — 14.49 MB → 14.28 MB. Astra's and WooCommerce's palette scripts each declare `wp-commands`, so `all_deps()` pulled the entire chain straight back in. It has to be a *family* trim, which is the same lesson the admin-bar trim in that file already records.

**The shell's deferral was paying for nothing**, for exactly the same reason. And it did not even work: a contributor that ran at boot registered its commands against a `core/commands` store that does not exist yet, and lost them.

## What this does

1. **Windows drop the palette family**, exempting block-editor screens — there the chain loads for the editor anyway, so the palette rides along for ~150 KB, and those are the screens whose stores hold commands worth harvesting.
2. **Contributors are hoisted to the shell**: dequeued from the boot document and appended to the deferred manifest, so they execute during the replay, *after* the store exists. A plugin's commands reach the palette once, on the shell, and cost no window anything. This fixes the lost-commands bug as well as the cost.
3. **A window may keep a contributor under the plugin's own route**, where it registers screen-specific commands. Matched conservatively, with `openstation_command_palette_contributor_owns_screen` as the escape hatch.
4. **⌘K no longer looks broken on first press** — a placeholder paints immediately in the panel's own position, inline-styled because `ai-assistant.css` is itself still loading during exactly the window it covers.

## Conviction rule (the bug worth reading)

`wp-block-editor` declares `wp-commands` itself — the editor *registers* commands, so the dependency runs opposite to a palette extension's. A plain closure walk convicts the entire block-editor stack and every plugin block script above it. So Core packages are never convicted, and the walk will not travel *through* one to reach a root. Contact Form 7's and MailPoet's block scripts are the real cases this was found on.

## Numbers, honestly

The size of the win is a property of the site, not of the trim. On a **clean install**, where the palette is the sole consumer of that chain, one Settings window goes from **85 requests / 4.06 MB → 43 / 0.83 MB** — 49% fewer requests, 80% fewer bytes.

On the **WooCommerce dev site** this was measured on, the chain has other legitimate consumers (Woo's abilities-api client pulls `wp-core-data` on every screen), so it correctly stays and the trim reclaims only the palette's own bundles: 137 req / 14.49 MB → 127 / 14.14 MB, document 306 → 282 KB, inline JS 202 → 180 KB.

An earlier revision of this branch reported far larger numbers; those came from the over-conviction bug described above and are corrected here.

## Testing

- PHPUnit **2523 tests, 0 failures** (26 in `Tests_OpenStation_ChromelessCommandPalette`)
- Vitest **4998 tests, 0 failures** (new `ai-palette-placeholder.test.ts`, plus manifest-dedupe coverage)
- `npm run lint:php` (phpcs -n), `lint`, `typecheck`, `build` all clean

**Needs a human pass:** open Settings/Tools/Users in windows and confirm they render and behave; press ⌘K and confirm the placeholder appears then swaps for the panel; confirm Astra/Woo commands now show in the palette (they did not before).

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-683-b351eff3a500e24420c9120ad070b1a4f38e446f.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
@claude can you review?

PR #684, opened 2026-08-26
Perf: a window stops building the admin bar it never draws

AllTerrainDeveloper:
Removing the render stopped the markup. It never stopped the work.

`_wp_admin_bar_init()` is hooked on `admin_init`, and `is_admin_bar_showing()` short-circuits to true for any admin request — the same short-circuit this codebase already documents as the reason the `show_admin_bar` filter cannot hide the bar on its own. So every window still instantiated `WP_Admin_Bar` and called `add_menus()`, which fires `admin_bar_menu` and runs **every** registered callback — core's twenty-odd nodes plus every plugin's, each resolving links, counting things, checking capabilities — then dropped the finished object on the floor, because nothing renders it.

The shell draws a real admin bar, once. A window drawing none should pay for none.

**The class is swapped, not the init unhooked.** `remove_action( 'admin_init', '_wp_admin_bar_init' )` would leave `$wp_admin_bar` null, and a plugin touching the global outside the hook — bad practice, entirely real — would fatal on it. Core exposes `wp_admin_bar_class` for precisely this, so a window gets a real `WP_Admin_Bar` subclass that works in every respect (`add_node()`, `get_nodes()`, the global is still an object) except that it never solicits nodes.

Measured interleaved A/B on one Settings window, 12 pairs: **728 ms → 717 ms median, 11 ms / 1.5%**. Modest on this install because it has no expensive masterbar; the saving scales with what the site's bar actually costs, and the handles the chromeless trim already names (`wpcom-notes-admin-bar`, `a8c-faux-inline-help`) are exactly where it is not modest.

---

## Where a window's time actually goes

Profiled while working on this, because it changes what is worth optimising next. **This section is context, not a claim about this PR** — the change above moves server time, not bytes.

### Assets are already free; the document is not

Measured in Chrome against the live shell, opening screens as real windows:

| window | requests | **from cache** | over the wire | transferred |
|---|---:|---:|---:|---:|
| Settings (`options-general.php`) | 132 | **129** | 3 | 283 KB |
| Tools (`tools.php`, never visited) | 101 | **99** | 2 | 163 KB |
| Media Library (`upload.php`) | 250 | 160 | 90 | 2,274 KB |

A window pulls 13–14 MB of assets and **transfers essentially none of it** — the shared admin-asset cache and the SW have it all. What a window actually costs on the wire is **its HTML document**, every time, uncacheable.

### So on a slow connection, per-window cost is document size

Modelled from measured gzip sizes at standard profiles (Slow 3G 400 Kbps, Fast 3G 1.6 Mbps, 4G ~9 Mbps). Steady state = warm shell, the case that happens all day:

| scenario | transferred | Slow 3G | Fast 3G | 4G |
|---|---:|---:|---:|---:|
| **Window open, warm shell** (today) | 68 KB gz | ~1.3 s | ~0.34 s | ~0.06 s |
| First window, cold cache — palette-only site, **before** #682/#683 | 2.87 MB gz | ~57 s | ~14.7 s | ~2.6 s |
| First window, cold cache — palette-only site, **after** | ~0.93 MB gz | **~19 s** | **~4.8 s** | **~0.85 s** |
| First window, cold cache — WooCommerce site, after | 2.79 MB gz | ~56 s | ~14.3 s | ~2.5 s |

The cold-cache rows are where the merged palette work pays: on a site where the palette is the sole consumer of the Gutenberg chain, first load drops by 1.94 MB gz — **3× faster on any connection**. On a WooCommerce site the chain has other legitimate consumers and correctly stays, so the win is small; that asymmetry is documented in `hooks-reference.md`.

### What that leaves

Since only the document transfers per window, document size is now the lever. Of the 282 KB Settings document: **96 KB is one WooCommerce `wcSettings` inline blob, 55 KB is media modal templates (19.7%), 28 KB is admin-menu markup rendered purely to be hidden by CSS.** That is 179 KB of 282 KB, re-sent in every window.

And the ceiling behind all of it: `admin-ajax.php` with a **no-op action** costs 515 ms on this install versus 606 ms for a full Settings window — **~85% of a window is WordPress + plugin bootstrap**, before any page work. No asset trimming reaches that; only not re-requesting the document does, and `isCacheableResponse()` currently rejects `no-store`/`private`, which is every admin response. Proposal for a session-scoped document cache to follow separately.

## Testing

- PHPUnit **2529 tests, 0 failures** — 5 new in `Tests_OpenStation_ChromelessAdminBar`, covering the class swap, that `admin_bar_menu` never fires, that the subclass is still a working `WP_Admin_Bar` (the null-global regression), and the filter opt-out
- `npm run lint:php` (phpcs -n) clean — the subclass lives in its own `class-*.php` per WPCS, required lazily from the filter since `WP_Admin_Bar` only exists inside `_wp_admin_bar_init()`
- `npm run build` clean, no drift

**Worth a human pass:** confirm windows still render with no admin bar and nothing regresses on screens whose plugins add admin-bar nodes.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-684-143fa7e3c2e1808a4ceaa181d0b151b1aecc6a33.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/docs/hooks-reference.md
+++ b/docs/hooks-reference.md
@@ -1746,6 +1746,22 @@ Controls the `Sec-Fetch-*` fallback in chromeless detection: when a request arri
 apply_filters( 'openstation_chromeless_sec_fetch_fallback', bool $allow );
 ```
 
+### `openstation_chromeless_silence_admin_bar` — Experimental
+
+Whether a window skips **building** the admin bar it never draws. Default `true` inside windows.
+
+Suppressing the render stops the markup, not the work. `_wp_admin_bar_init()` is hooked on `admin_init`, `is_admin_bar_showing()` short-circuits to true for any admin request, and so every window still instantiated `WP_Admin_Bar` and called `add_menus()` — which fires `admin_bar_menu` and runs **every** registered callback, core's twenty-odd nodes and every plugin's, each resolving links and checking capabilities — before dropping the finished object on the floor. The shell draws a real bar once; a window drawing none should pay for none.
+
+The class is swapped rather than the init unhooked, deliberately. `remove_action( 'admin_init', '_wp_admin_bar_init' )` would leave `$wp_admin_bar` null, and a plugin that touches the global outside the hook would fatal on it. Core exposes `wp_admin_bar_class` for exactly this, so a window gets a real `WP_Admin_Bar` subclass that works in every respect — `add_node()`, `get_nodes()`, the global is still an object — except that it never solicits nodes.
+
+How much this saves is a property of the site's admin bar: roughly 1.5% of a window's server time on a stock-ish install, and materially more wherever the bar is expensive (a WordPress.com masterbar, a plugin adding counted nodes).
+
+```php
+// Let a window build the bar, for a plugin that relies on
+// `admin_bar_menu` firing for a side effect rather than a node.
+add_filter( 'openstation_chromeless_silence_admin_bar', '__return_false' );
+```
+
 ---
 
 ### `openstation_chromeless_admin_bar_top_values` — Experimental
--- a/includes/core/class-openstation-silent-admin-bar.php
+++ b/includes/core/class-openstation-silent-admin-bar.php
@@ -1,0 +1,38 @@
+<?php
+/**
+ * OpenStation — the admin bar a window builds instead of the real one.
+ *
+ * **This file is required lazily, not at bootstrap.** It extends
+ * `WP_Admin_Bar`, which WordPress only loads inside
+ * `_wp_admin_bar_init()` — requiring this at plugin load would fatal
+ * on a missing parent. The `wp_admin_bar_class` filter runs *after*
+ * that require, which is the one moment the parent is guaranteed
+ * present, so {@see openstation_chromeless_silence_admin_bar()} pulls
+ * this file in there.
+ *
+ * @package OpenStation
+ */
+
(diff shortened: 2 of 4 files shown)


AllTerrainDeveloper:
@claude can you review?

PR #687, opened 2026-08-26
Corkboard: keep a sparse board centred and explained

epeicher:
## Summary

A site with two posts made the Corkboard look broken rather than empty: the two nodes drifted off-centre after Fit, nothing said what a thread was or when one would appear, and the full-canvas loader ran for waits that were not long.

- **Small boards open centred and stay put.** Boards of up to twelve nodes are seeded deterministically around the world origin (a lone node at the centre, a pair side by side, otherwise an even ring) and run to rest before the first paint, so the frame the camera is given is the final layout. Larger boards keep the random spread but are recentred, and nodes joining a live board keep the short warm-start so nothing on screen snaps. Fit maths is shared (`layout.ts`) and a fit requested while the host has no size yet is replayed once it does.
- **The board explains itself.** A "No threads yet" note says a thread joins two posts when one links to the other, and where author, terms, comments, media, revisions and Group by live (hidden while a grouping is active). An empty board says whether the site has nothing to pin yet, the toolbar has filtered everything out, or every type is switched off.
- **The loader is for long waits.** The stage's full-canvas wash paints only for a graph fetch that runs past 400 ms; shorter waits show as the toolbar status line. The window resolves its render promise once it is interactive, so the shell's loader covers just the cold bundle + renderer start-up. Chip counts from `/post-types` are patched in place instead of rebuilding the toolbar under the user.
- **`/nodes` tells an omitted `types` query (every registered type) apart from an explicitly empty one (no types).** The route registers no default for the parameter, because the dispatcher installs a registered default before the callback and `has_param()` would see it; the shell always sends the parameter, so every chip off is an empty board.
- Docs (`docs/hooks-reference.md`) corrected: only internal links form threads; terms, authors and comments are satellites and Group by facets, not edges.

Labels are unchanged: the card treatment the issue proposed was tried and dropped as too crowded on a board with more than a handful of posts.

## Testing

- `npm run build`, `npm run lint`, `npm run typecheck`
- `npm run test:js` — 397 files, 5,041 tests (new: `content-graph-layout`, `content-graph-board-notice`, `content-graph-loading-overlay`, `content-graph-rest`)
- `npm run test:php -- --filter ContentGraph` — 18 tests; the `/nodes` omitted-vs-empty case now goes through `rest_get_server()->dispatch()`
- phpcs on the touched PHP files
- wp-env QA on localhost:8890: stock 1 post + 1 page (pair centred, no drift, Fit stable, note shown, no console errors), then 29 seeded nodes / 26 threads across two authors, three categories, three tags, three years and a private post: chips, Group by, notices, loader timing, and `/nodes` with `types` omitted / empty / listed

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-687-09295dd8fdd6bc51f395c36af3934d78986b958f.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


epeicher:
Follow-up in 33d3bfe2, after review and QA of the first commit:

- **Cards dropped.** The card treatment (title + type/author/month under each node) read as crowded once a board had more than a handful of posts, so labels are back to the original pills and `card.ts`, the `singular` descriptor field and their tests are gone. The issue's other three points (centring, notices, loader timing) stand.
- **`/nodes` regression fixed.** The first commit's `has_param( 'types' )` branch never fired through a real dispatch: the route registered `'default' => ''`, the dispatcher installs registered defaults into the request before the callback, and `has_param()` walks that slot too, so every request that omitted `types` got an empty board. The unit test passed only because it called the callback directly. The default is removed and the test now goes through `rest_get_server()->dispatch()`.
- **Toolbar counts patched in place.** With the render promise resolving before the first fetch, rebuilding the toolbar when `/post-types` answered reset chips, search and Group by under the user's pointer.
- **Settle-before-paint only for boards laid out from scratch**, so toggling a type on no longer snaps the nodes already on screen.
- Notices: an "every type is switched off" state, unknown counts blame the chips rather than the site, the "No threads yet" note hides while a grouping is active, copy no longer mentions cards.
- Loading wash follows `--os-ui-surface` via `color-mix()` so it dims a dark board too; a few comment/docblock fixes.

PR #688, opened 2026-08-26
Perf regression sweep: every finding, fixed

AllTerrainDeveloper:
Works through the **OpenStation · trunk `a4967bab` Perf Regression Sweep**, point by point. The deferral machinery was sound; what broke was everything that assumed the assets and modules it moved were still resident.

---

# Measurements

Same site, same screens, measured three ways. WooCommerce install on WP 7.1 (`:8889`), admin user.

## Cold — nothing cached

`node bin/boot-cost.mjs`: fetches the document and every `<script src>` / `<link rel=stylesheet>` **the server printed**, so it is deterministic — same code in, same numbers out, independent of cache state or how long a tab sat there.

| screen | classic wp-admin | OpenStation window | delta |
|---|---|---|---|
| `options-general.php` | 160 req · 14.81 MB · 2.95 MB gz | **128 req · 14.26 MB · 2.82 MB gz** | −32 req · −0.55 MB |
| `tools.php` | 132 req · 13.50 MB · 2.64 MB gz | **100 req · 12.96 MB · 2.51 MB gz** | −32 req · −0.54 MB |
| `edit.php` | 137 req · 13.61 MB · 2.67 MB gz | **106 req · 13.08 MB · 2.55 MB gz** | −31 req · −0.53 MB |

**On this site the cold win is modest, and that is a property of the site.** WooCommerce's abilities-api client pulls `wp-core-data` on every admin screen, so the Gutenberg chain has consumers besides the palette and correctly stays. Where the palette is that chain's only consumer, the same trim takes all ~10.66 MB raw / 1.94 MB gz of it.

## Warm — caches populated, normal navigation

Browser Resource Timing. `transferSize: 0` means served from cache.

| | requests | from cache | over the wire | transferred | load |
|---|---:|---:|---:|---:|---:|
| classic `options-general.php` | 169 | 165 | 4 | 344 KB | 1133 ms |
| **window `options-general.php`** | 135 | 131 | 4 | **283 KB** | 1188 ms |
| **window `tools.php`** (2nd window, same shell) | 105 | 102 | 3 | **163 KB** | 1067 ms |
| OpenStation shell boot | 212 | 175 | 37 | 967 KB | 2444 ms |

**In the steady state a window transfers only its HTML document.** 131 of 135 requests are cache hits; the 4 that cross the wire are the document and three uncacheable odds. The same is true of classic wp-admin — the shared admin-asset cache is origin-wide, so it serves both — which is why the warm gap is ~18% on bytes and ~20% on requests rather than anything dramatic.

The shell costs 967 KB warm to boot, amortised across every window opened afterwards.

## Server time — unaffected by any of this

Median of 7 authenticated requests per URL:

| screen | classic | window |
|---|---:|---:|
| `options-general.php` | 599 ms | 596 ms |
| `tools.php` | 564 ms | 628 ms |
| `edit.php` | 628 ms | 626 ms |

**Neither is faster on the server, and no asset work can change that.** An `admin-ajax.php` call with a no-op action — full WordPress + plugin bootstrap, no admin page render — costs 515 ms on this install against 606 ms for a whole Settings window. **~85% of what a window costs is booting WordPress.** The remaining lever is not re-requesting the document at all, which is what hover prewarming and the speculative-document store exist for.

### Method notes

- The two instruments count differently: `boot-cost` counts assets the **server printed**; the browser counts everything **fetched**, including images and fonts. Compare within a row, not across tables.
- An earlier draft of this table reported classic wp-admin re-transferring 14 MB when warm. That was wrong: `location.reload()` forces revalidation, so the measurement was a forced reload, not a warm navigation. Re-measured as a normal navigation, classic caches properly. The figures above are all normal navigations.

---

# The findings

## High (5/5)

**Notes: drops rejected, cross-user sync never ran.** Four registries — `canvas-payloads`, `recycle-bin-payloads`, `tile-payloads`, `heartbeat` — held module-level state while compiled into both the shell and `notes.js`. Now `createSharedStore`, plus a **built-bundle boundary test**: vitest imports both sides of a seam into one module graph and structurally cannot see this class of bug, so the test reads the built bundles instead.

**Customizer Widgets panel empty.** Two mechanisms. Core packages were identified by **src path**, but the Gutenberg plugin re-registers the whole `wp-*` family from its own directory (verified in `build/scripts.php`) — so the guard answered false for every package on a Gutenberg site and the walk convicted the entire editor stack. Packages are now identified by name. And nothing stopped the trim removing a handle a **surviving** handle still depended on; `openstation_protect_survivor_dependencies()` now subtracts the closure of everything still printing.

**Third-party bundles lost the `wp.*` globals.** `openstation_resolve_script_dependencies()` resolves a lazy handle's closure and the loader replays it before the bundle, plus `docs/migration-wp-package-globals.md` and a starter widget that no longer teaches the assumption that breaks.

**Games launched outside the hub painted unstyled.** `launchGame()` is compiled into the shell bundle too; three documented paths reach it with no hub window.

**Speculative documents.** Explicit reloads bypass the store; `*-new.php` refused by filename; held pages and the restore list cleared at `wp-login.php`; a dwell required before speculating, and the active tab skipped.

## Medium / Low (10/10)

Palette retry · notes in-shell drag · `svg-splines` (metadata stub, so `listWindowLinkRenderers()` keeps its documented promise) · dock-constellation keyboard · theme teardown with retry · note-less first public note · file-drop replay guards · `ownerHandle` · prewarm toggle reaching the running worker · phpcs excludes `.claude/`

## Found on the way (3/3)

**Preferences double-save** — fixing the overlap exposed a second cause: the confirmed baseline was cloned from the *live* state object, absorbing fields the payload never sent. · **WP Explorer 404** — entity `enabled` flag *and* read-only probes silenced. · **AI palette fade race.**

## Review candidates (7/7)

| | verdict |
|---|---|
| hover prewarm, no acting-URL guard | ✅ confirmed — predicate now shared with the SW |
| discarded prewarm skips cleanup | ✅ confirmed — leaked a connection-bridge entry per hover |
| prewarm/open race → two windows, one id | ✅ confirmed |
| bridge runs after footer scripts | ❌ **refuted** — `admin_footer` is core line 41 vs 47 |
| WP 6.3–6.8 hoist loses commands | ✅ confirmed |
| folder preview unstyled | ✅ confirmed |
| stale `architecture.md` | ✅ rewritten |

## Tooling (2/2)

phpcs no longer scans `.claude/worktrees/` · `updateOsSettings()` accepts the two PWA flags, with the same worker notification the toggle sends.

---

**One notable non-obvious fix:** the served `sw.js` carried per-user flags, so its bytes differed between an anonymous and a logged-in request — any in-scope logged-out navigation installed a "new" worker and tripped the shell's `controllerchange` hard reload. The preamble now carries `pluginUrl` only; the shell posts the flags at boot, computed server-side so the `openstation_pwa_admin_asset_cache` filter's site-wide veto still wins.

## Testing

PHPUnit **2542**, Vitest **5045**, phpcs, lint, typecheck, build — all clean. CI green.

**Not verified by me:** browser behaviour of the fixes themselves. The three worth a human pass are the Customizer Widgets panel with Gutenberg active, a note dropped on the wallpaper, and a game launched from the challenge toast.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-688-a16c6c739dbac85febd41c7c6b0e8ed778055963.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
@claude can you review?

AllTerrainDeveloper:
@clade can you review?

AllTerrainDeveloper:
@claude can you review?

epeicher (approved):
Thanks @AllTerrainDeveloper for fixing all the findings on the regression testing! I tested it manually and also used some Claude agents, and all findings have been resolved. I will suggest some follow-ups to avoid loading Gutenberg bundles when the Gutenberg plugin is activated, but that can be tackled as a follow-up. This LGTM!

P.S. I see that the latest Claude review has found something, but I also think that can be addressed in a follow-up.

AllTerrainDeveloper:
> Thanks @AllTerrainDeveloper for fixing all the findings on the regression testing! I tested it manually and also used some Claude agents, and all findings have been resolved. I will suggest some follow-ups to avoid loading Gutenberg bundles when the Gutenberg plugin is activated, but that can be tackled as a follow-up. This LGTM!
> 
> P.S. I see that the latest Claude review has found something, but I also think that can be addressed in a follow-up.

Sorry I didn't realise...

epeicher:
For reference, this is the report with all issues fixed https://claude.ai/code/artifact/fc34eb6c-bfb3-4460-984c-a71f768d2e37

PR #689, opened 2026-08-27
Show a loader in place for every lazy-loaded surface

AllTerrainDeveloper:
Windows already had a loading affordance — `src/window/loading.ts` paints an overlay between `markWindowContentLoading()` and `markWindowContentReady()`, with a 120 ms delay so a fast load leaves no trace. Three lazy surfaces never announced themselves to it and showed **nothing at all**.

## Games opened no window until the bundle arrived

`launchGame()` awaited `ensureGameRender()` *before* `registerWindow()`. A click fetched the game, its engine and sometimes a dictionary asset with no window on screen to hang a spinner on, then a window appeared. The worst of the three, because there was no surface to put a loader on.

The load moves inside the render callback, where `Window.hydrateNative()` already holds the overlay for as long as a render promise is pending — the same shape native windows use for their own lazy bundles.

That needed the window's size a round trip earlier than the JS def that carries it, so **`openstation_register_game()` takes a `window` argument**: any subset of `width` / `height` / `minWidth` / `minHeight`, clamped rather than rejected, because an unopenable window is a worse answer than an oddly-sized one. The def still wins once it lands, so a game declaring only in JS keeps working — it opens at the framework default the first time and at its own size after. Both built-in games declare both now.

## Preferences sat empty while its panel bundle loaded

…and on a rejection sat empty **permanently**, with a console line as the only evidence. It shows a loader now, and on failure a message with a Retry.

Making that Retry mean anything meant fixing the loader: `loadOsSettingsPanelBundle` memoized the rejected promise *and* left the dead `<script>` in the DOM, so a retry would have re-awaited the rejection forever and — via the `existing` branch — attached listeners to a script whose `error` had already fired. Both are cleared before it rejects.

## A widget card was empty while its script loaded

A server-registered widget's `mount` awaits its own bundle; the layer only wired `result.then( onResolve )`.

## The shared affordance

`src/ui/inline-loader.ts` — same 120 ms delay, `role="status"` (assertive on failure), and `<os-spinner>` when `customElements.get()` says the tag will upgrade, inline markup when it will not.

That fallback is not defensive padding. The component ships in the lazy `shell-overlays` bundle, and a widget can mount before it lands, so an `<os-spinner>` there would be an inert unknown element for exactly the window this module exists to cover. Awaiting another bundle in order to say "please wait" is self-defeating — `ai-assistant/loading-placeholder.ts` already records the lesson.

**No new bundle and no new request.** `shell-overlays` is preloaded after first paint anyway for toasts, confirm dialogs and window chrome. The 335 K `os-components` kit is only ever fetched by `wp.os.loadComponents()` for third-party plugins, and was never on the table.

## Two things found while testing the above

**Enabling Games needed an F5.** The `games` option is read on `plugins_loaded`, so the request that *writes* it already decided games were off and cannot report the window, icon and game list it would now register. Only a later request can. The developer-mode toggle already spent a `wp.os.refreshMenu()` for exactly this reason; the Extended Options save fired `EXTENDED_OPTIONS_CHANGED` and stopped, and the only listener for that is the agents renderer. Both now go through `spendMenuRefresh()` — on success only, and once for coalesced flips, since the trailing save owns it.

**`<os-disclosure>`** — a new component, since nothing in the kit collapsed. Deliberately not `<details>`/`<summary>`: its open state cannot be styled or animated consistently across engines, and a `<summary>` slotted through a shadow root loses the behaviour that made it worth using. What the native element really buys is semantics, so those are reproduced in full — a real `<button>` with `aria-expanded` / `aria-controls`, and a body carrying the actual `hidden` attribute, so a collapsed panel is out of the tab order rather than merely invisible.

The Components tab's missing-import demo now folds away, giving the two browser panes back the ~200px it took on every visit. Its live copies moved *outside* the disclosure: a closed body is `display:none`, which that markup has always avoided on purpose — a demo that only fires when someone expands the section is not a demo of what happens on page load.

## Tests

Every fix is pinned by a test **verified to fail without it**:

| guard | reverted behaviour it catches |
|---|---|
| `games-launch.test.ts` | `expected [ 'script-fetched', … ] to deeply equal [ 'window-opened', 'script-fetched' ]` |
| `extended-options-save.test.ts` | two failures — the refresh, and the coalescing |
| `inline-loader.test.ts` | delay, failure state, retry, spinner fallback, reduced motion |
| `os-disclosure.test.ts` | hidden body, real button, aria wiring, per-instance ids |
| `gamesRegistry.php` | `window` sanitization and its trip to the payload |

## Gates

`tsc` 0 · `eslint` 0 · vitest **5098/5098** · build 0 · `phpcs -n` 0 · PHPUnit **2554, 0 failures**. Desktop bundle 561 K → 563 K.

## Docs

`hooks-reference.md` (the new `window` arg and why it isn't redundant), `javascript-reference.md` (the open-before-fetch ordering and what it means for a `render` that throws), `docs/examples/register-game.md`, `components-reference.md`, and a line in `AGENTS.md` on the question that would have caught the F5 bug at review time: *does this toggle change what the server registers?*

## Not verified by me

Browser behaviour. Worth a human pass: a game launched from the hub (window should appear instantly, at the right size, first time), Preferences while `os-settings-panel` loads, a widget card on a cold boot — the one path where the spinner fallback can actually fire — and the Components tab with developer mode on.


AllTerrainDeveloper:
@claude can you review?

PR #690, opened 2026-08-27
Dynamic dock: auto-hide behavior setting, and one shared work area so windows, widgets, icons and graphs stay clear of the dock

epeicher:
## What it does

Adds **OpenStation Preferences → Appearance → Desktop layout → Dock behavior**: `Static` (default, always on screen) or **`Dynamic`**, the auto-hide dock. The rail folds into a thin indicator line at its edge (iOS home-indicator style) and morphs back into the full dock, through the View Transitions API, when the pointer reaches that edge of the screen, while it or one of its flyouts is hovered, or when a tile has keyboard focus. With it on, windows, icons and widgets get the whole desktop and the dock rides over them only when summoned. The control sits in the Desktop layout section under Dock size, since every layout's rails can park.

In the **Split** layout the sidebar answers for itself: a **Sidebar behavior** control inside the Split card, independent of the bottom dock's, so a folded sidebar over a static dock (or the reverse) is a valid desk. Both layout cards now carry their control at all times, Placement in Unified and Sidebar behavior in Split, so picking a layout no longer grows one card and shifts the other; and a rail set to Dynamic is drawn in the card's preview as the thin line it folds into, with a slow breathe.

Underneath, windows, widgets, desktop icons, sticky notes and the graph windows no longer guess at where the dock is. One measured **work area** (the desktop area minus the band the dock pill floats over; nothing at all for a dynamic rail) drives every default placement, so nothing opens, restores or packs under the dock. Explicit actions keep the whole desktop: maximize and snap fill it, and a window dragged under the dock stays where the user put it.

## Rationale

There was no single answer to "where may content go". `.os-area` reserved the dock with a hardcoded `padding-bottom: 80px` whatever edge the dock was on, the window manager halved `parent.clientWidth`, widgets clamped against the whole area, the Corkboard fitted a graph into its whole host. The visible result: the dock covering the last row of a window, icon columns ending under the pill, and Fit framing nodes into space the user could not reach. A native window registered at 1080×720 (the Corkboard) opened straight under the dock on a laptop.

## Implementation

`src/work-area/compute.ts` is the pure geometry: only chrome that *overlaps* the area claims a band (a side rail is a flex sibling, so it claims nothing), and a rail claims the edge it is nearest to (a one-tile pill is taller than wide and still means "bottom"). `src/work-area/index.ts` measures `#os-area` and every `.os-dock`, with a `ResizeObserver` per rail, and publishes on change only:

```ts
// desktop-area-local; the space a window's x / y resolve in
const { rect, insets } = wp.os.workArea.get();
// how far an element hangs outside it, in its own px — what the graphs subtract before centring
const inset = wp.os.workArea.insetsOf( host );
```

The same numbers land on `#os-shell` as `--os-work-area-inset-{top,right,bottom,left}` so `.os-area`, `.os-icons` and `.os-widgets` reserve the band from CSS. State lives in `createSharedStore( 'os/work-area' )` because the Corkboard and posts-window bundles are lazy and would otherwise read an empty module copy. The snapshot freezes across the Exposé cycle (the rails animate to width 0 and back; measured live that would walk the inset 84 → 0 → 84 one frame at a time), and the viewport origin is rebuilt from the offset chain so the desktop-switch slide never leaks a transformed rect.

`WindowManager.open()` fits a window nobody positioned into the work area (a registered size taller than the reachable height shrinks, down to its minimum; the cascade origin is pulled up), and the `os.window.geometry` filter context gains `workArea`.

A parked dynamic rail is the same element folded into a 180×5px indicator line in the accent at 72% alpha (`--os-dock-indicator-length` / `-thickness` / `-bg`, theme-tunable); nothing is transformed, since the dock tooltip is `position: fixed` and a transformed ancestor would become its containing block. The behavior is worn **per rail** as `data-os-dock-behavior` (`dockBehavior` on `#os-dock`, `sideDockBehavior` on `#os-side-dock`) rather than as a body class, because Split's two rails answer independently: PHP stamps the dock in the shell template so the first paint is already parked, the settings apply pass re-stamps both, and `src/dock-behavior.ts` re-stamps whenever the layout dispatcher rebuilds a rail (the Split sidebar is a fresh element every time). That module alone decides parked vs. revealed — a full-width edge zone (the admin bar's pseudo-element trick only covers the bar's own width; a centred pill needs the whole edge), a keep-out band reaching twice the rail's height from the screen edge so the strip a window's bottom actions live in never folds it by accident, its flyouts, keyboard focus, and the pointer leaving the browser window freezes the state rather than folding the dock — and runs each flip inside `document.startViewTransition()` with a transient `view-transition-name`, so the browser morphs the line into the pill; the root is opted out of the snapshot so the desktop stays live during the 260ms morph, one morph at a time per rail, plain CSS transitions where the API is missing and no motion under `prefers-reduced-motion`. An `openstation_dock_behavior` filter mirrors `openstation_admin_bar_mode` for the dock. In the Preferences cards, a Dynamic rail's preview bar shrinks by transform (its geometry stays an inline style from the preview table) and breathes on opacity only, both off under reduced motion.

Public surface is `Experimental`: `wp.os.workArea`, `os.work-area.changed` / `os-work-area-changed`, the `--os-work-area-*` properties, the `dockBehavior` / `sideDockBehavior` settings and the PHP filter. Docs in `javascript-reference.md`, `hooks-reference.md`, `architecture.md`, `examples/work-area.md`; the rule is written up in AGENTS.md.

## Testing instructions

```bash
npm run test:js          # work-area, dock-behavior, os-settings-dock-behavior, window-open-default-fit
npm run test:php -- --filter 'Tests_OpenStation_OsSettings|Tests_OpenStation_Render'
```

In the shell:

1. Open **Corkboard** from the desktop icon: it opens above the dock. Maximize it: it covers the dock. Drag a window down: it may go under the dock.
2. Drag the Corkboard so its lower part hangs under the dock and press **Fit**: the graph centres in the part above the dock.
3. **Preferences → Appearance → Desktop layout → Dock behavior → Dynamic**: the dock folds into the line; move the pointer to the bottom edge of the screen and it morphs back; hover a tile's flyout and it stays; move away and it folds. Icons and widgets extend to the bottom. Reload: no flash of the dock. The preview bars in both cards thin out.
4. Pick **Split**, then **Sidebar behavior → Dynamic** in its card: the sidebar folds into a line on the left edge while the bottom dock keeps whatever Dock behavior says, and vice versa. Switching between Unified and Split moves nothing in the cards.
5. Back to **Static**: the reserved band returns. `wp.os.workArea.get().insets.bottom` reads the pill's height plus its floor margin and the 8px gap (`84` with the Compact dock size; more for Default and Large), and `0` with a side dock or a dynamic dock.


AllTerrainDeveloper:
How do you feel if we hide it with fancy animation into a line iOS style? 

Actually, using the Transitions API?

epeicher:
> How do you feel if we hide it with fancy animation into a line iOS style?
> 
> Actually, using the Transitions API?

Great idea! This is how it looks like 

https://github.com/user-attachments/assets/c3786f8f-95db-4727-82f1-33f439d76481

AllTerrainDeveloper:
@epeicher that's exactly what I meant!

But I have a suggestion. let's make the area where the mouse makes the dock disappear double or trile the size of the dock height. So the user can navigate freely without fear of hidding it accidentally. Also, let's make the lines wider too

AllTerrainDeveloper:
Also as a follow up PR, can we apply the same concept to the top master bar? And add new tokes for the theme, like minimun width, colour (default to accent's colour) with alpha channel (transparency).

AllTerrainDeveloper:
Another improvement, is even possible to avoid hiding it when the mouse leaves the window interelly? eg:

https://github.com/user-attachments/assets/f317e8f4-f27a-42c7-9dff-c86f1d3b00c3

Sorry for being a pain 😆

epeicher:
Thanks @AllTerrainDeveloper for the feedback! I have implemented all the suggestions, please see a video with them 

https://github.com/user-attachments/assets/4df0bdf2-398c-44a4-96be-1092e47c6c4f


> Sorry for being a pain 😆

Quite the opposite; feedback is a gift, and I'm all in to polish all the details, so thanks for being a pain! 😄

epeicher:
After some latest improvements suggested by @AllTerrainDeveloper, this is the final version of the animations and settings. Now the sidebar and dock can have independent behaviors, or they can be unified:

https://github.com/user-attachments/assets/1b5cc005-d010-4dca-863b-902f8f186975

PR #691, opened 2026-08-27
Trash: Hide the list chrome while the bin is empty

mmtr:
## Proposed Changes

- An empty Trash shows only its empty state. The toolbar (type filters, search, Refresh, Empty Trash) and the table stay hidden until the bin holds something.
- The Title and By columns drop their per-column "Filter…" inputs. The toolbar's search box is the only one now.
- A filter or search that matches nothing keeps the toolbar and reads "No items match the current filter or search." instead of claiming the Trash is empty.

Before | After
--- | ---
Screenshot 2026-08-27 at 13 37 13 | Screenshot 2026-08-27 at 13 37 24
Screenshot 2026-08-27 at 13 37 52 | Screenshot 2026-08-27 at 13 38 00

Filters, sorting and a destructive Empty Trash button all acted on a list that wasn't there, and the loudest control on screen did nothing.

## Testing Instructions

1. Start with an empty Trash, then open Trash from the dock. Make sure the window shows only the bin glyph and "The Trash is empty.", with no toolbar, no column headers and no Empty Trash button.
2. Leave the window open and trash a few posts from Posts. Make sure the toolbar and the table appear without a reload.
3. Type something in the search box that matches nothing. Make sure the toolbar stays put and the table reads "No items match the current filter or search.", then clear the search and make sure the rows come back.
4. Pick the Comments filter tab on a bin with no trashed comments. Same thing: toolbar stays, table says nothing matched.
5. Look at the Title and By column headers. Make sure neither has its own "Filter…" input, and that clicking either header still sorts.
6. Tick a row. Make sure Restore, Pin to desktop and Delete forever appear, and that Restore still works.
7. Click Empty Trash and confirm. Make sure the window lands back on the empty state with no toolbar.
8. With the Trash window open and empty, drag a post from WP Explorer onto it. Make sure the drop still trashes the item and the chrome comes back.
9. Reopen the Trash on an empty bin a second time. Make sure it opens straight into the empty state, with no flash of skeleton rows or toolbar first.

### DIFF
--- a/assets/css/recycle-bin.css
+++ b/assets/css/recycle-bin.css
@@ -38,11 +38,14 @@
 /*
  * Honour the `hidden` attribute. The browser's UA stylesheet
  * sets `[hidden] { display: none }`, but our explicit
- * `display: flex` above wins on specificity. This rule restores
- * the expected behaviour so `bulk.hidden = true` from JS
- * actually hides the bulk action group when nothing is selected.
+ * `display: flex` above wins on specificity. These rules restore
+ * the expected behaviour for every group the JS hides: the bulk
+ * actions (nothing selected), and the whole toolbar plus the
+ * standalone empty state (nothing in the bin).
  */
-.os-recycle-bin__toolbar-right[hidden] {
+.os-recycle-bin__toolbar[hidden],
+.os-recycle-bin__toolbar-right[hidden],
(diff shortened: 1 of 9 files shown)


AllTerrainDeveloper:
Lovely Miguel

PR #692, opened 2026-08-27
Perf: the shell stops bouncing /wp-admin/ through the portal

AllTerrainDeveloper:
Entering the admin costs two 302s before the document that renders is even requested — **77.4 ms** of full WordPress bootstrap, and the largest untouched item in the perf report.

Both hops are a round trip to nowhere.

## The round trip hands back what it was given

Typing `/wp-admin/edit.php?post_type=page`:

| Hop | Where | What it decides |
|---|---|---|
| 1 | `admin_init` → `openstation_redirect_plain_admin_to_portal()` | 302 to `/openstation/?target=%2Fwp-admin%2Fedit.php%3Fpost_type%3Dpage` |
| 2 | `parse_request` → `openstation_handle_portal_request()` | validates `target` against the wp-admin allowlist, gets `edit.php` back, reattaches `post_type=page`, 302 to **the same URL** + `desktop_mode_portal=1&desktop_mode_portal_intent=1` |
| 3 | renders | `currentPage` = flags stripped = the URL you typed |

And the flag pair it pays for is inert: it has exactly two consumers — `src/boot/auto-open.ts:62` and `src/desktop.ts:3343` — and **both test `fromPortal && ! fromPortalIntent`**. So `true`/`true` is semantically identical to no flags at all; cases 1 and 2 of the auto-open truth table both mean "open `currentPage`", and `tests/vitest/boot-auto-open.test.ts` already pins that.

The three reasons in the function's own docblock have all lapsed:

- *"land in the shell, not vanilla admin"* — the shell enqueues on any admin page where `openstation_is_enabled()`; no flag gates it.
- *"unifies the address bar"* — abandoned deliberately. The `history.replaceState( config.portalUrl )` was removed precisely **because** this round trip made every reload flash.
- *"honors the saved session's focused window"* — never happens on this path: `target` is always set, so `has_intent = true` and the session is never consulted.

## The change

`openstation_portal_forward_is_redundant()` answers the portal's question locally instead of over two HTTP hops. Conservative by construction — every "don't know" answers false, so the forward survives wherever the destination might genuinely differ:

1. the path must resolve through the portal's own wp-admin allowlist;
2. the resolved filename must be what `$pagenow` reports — a disagreement means a rewrite is in play and we can't claim to know what renders here;
3. the query must carry none of the four keys the portal strips (`openstation_chromeless`, both portal flags, `target`).

Multisite `network/` paths fail (1) and keep forwarding, so their current landing behavior is unchanged — now pinned by a test rather than incidental. Note that behavior is arguably wrong (you ask for `network/sites.php`, you land on the Dashboard), but fixing it is a separate call and out of scope here.

New `openstation_skip_redundant_portal_forward` filter (default `true`) is the escape hatch for a plugin that hooks `openstation_handle_portal_request` for side effects. It is applied *after* `openstation_admin_redirect_to_portal`, so that filter still fires on every candidate request and its contract is untouched.

## Result

Typing `/wp-admin/` renders in place: **zero redirects**, and no portal flags left in the address bar.

`/openstation/` is unchanged — one hop, inherent (a front-end URL can't render wp-admin in-process, since `WP_ADMIN` must be defined before `wp-settings.php` runs and plugins load inside it).

## Docs

- `docs/hooks-reference.md` — new filter entry next to `openstation_admin_redirect_to_portal`.
- `docs/architecture.md` — browser-flow step 1 rewritten; the forward is now the exception, not the rule.

## Testing

- `npm run test:php` — 2565 tests, 20228 assertions, 0 failures (portal class 46/46).
- `npm run lint:php` — 228 files, 0 errors.
- `php -l` on both changed PHP files; `npm run build` produces no diff.

New coverage: the skip on an ordinary admin URL and on `/wp-admin/`, the surviving forward for network-admin paths and for a query the portal would rewrite, the filter override, and five unit cases on the predicate. `test_admin_redirect_preserves_percent_encoded_slashes` was repointed at a network-admin activate link — the single-site URL it used no longer reaches that code path, and the encoded slash it guards is just as present on the path that does.

Verified manually on a local install: `/wp-admin/` and deep links land with no 302, session restore and deep-link-over-session both behave, `/openstation/` and detach-to-new-tab unchanged.


AllTerrainDeveloper:
@claude any regression you might see?

PR #695, opened 2026-08-27
Fix: Sync active Post/Page tab when navigating to Add Post from the Dock

### LINKED ISSUE #693: Add Post / Add Page tabs fail to highlight when opened from the Dock when the window is already active
When the **All Posts** or **All Pages** window is already open and active, clicking **Add Post** or **Add Page** from the Dock's hover menu correctly navigates the window to the editor. However, the tab strip remains highlighted on **All Posts** or **All Pages** instead of switching to the corresponding **Add Post** or **Add Page** tab.

https://github.com/user-attachments/assets/e95e1519-894d-4880-905c-47f13a93ad74

## Steps to Reproduce

1. Open the **All Posts** or **All Pages** window.
2. Keep the window active.
3. Open the Dock's hover menu.
4. Click **Add Post** or **Add Page**.
5. Observe that the window navigates to the editor.
6. Observe that the tab strip still highlights **All Posts** or **All Pages**.

## Expected Behavior

The active tab should update to **Add Post** or **Add Page** when the Dock action navigates to the corresponding editor.

## Actual Behavior

The editor opens correctly, but the tab strip remains highlighted on the previously active tab.

KarunyaChavan:
This PR fixes the issue where the active tab inside a window remains stuck on the previous tab (e.g., "All Posts") when opening a new Post or Page from the Dock menu.

|                                                   Before                                                  |                                                   After                                                   |
| :-------------------------------------------------------------------------------------------------------: | :-------------------------------------------------------------------------------------------------------: |
|  |  |

Closes #693 

### 2) What Changed?
* **`src/window/index.ts`**: Added `syncActiveTab(this, url)` inside `Window.navigateTo()` to optimistically set the active tab before the iframe navigation starts.
* **`src/utils.ts`**: Exported `PAGE_IDENTITY_PARAMS` to share URL identity logic across modules.
* **`src/window/tabs.ts`**: Refactored `findPageOwnerTab` to skip identity parameters that are already handled by `pageIdentityKey`, removing duplicate fallback logic for `post_type`.
* **`src/window/tabs.test.ts`**: Removed invalid test assertions that incorrectly mapped the generic `post.php` URL to the "Add Post" tab.

### 3) How it Fixed?
* When `post-new.php` is opened, the WordPress block editor replaces the URL with `post.php?post=X&action=edit` before the iframe's `load` event fires.
* Since OpenStation previously synced tabs only on `load`, it only saw `post.php` (an off-menu URL) and fell back to keeping the current tab ("All Posts") active.
* By calling `syncActiveTab` optimistically during `navigateTo`, the "Add Post" tab is highlighted immediately using the original `post-new.php` URL.
* When the `load` event fires later with `post.php`, the system falls back to keeping the currently highlighted tab ("Add Post"), resulting in the correct behavior.
* The refactor in `tabs.ts` removes duplicate parameter checks and ensures the matching logic works reliably for both standard posts and Custom Post Types (CPTs).


AllTerrainDeveloper (approved):
Thank you @KarunyaChavan!

PR #700, opened 2026-08-28
Shell screen: boot the desktop from its own admin page instead of riding on the Dashboard

epeicher:
## What it does

The desktop shell boots from its own admin screen, `admin.php?page=openstation`, instead of being painted over the Dashboard (or whichever admin page the portal forwarded to). No other screen's scripts, render callback or hidden HTML reach the shell document any more.

Measured on the QA instance with Gutenberg 23.6 active, `SCRIPT_DEBUG` off, trunk vs this branch, `node bin/boot-cost.mjs --path /wp-admin/`:

| | requests | raw | gzipped |
|---|---:|---:|---:|
| trunk (shell on `index.php`) | 134 | 7.34 MB | 2.20 MB |
| this branch (shell screen) | 47 | 1.15 MB | 0.36 MB |

The 89 files that left are Gutenberg's editor chain (`editor`, `block-editor`, `components`, `media-utils`, react-dom, ...) and the Dashboard-only Core scripts (`updates`, `moment`, `dashboard`). `wp.data` / `wp.apiFetch` / `wp.editor` are no longer in the shell realm until ⌘K loads the palette chain.

## Rationale

#688 made the palette trim respect the dependency graph. That was right for windows, but it exposed what the shell had been getting for free: Gutenberg's experimental Dashboard page enqueues its loader on `index.php`, the loader depends on `wp-editor`, and the shell document *was* `index.php`, so the whole closure printed in a realm that never renders it (162 requests / 20 MB on the plugin fixture). Gutenberg is the loudest case, not the problem: because the portal forwards to the last-focused window's URL, the shell page could just as well be `edit.php` or `post.php`, paying that screen's queue and server render and then loading the same screen again inside a window. Every deferral so far has been fighting the enqueue list of a screen the shell never shows; the durable fix is to stop borrowing a screen.

An admin page rather than a standalone document served from `parse_request`: `is_admin()` must be true and `admin_menu` / `admin_enqueue_scripts` must fire, because those are the documented contract behind every `openstation_register_*` call and every plugin that gates registration on `is_admin()` at load time.

## Implementation

**`includes/shell-screen.php`** registers the page under an empty parent (no menu entry, capability `read`, screen id `admin_page_openstation`) and owns:

- `openstation_is_shell_request()`, the one predicate for "this request paints the shell": the shell screen for an enabled user, or a solo boot. Every hook that used to spell out "enabled, not chromeless, not classic" (shell markup, body class, assets, palette deferral and hoist, preload hints, native-window templates and scripts, PWA head tags, desktop themes, Divi compat) reads it now.
- `openstation_shell_boot_target()`: `currentPage` comes from the validated `target` arg (same sanitiser as the portal, refuses the screen itself), else the session's focused window, the default window, the Dashboard. `fromPortal` is true by construction; `fromPortalIntent` is the `intent` arg. The auto-open matrix in `src/boot/auto-open.ts` is unchanged.
- `openstation_shell_boot_target_meta()`: the entry window takes its title and icon from the dock entry for the page, since `$title` on the screen is "OpenStation" and `$parent_file` is empty.
- `openstation_shell_dequeue_handles` (Experimental): a site can name globally-enqueued handles it never wants on the shell. Reuses `openstation_protect_survivor_dependencies()` from #688; a handle a survivor still depends on is refused with `_doing_it_wrong()`.
- An `admin_title` filter so the tab reads "OpenStation ‹ Site" rather than "‹ Site".

**Routing** (`includes/portal.php`):

```
/openstation/                         → admin.php?page=openstation            (screen resolves the entry)
/openstation/?target=/wp-admin/x.php  → admin.php?page=openstation&target=%2Fwp-admin%2Fx.php&intent=1
/wp-admin/edit.php (plain GET)        → straight to the screen, one hop, when the portal would only hand the URL back
/wp-admin/network/sites.php           → through /openstation/?target=… as before (allowlist miss)
index.php?desktop_mode_portal=1       → alias: screen with that URL as target (+intent when the intent flag was set)
```

The frozen `desktop_mode_portal` / `desktop_mode_portal_intent` values stay frozen; they are aliases now (AGENTS.md updated). The alias route runs before `openstation_admin_redirect_to_portal`, since a URL naming the desktop is not a plain admin page. The screen itself, chromeless loads, solo boots and classic-flagged requests are exempt from the interceptor. The toggle-on redirect and the PWA `start_url` point at the screen.

**JS**: `src/shell-url.ts` exports `isShellDocumentUrl()`. The iframe URL builder (`withChromelessParam`, which every iframe src, tab navigation and in-place navigation passes through), session restore, the boot entry window, hover prewarm and the service worker's `isSpeculatableDocument` all refuse the shell URL, so the desktop can never open inside a window.

**Judgment calls worth a look**

- `openstation_admin_redirect_to_portal` returning `false` now leaves plain admin pages as classic admin (the shell cannot render there any more); the desktop lives at `/openstation/` and its screen. Documented in the migration note.
- Plain admin GETs skip the portal hop. The portal's auto-enable is moot on that path (the interceptor only runs for enabled users); `openstation_skip_redundant_portal_forward` (return `false`) still forces the hop.
- No network-admin twin screen: network URLs already funnelled to the site-admin shell.
- Windows are untouched. The Dashboard *window* on a Gutenberg site still carries the editor chain (#688's survivor rule); that is now the entry window's cost on a fresh session rather than the shell's, and out of scope here.

Docs: `docs/migration-shell-screen.md` (new), `architecture.md`, `hooks-reference.md`, `pwa.md`, `DEVELOPMENT.md`, README index.

## Testing instructions

```bash
npm run build && npm run lint && npm run typecheck && npm run test:js   # 415 files / 5219 tests
npm run test:php                                                        # 2633 tests; new Tests_OpenStation_ShellScreen (26)
npm run lint:php
```

On a wp-env instance with the Gutenberg plugin active (`bin/setup-wp-env.sh` does that), `wp config set SCRIPT_DEBUG false --raw`, then:

```bash
git switch trunk && npm run perf:boot-cost -- --label trunk --out /tmp/trunk.json
git switch perf/dedicated-shell-screen && npm run perf:boot-cost -- --label branch --out /tmp/branch.json
npm run perf:boot-cost -- --diff /tmp/trunk.json /tmp/branch.json
```

The final URL printed for `/wp-admin/` should be `admin.php?page=openstation&target=%2Fwp-admin%2Findex.php&intent=1`; `--path /openstation/` lands on the bare screen, `--path '/wp-admin/index.php?desktop_mode_portal=1'` on the screen with that target.

In the browser: log in, visit `/wp-admin/`, confirm the desktop boots on the new URL with the Dashboard window opened and titled "Dashboard", `wp.data` undefined until ⌘K, a bare `admin.php?page=openstation` reload restores the session without a duplicate window, `?openstation_solo=<id>` still renders in place, and the toggle in classic admin lands on the screen.

Re-verified against the perf regression sweep report on this branch: palette deferral, native window styles, session restore, live refresh, desktop-theme first paint, hover prewarm and solo mode all behave as before; no console errors.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-700-05a5e3659164b4f25e3cd6c3dcd73bc82d1e9ee9.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


epeicher:
@claude review this

PR #704, opened 2026-08-28
Multisite: support the network admin, and scope desktops per site

mmtr:
## What it does

Makes the shell work on a multisite network.

- **The network admin gets its own shell screen**, at `wp-admin/network/admin.php?page=openstation`. Its dock is the network menu, and its screens open in windows there like a site's do.

Screenshot 2026-08-28 at 17 52 06

- **A Network Admin dock tile** on each site, mirroring the admin bar's Network Admin node with core's own capability gates. It paints with the admin menus rather than among the app launchers, second in the run, next to the Dashboard.

Screenshot 2026-08-28 at 17 52 26

- **Links that leave the current admin open a browser tab.** Sites > All Sites > any site's Dashboard is the common one.
- **Each site keeps its own desktop.** The session was one blob shared by the whole network.

None of it worked before. Switching to OpenStation from `/wp-admin/network/` landed you on the main site's dashboard, because the target allowlist only resolved site filenames and quietly dropped the rest. 

## Testing Instructions

**This cannot be tested on the Playground link.** Playground serves a single-site install, and every path below needs a network. Use a local multisite:

1. Add `{ "multisite": true }` to `.wp-env.override.json`, run `npm run env:start`, then create two sites:
   ```
   npx wp-env run cli wp site create --slug=site2 --title="Second Site" --email=wordpress@example.com
   npx wp-env run cli wp site create --slug=site3 --title="Third Site" --email=wordpress@example.com
   ```
2. Log in as `admin` / `password` and turn OpenStation on.

**The network admin**

3. Go to `/wp-admin/network/` and click **Switch to OpenStation** in the admin bar. Make sure you land on the network desktop, not the main site's, and that the address is `wp-admin/network/admin.php?page=openstation`.
4. Make sure the dock reads Dashboard, Sites, Users, Themes, Plugins, Settings, all in the first group with the Core menus, and that each one opens the network screen of that name (Sites is the network's site list, not a 404).
5. Open **Switch to Classic Admin** from that desktop. Make sure you return to the network admin, not to the main site.

**Leaving an admin**

6. From the network desktop, open Sites > All Sites, then click **Dashboard** on any site's row. Make sure it opens in a **new browser tab**, that the desktop you were on is still there in the original tab, and that its dock has not changed.
7. Repeat with the main site's own Dashboard row. Same expectation: a new tab. (This one shares the `/wp-admin/` prefix with the network admin, so it is the case that regressed twice.)
8. Still on the network desktop, open Sites > All Sites and click **Edit** on a site instead. Make sure that one opens in a window, since it is a network screen.

**The dock tile**

9. Go to any site's desktop. Make sure a **Network Admin** tile sits second in the dock, right after Dashboard, and that hovering it fans out Dashboard, Sites, Users, Themes, Plugins, Settings.
10. Click a row. Make sure it opens in a new browser tab and the site desktop you were on is untouched.
11. Switch to the split layout in OpenStation Preferences > Appearance. Make sure the tile moved to the top of the sidebar with the rest of the admin menu.
12. Make sure the tile is **absent** on the network desktop, where the dock already is that menu.
13. As a user without `manage_network` (create an editor on one site and log in as them), make sure there is no tile at all.

**Per-site desktops**

14. On site 2's desktop, open two or three windows and arrange them. Go to site 3, open different windows. Return to site 2 (`/site2/wp-admin/`).
15. Make sure site 2's desktop is exactly as you left it, and that site 3's windows are not in it.

**Single site, unchanged**

16. On a plain single-site install, make sure there is no Network Admin tile, and that switching in and out of OpenStation behaves as it did before.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-704-b6144259dda263c79ebaaa8d0057546a7fc905b6.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


epeicher:
Took this over and fixed the remaining issues. Root cause: the network admin shared the main site's session blob, and the two desktops derive the same window ids from different admins (`index-php` is the site dashboard on one, the network dashboard on the other). So the network desktop restored the site's dashboard window from the shared session, and the dock's Dashboard tile folded onto it. The reverse leak happened too.

Fix (b6144259): the network admin now has its own session key (`desktop_mode_session_network`). The session REST route runs in the main site's blog context whichever desktop saves, so the network screen's `sessionUrl` carries `network=1`, honoured only with `manage_network`. Both read and write filter windows to the session's own admin scope, so blobs polluted before the split heal on the next load.

Re-verified the PR's test plan on a local subdirectory multisite: network dock opens network screens, site rows open browser tabs, per-site desktops stay independent, exit/re-enter round-trips to the right admin.

Issue #708, opened 2026-08-29
Bottom and Left Dock Problem


Eiron-Van:
dynamic docks automatic hide problem
when hovering on the menus of plugins the dock disappear preventing from clicking on the menus

## Steps to reproduce

activate dynamic dynamic for either left and bottom docks - hover on the plugins and hover on the menus that appear - after hovering the docking will hide with the menu.

<img width="1123" height="399" alt="Image" src="https://github.com/user-attachments/assets/6f4a8644-76cd-4d2b-8f4f-3d90d80a4e7d" />

<details><summary>Environment</summary>

```
Plugin version:    1.1.4
WordPress version: unknown
User agent:        Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/151.0.0.0 Safari/537.36
Viewport:          1920x1006
Platform:          Win32
Current URL:       https://omnihub.com.ph/wp-admin/index.php?desktop_mode_portal=1&desktop_mode_portal_intent=1
```

</details>

AllTerrainDeveloper:
Releasing a hotfix for this! Thank you!

AllTerrainDeveloper:
1.1.5 is being released in 6 hours

PR #709, opened 2026-08-29
Hotfix: a dock flyout holds a rail out, it never summons a parked one

AllTerrainDeveloper:
## The bug

With Split layout — static sidebar on the left, **dynamic** dock at the bottom — hovering a sidebar tile to fan out its constellation menu **summoned the bottom dock**. The pill rose over the panel, landed under the pointer, and took the `pointerout` that dismissed it. The menu was unusable: it closed the moment you moved into it.

## Root cause

`src/dock-behavior.ts` → `wanted()`:

```ts
if ( overFlyout ) {
    return true;   // ← for EVERY dynamic rail
}
```

`overFlyout` is one boolean for the whole shell, set from `target.closest( '.os-constellation, .os-dock-peek' )`. Both flyouts are appended to `document.body` and neither records which rail opened it, so the test can only ask *"is the pointer on a flyout"* — never *"on **this** rail's flyout"*. Answering `true` turned a rule meant to stop a revealed rail retracting out from under its own panel into one that summoned every dynamic rail in the shell, on behalf of a menu belonging to a different (here: static) rail.

The dismissal follows mechanically: the revealed pill covers the panel's lower rows, so the next `pointerout` off the panel has a `relatedTarget` that is neither the panel, the anchor tile, nor any tile — `stillInside` is false and the constellation's hide timer closes it.

## The fix

A flyout can **hold** a rail out; it can never **summon** a parked one.

```ts
if ( overFlyout ) {
    return rail.classList.contains( REVEALED_CLASS );
}
```

No ownership lookup needed: a rail that is already out is the only one a flyout can be holding, and that is precisely the job the rule was added for — the module docblock already said the flyouts "would otherwise retract the rail out from under the pointer".

This also closes a second path to the same symptom that ownership alone would have missed: a tall panel running down into the dock's own 20px edge band, where travelling the menu's rows summoned the dock on top of them. The pointer is reading a menu there, not reaching for the dock.

A deliberate summon still works — pointer to the bottom edge, *not* over the panel, reveals the dock as before.

## Tests

Two regressions in `tests/vitest/dock-behavior.test.ts`:

- `a flyout does not summon a parked rail` — the rule in isolation.
- `a static rail's flyout leaves the dynamic rail parked (Split)` — the reported shape: static `#os-side-dock` + dynamic `#os-dock`, pointer onto the sidebar tile, then its body-level panel, then down the panel into the dock's edge band. The dock stays folded throughout.

The existing `a flyout under the pointer holds the rail out` is unchanged and still passes — it reveals the rail first, which is the case that must keep working.

## Docs

`docs/hooks-reference.md` — the `openstation_dock_behavior` / `dynamic` row said "or on one of its flyouts", which reads as though a flyout summons. Now states the hold-only rule and why.

## Gates

`npm run build` · `npm run lint` · `npm run typecheck` · `npm run test:js` (415 files, 5222 tests) — all green. No PHP touched.

## How to verify on :8889

1. OpenStation Preferences → Appearance → Desktop layout → **Split**, Dock behavior → **Dynamic**, Sidebar behavior → **Static**.
2. Hover a left-sidebar tile until the constellation fans out, then move the pointer into the panel and down its rows.
3. Before: the bottom dock rises and the panel vanishes. After: the dock stays a line, the panel stays open, rows are clickable.
4. Sanity: pointer to the bottom edge on bare desktop still summons the dock; opening the dock's *own* constellation and moving up into it still holds the dock out.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-709-b767fa7d596e82b2086b1d52791c0c36e0876f10.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
Closes https://github.com/WordPress/openstation/issues/708

PR #711, opened 2026-08-29
Window channels: one set of books, so a window can leave its loading overlay

### LINKED ISSUE #707: Windows never leave the loading overlay — content loads, os-ready arrives and is accepted, but WINDOW_CONTENT_LOADED never fires
Observed on v1.1.4. Every window opens to the loading overlay (the pulsing logo) and stays there indefinitely — **iframe and native windows alike** (a native window with no iframe and `_iframeBridgeReady: false` sticks the same way, so the iframe bridge leg below is incidental; the common leg is the loading bookkeeping). The content behind the overlay loads completely.

Onset was **mid-session**: the same v1.1.4 install had been opening windows normally for hours, then every window began sticking, and it survives hard reloads and a settings revert (I tested a custom accent I had set that afternoon — reverting it changed nothing). No OpenStation change at the boundary.

**Measured, link by link** (macOS Chrome, live install, all from the shell document's console):

1. The window's iframe reaches `readyState: "complete"`, with the correct `os-chromeless` body class and all bridge scripts present. **Zero console errors** in shell or iframe across a fresh load.
2. The chromeless bridge's `os-ready` **does arrive at the shell**: a `message` spy on the top window sees it with the right origin and `event.source === iframe.contentWindow`.
3. The shell **accepts it**: the Window instance shows `_iframeBridgeReady: true` — so `handleWindowMessage` reached the `os-ready` branch and `markWindowContentReady( win.id )` ran.
4. Yet the window body stays `os-window__body--loading`, which means `HOOKS.WINDOW_CONTENT_LOADED` never fired.
5. Firing it by hand — `wp.os.hooks.doAction( wp.os.HOOKS.WINDOW_CONTENT_LOADED, { windowId } )` — **clears the overlay instantly and the window is fully usable**, so the subscriber side (`src/window/loading.ts`) is installed and healthy.
6. Injecting a **second** `os-ready` from inside the iframe's own realm (so origin and source both pass the `handleWindowMessage` gates — receipt confirmed by the same spy) does **not** clear it.

So the break is confined to one line: `markWindowContentReady`'s `_loadingWindows.delete( windowId )` returns `false` — the id is not in the Set that call consults, even though *something* marked the window loading (the overlay class and `aria-busy` are applied normally). Every guard on this path returns silently, which is why the failure produces no signal anywhere.

## Reading

`src/window-channels.ts` keeps `_readyWindows` / `_loadingWindows` as plain module-level Sets. `src/window/iframe-bridge.ts`'s own comment a few lines away documents the hazard that shape carries — "a plain module-level `let` here would give each bundle its own private copy" — and routes its admin-link deps through `createSharedStore` for exactly that reason, citing AGENTS.md § "Cross-bundle state". If `window-channels.ts` is included by more than one bundle, the copy that registers the loading mark and the copy the `os-ready` path deletes from are different objects, and the loaded action can never fire. That would produce precisely the six measurements above, including why a hand-fired action works while a hand-delivered `os-ready` doesn't.

I can't say from outside which bundles duplicate the module — that reading is offered as the shape that fits every measurement, not as a bisection.

Two adjacent observations from the same session, noted in case they're useful, not measured further:

- After the manual rescue, the `os-window__body--revealing` class also persists for minutes (its `animationend` cleanup never runs). Possibly the same family, possibly fallout from clearing the overlay by hand.
- Closing a hand-rescued window leaves the desktop unable to open that window again until a full page reload — consistent with the same bookkeeping retaining state the open path won't re-enter.

## Workaround

Per stuck window, from the shell document's console (step 5 above, packaged):

```js
wp.os.hooks.doAction( wp.os.HOOKS.WINDOW_CONTENT_LOADED, { windowId: wp.os.windowManager.getAll().at( -1 ).id } );
```

The rescued window is fully usable. Two caveats: after **closing** a rescued window it won't open again until a full page reload (the second adjacent observation above), and plain `/wp-admin/` is unaffected throughout.

AllTerrainDeveloper:
Fixes #707.

## What was happening

`src/window-channels.ts` keeps five per-window registries — parent-side subscribers, native-side subscribers, transport readiness, visual loading state, and the queue holding a `send()` issued before the content is ready. All five were plain module-level `Map`s / `Set`s.

That module compiles into **two** bundles. `createWindowElement()` marks a window loading from `window-system.js`; `native-windows.ts`' synthetic-iframe readiness signal and `connection/index.ts`' subscriber registration run from `desktop.js`. Each bundle therefore kept its own books, and every pairing failed silently in one direction.

The visible one is the loading → ready edge: the copy that registered the mark and the copy the ready signal deleted from were different objects, so `_loadingWindows.delete()` returned `false`, `WINDOW_CONTENT_LOADED` never fired, and the window sat under its loading overlay with its content fully painted behind it. Every guard on that path is an early return, which is why it produced no signal anywhere — no console error, a hand-fired action clearing it instantly, and a second `os-ready` doing nothing. That is exactly the six-measurement set in the issue.

Three more pairings broke the same way:

| Pairing split across the seam | Symptom |
|---|---|
| `enqueueWindowSend()` ↔ the flush | a `Window.send()` issued before ready is never delivered |
| `addParentSubscriber()` ↔ `dispatchFromWindow()` | `Window.on( channel, cb )` never fires |
| `addNativeSubscriber()` ↔ `dispatchToNative()` | native `windowApi.on( channel, cb )` never fires |
| `clearWindowChannels()` on close | the other copy keeps a closed window's ready flag |

That last row is the issue's second adjacent observation: a window rescued by hand and then closed hits `readyWindows.has( id ) === true` in the stale copy on reopen, skips its flush, and sticks again — which reads as "won't open until a full page reload".

## The fix

All five registries move into one `createSharedStore( 'desktop-mode/window-channels' )` record — the treatment `src/window/iframe-bridge.ts` already applies to its admin-link deps two files over, for the same reason and citing the same `AGENTS.md` section.

## Tests

Vitest imports both sides of a seam into one module graph, so a bundle split is invisible to it by default. `tests/vitest/window-channels-cross-bundle.test.ts` builds the seam deliberately: `vi.resetModules()` between two dynamic imports yields two module instances, which is what two Vite IIFE bundles produce in the browser. **Five of its six pairings failed before this change**; all six pass after.

`tests/vitest/bundle-shared-state.test.ts` gains the shell ↔ `window-system` seam, pinning `desktop-mode/window-channels` and `desktop-mode/admin-link-deps` in both built bundles — so a revert to module-level state fails there too.

## Verified manually

On a live install: an `iframeContent` native window (the deterministically-broken case — its loading mark and ready signal always landed in different bundles) now clears its overlay and fires `os-window-content-loaded`. Ordinary iframe and native windows unaffected; close/reopen cycles clean.

## Docs

`docs/architecture.md` gains a subsection under "Two window types" describing the one channel bus and why its state crosses the bundle seam.

## Not resolved

The report says *every* window stuck, iframe ones included. For a plain iframe window both ends of the handshake (`createWindowElement`, `iframe.load`, `os-ready`) sit in `window-system.js`, so I could not reproduce a stick there from this seam alone. This change removes the whole class; if a plain iframe window still sticks on this build there is a second cause worth a separate issue.

## Checks

`npm run build` · `npm run lint` · `npm run typecheck` · `npm run test:js` (5256 passed) — all clean. No PHP touched.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-711-e017890a151e332c6fc89c352b5ec95f541c58c2.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
@claude can you review? ae we missing other hooks?

AllTerrainDeveloper:
Good catch — verified and fixed in e017890a.

I checked it against the built bundles rather than the import graph, and it reproduces exactly as described:

```
desktop.js:18893       const _syntheticIframes = new Map()   ← the ONLY .set()
window-system.js:1690  const _syntheticIframes = new Map()   ← only use is .get()
```

Nothing ever wrote the `window-system` copy, so `getSyntheticIframe()` returned `null` unconditionally in the bundle where `Window.send()` runs. For an `iframeContent` native window `this.iframe` is null as well, so `send()` resolved no postMessage target and fell through to `dispatchToNative()` — where a shell-synthesised body has no subscriber, because there is no render callback to have called `windowApi.on()`. Silently dropped, exactly as you described.

**One correction to the scope.** Of the four registries, only `syntheticIframes` is actually split:

| Registry | desktop.js | window-system.js | Split? |
|---|---|---|---|
| `_syntheticIframes` | 6 | 2 | **yes — live bug** |
| `_connections` | 7 | 0 | no |
| `_connectionsByTarget` | 7 | 0 | no |
| `_connSeq` | 2 | 0 | no |

The other three are shell-only today. I moved them anyway and said so in the commit message as hardening rather than a fix — the module is already known to be multi-bundle, so leaving three module-level registries behind just means the next symbol a lazy bundle imports from here re-opens the same hole.

`connection-cross-bundle.test.ts` drives the seam; three of its four cases fail without the change. The seam builder both cross-bundle tests use is now `tests/vitest/helpers/bundle-seam.ts`, and `desktop-mode/connection` is pinned in `WINDOW_SYSTEM_SHARED_KEYS`.

Your minor note also checks out — `desktop-files/openers.ts` state (`userAssociations`) appears only in `desktop.js`, so not an active bug. Left alone.

Gates on the updated branch: `npm run build` · `lint` · `typecheck` · `test:js` (419 files, 5261 tests) all green.

PR #712, opened 2026-08-29
Command palette: a harvested command that throws says so, instead of nothing at all

### LINKED ISSUE #705: Command palette lists commands registered via the public core/commands store but never invokes them — native commands fire, publicly-registered ones silently no-op
> **UPDATE (same day, after deeper live instrumentation — details in [the follow-up comment](https://github.com/WordPress/openstation/issues/705#issuecomment-5456956037)):** the framing below is superseded in one important way. The command plumbing is healthy end to end — match, pick, run, and render all work when driven programmatically (`wp.os.ai._real._pickCommand(...)` runs the command and renders its result), and this applies equally to `wp.os.registerCommand` and `core/commands`-store registrations. What is dead is the path from **real input events** (keyboard Enter, mouse click on the highlighted row) into `_onSubmit` / the row click handler, on the identical healthy state. So this is an event-wiring defect in the Site Assistant, not a registry/invoke-table split. The original report is kept below for the record.

Measured on **v1.1.4**, on a plain wp-admin page with **no OpenStation windows open**. The same user-visible behavior existed on v1.1.3, so this does not appear to be introduced by v1.1.4 — but the mechanism below is only measured on v1.1.4.

## Observed

A plugin registers 19 commands through the public `@wordpress/commands` store at boot. In the ⌘K palette they **list** correctly. Selecting any of them is a **silent no-op**: no error, no network request, palette just closes. Native OpenStation commands fire normally in the same session.

## Measurements that isolate it

1. **The store is healthy, before and after a dead click.** `wp.data.select( 'core/commands' ).getCommands()` returns all 19 with callbacks attached, and invoking a callback directly from the console runs it correctly — including re-checked immediately after a palette click that did nothing. The palette is not corrupting or rewriting the store.
2. **No windows were open**, so the chromeless iframe bridge and its documented invoke race (`os-commands-invoke` / callback-cache unsubscribe) are not involved. Reproduced identically on the desktop portal and on a plain admin page.
3. **Timing is not the mechanism.** A minimal control command registered from the console *after* the palette runtime had loaded — a fulfilled `REGISTER_COMMAND` dispatch, `category: "action"`, witnessed by the live store —

```js
wp.data.dispatch( 'core/commands' ).registerCommand( { name: 'probe/live', label: 'ZZZ Live Probe', callback: ( { close } ) => { console.log( 'LIVE PROBE FIRED' ); close(); } } );
```

   — **also lists and also never fires.**

## Reading

The palette appears to consult the public `core/commands` store for **listing** while resolving **invocation** from an internal path that publicly-registered commands never reach — only commands originating inside the runtime itself fire. If that reading is right, no plugin can ship a working palette command through WordPress's own commands API.

Worth noting: #683's test checklist verifies third-party commands now "show" (Astra/Woo). On this site they show — and cannot fire. Showing masks the severed half, which may be why this survived that fix's verification.

Happy to run any further console probes against the live install.

AllTerrainDeveloper:
Fixes #705.

## What was happening

The commands were being invoked the whole time. `ShellCommandHarvester.runWithNavCapture()` was discarding what happened:

```ts
try { cb( { close: () => {} } ); }
catch { /* a callback that throws shouldn't break the palette */ }
```

The reporting plugin ([`signal-and-noise-tools`](https://github.com/juanlentino/signal-and-noise-tools)) writes every command as a helper call:

```js
callback: function ( args ) { run( 'Purge caches', 'signal-noise/purge-all-caches', {}, args.close ); }
callback: function ( args ) { navigateTo( tab.url, args.close ); }
```

The callback's own source mentions no `location`, so `classify()` files it as an `action` — which is why the commands **listed** correctly, and why #683's checklist passed. Their `executeAbility()` then reaches `window.sntAbilityRun`, defined by a sibling script that is not in the palette family the shell replays onto the shell page. It was `undefined`, the callback threw `ReferenceError` on its first statement, and the throw went into that `catch`.

A command failing on its first line and a command running perfectly were the same event from outside. That is why the reporter's own instrumentation kept clearing every layer — match, pick, run and render were all genuinely healthy; the layer that knew was silent.

## Three defects, same six lines

**`close` was a no-op stub.** WordPress documents the handler as `callback( { close } )`, and commands written against that contract call it — usually first, so the overlay is gone before the work starts. The harvester invented its own `() => {}` and ran the command under a palette still sitting over the result.

**A thrown callback was swallowed.** It now propagates to `_runCommand`, which already renders `Command /<slug> failed: <message>` and fires `HOOKS.COMMAND_ERROR` — the surface built for exactly this, never reached.

**The `location` sandbox those two were nested inside never existed.** `location` is `[LegacyUnforgeable]`, so `Object.defineProperty( document | window, 'location', … )` throws `TypeError: Cannot redefine property: location` — measured on a live shell, not inferred. Both `installShadow` calls threw, both were caught, `shadowed` stayed empty, `captured` was always `null`, and the "capture the navigation and open it as a desktop window" branch is dead code that has never run. The file's own header comment and `docs/architecture.md` both already said this interception is impossible; only this function believed otherwise. It goes, and the comments now describe what actually guards the shell.

## A limitation this makes visible

Once errors surface, the reporter will see their `goto-*` commands hard-navigate the shell. That is not a regression — it is what has always happened, now legible. A callback that navigates through a helper names no sink in its own source, classifies as an action, and has no runtime net under it, because none is constructible. `docs/architecture.md` now states this and gives the two ways out: a string-literal URL the classifier can extract, or `wp.os.registerCommand()` with `ctx.openInWindow()`.

## Also

- The palette's busy guards in `_onSubmit` / `_runCommand` returned silently. From outside, a dropped keypress and a command that ran and did nothing are identical — the first question anyone debugging this surface has to answer, and one the reporter explicitly asked for. They log now.
- A command row closes over the command it was rendered from rather than re-deriving it through `data-index` against a freshly computed match list. Correct today only because every registry change repaints synchronously; silently wrong (a row runs its neighbour, or nothing) the moment that stops holding.

## Verification

Reproduced and fixed against the reported shape on a live install, driving real keyboard and mouse events:

| | Before | After |
|---|---|---|
| Helper-call callback that throws | nothing at all | `Command /global-zprobe-throw failed: zzzMissingHelper is not defined` |
| Callback calling `args.close()` | palette stayed open | palette closes |
| `Posts` / `Add new post` / mouse click on a row | worked | works |

Two new test files (7 tests) pin the invoke contract and the row/command pairing. Full suite green: 5268 tests, plus lint, typecheck and build.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-712-692b8442c283c58499bdfc8657057ddb715806e0.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
@claude can you review?

PR #713, opened 2026-08-29
Widgets: declare the card's token contract, so a widget's muted text stops landing at 3.81:1

### LINKED ISSUE #703: Post Stats references --os-ui-color-text-subtle, which is declared nowhere — its text lands at 3.81:1 on the widget's own dark card
Observed on v1.1.4. The bundled Post Stats widget (`.dm-poststats__*`) references `--os-ui-color-text-subtle` in all four of its color rules. That custom property is declared nowhere in the plugin, while 158 other `--os-ui-*` properties are — so every rule falls through to its hardcoded light-theme grey fallback.

**Measured:** on the widget's own dark card, the subtle text lands at **3.81:1**, against **18.40:1** for the sibling `.os-widgets__title` on the same surface. Below WCAG AA for the small text sizes involved.

The declared family uses `--os-ui-fg-muted` / `--os-ui-text-muted`, so this reads as a token rename that left one consumer behind. Pointing the four `.dm-poststats__*` rules at the declared muted token (or declaring `--os-ui-color-text-subtle` alongside its 158 siblings) resolves it either way.

Independent of #702 — different stylesheet, different mechanism — but the same family of defect: a custom property consumed where it is not defined, landing on a fallback that was never designed for the surface it paints.

AllTerrainDeveloper:
Fixes #703.

## Not fixed on trunk — and wider than the issue

`--os-ui-color-text-subtle` is still declared nowhere. It isn't only Post Stats: **five** bundled widget stylesheets read it, and two sibling names are undeclared too.

| Token | Read by | Declared? |
|---|---|---|
| `--os-ui-color-text-subtle` | post-stats, site-views, recent-comments, jazz-quote, starter | ❌ |
| `--os-ui-color-border` | post-stats, site-views, recent-comments, jazz-quote, starter | ❌ |
| `--os-ui-color-text` | jazz-quote, starter | ❌ |
| `--os-ui-color-surface` | documented, read by none | ❌ |
| `--os-ui-color-accent` | post-stats, recent-comments, starter | ✅ but hard-pinned |

These five are a **contract**, not incidental names: they are the list in `src/plugins/starter-widget/styles.css`, the file plugin authors copy to start a widget, which told them the names follow the theme. They never did.

## Why the obvious fix was the wrong one

The issue suggests pointing the rules at `--os-ui-fg-muted`. That regresses Legacy.

`.os-widgets__card` paints `rgba( 20, 20, 22, 0.55 )` — a hard literal, no token — and its `color` is `--os-ui-fg-on-accent`, which Legacy keeps at `#fff`. **The widget card is a fixed dark glass panel in every desktop theme.** So a name that flips with the theme is the wrong chain:

| On the card glass (`#141416`) | Now | via `--os-ui-fg-muted` | This PR |
|---|---|---|---|
| Palette | 3.81:1 ❌ | 8.61:1 | **9.13:1** ✅ |
| Legacy | 3.81:1 ❌ | **2.54:1** ❌❌ | **9.33:1** ✅ |

So the contract is declared mostly as literals in `variables.css`, the one owner. The two text names do chain — through `--os-ui-fg-on-accent` and `--os-fg-muted`, the two names Legacy keeps light because they describe the same on-dark chrome this card is. A theme that wants to move a widget card declares these names itself, the way Legacy already declares the accent.

## Also in scope, same defect

- **The card now reads `--os-ui-color-surface`** for its own background, so the value a widget is told is "the card's glass" is the one the card actually paints.
- **`--os-ui-color-accent` was a hard `#f252fc`.** The accent picker writes `--os-ui-accent` inline on `<body>` and the shell, so a widget told to "use the accent for buttons and links" stayed brand pink after the user picked teal while every control around it moved. It reads `--os-ui-accent` now. Default value unchanged; Legacy declares this name itself and is untouched.

## What changes visually

Widget card metadata gets legible, and **dividers get quieter**: the widgets' own fallback was `#e5e7eb`, a light-theme line that on dark glass drew the loudest edge in the station. `0.12` is what the window and dock edges use.

## Tests

`tests/vitest/widget-card-token-contract.test.ts` (8 tests) parses the contract list back out of the starter widget's own comment and pins:

- every documented name, and every `--os-ui-color-*` a bundled widget reads, is declared;
- both text names clear AA on the card glass **under the palette and under Legacy** — the guard that would have caught the `--os-ui-fg-muted` chain;
- the accent resolves through `--os-ui-accent`;
- the card paints the surface token it publishes.

Legacy's frozen manifest is untouched.

## Docs

- `docs/examples/register-widget.md` — new "Theme tokens" section for widget authors, with the "pick a fallback that reads on dark" warning.
- `docs/desktop-themes.md` — the contract as its own subsection under the `--os-ui-*` palette, flagged as not flipping with the theme.
- `src/plugins/starter-widget/styles.css` — its comment claimed OpenStation "sets these variables on the card body element" (nothing did) and called the accent "Brand blue".

## Gates

`npm run build` ✅ · `npm run lint` ✅ · `npm run typecheck` ✅ · `npm run test:js` ✅ 5276/5276

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-713-ec7aeeb47c0b0ceb06bbb2cdcde33552c006653f.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/assets/css/desktop.css
+++ b/assets/css/desktop.css
@@ -925,7 +925,13 @@ body.os-has-fullscreen-window .os-shell {
 	display: flex;
 	flex-direction: column;
 	pointer-events: auto;
-	background-color: rgba( 20, 20, 22, 0.55 );
+	/*
+	 * The glass is also `--os-ui-color-surface`, the one name in the
+	 * widget contract that describes the card itself. Declared in
+	 * `variables.css` so a widget reading it and the card painting it
+	 * cannot drift apart.
+	 */
+	background-color: var( --os-ui-color-surface, rgba( 20, 20, 22, 0.55 ) );
 	/*
 	 * WIDGET texture slot — layered over the frosted card colour, so a
 	 * translucent texture still gets the blur underneath.
--- a/assets/css/variables.css
+++ b/assets/css/variables.css
@@ -1066,7 +1066,6 @@ body.os-active {
 	   text keeps its own colour, so the wash has to stay light enough
 	   for Starlight to carry over it. */
 	--os-ui-search-highlight-bg: rgba(248, 242, 182, 0.3);
-	--os-ui-color-accent: #f252fc;
 	/*
 	 * The content graph's cluster hue is NOT retinted. Graph node
 	 * colours are composed artwork that encodes meaning — the docs list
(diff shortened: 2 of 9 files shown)


AllTerrainDeveloper:
### Follow-up: Legacy had to answer the two new literals

`Tests_OpenStation_DesktopThemesLegacy::test_legacy_answers_every_palette_literal` caught this — my mistake for not running the PHP suite before opening. The guard is right, and its reasoning is the interesting part: a `var()` fallback fires only when a name is **undeclared**, and a theme compiles onto `body.os-desktop-theme-<slug>`. A name Legacy omits isn't undeclared — it's *inherited from the palette*. So Legacy would have silently got the station's glass and a `0.12` hairline in place of the divider it drew.

Both new literals now answer at their consuming rule's fallback, which per the standing rule *is* the pre-brand value:

| Token | Legacy | Why |
|---|---|---|
| `--os-ui-color-surface` | `rgba( 20, 20, 22, 0.55 )` | The glass `.os-widgets__card` already painted |
| `--os-ui-color-border` | `#e5e7eb` | The header divider four of the five bundled widgets drew |

**The two text names are deliberately not pinned.** They're derivations, which the guard skips by design — they resolve through `--os-ui-fg-on-accent` and `--os-fg-muted`, names Legacy already answers, so the contrast fix reaches Legacy too (9.33:1). Pinning them would have severed that.

One correction to the PR description: **"dividers get quieter" applies to the brand look only.** Legacy keeps `#e5e7eb`, which is the point of Legacy.

Count 467 → 469. The snapshot's own docblock provides for exactly this: the freeze protects *collected* values, and a token minted after the snapshot has none to protect.

Full suite green: **PHPUnit 2646** ✅ (6 pre-existing skips) · `test:js` 5276 ✅ · `lint:php` ✅ · `lint` ✅ · `typecheck` ✅ · `build` ✅

PR #714, opened 2026-08-29
In-window navigation: ask before painting, so a cancelled "Leave site?" stops stranding the window

### LINKED ISSUE #699: Users Window gets stuck in loading state when tab switching is cancelled while Adding an User
Window gets stuck in loading state when tab navigation is cancelled via native `beforeunload` prompt

## Steps to Reproduce:

1. Open the Users window and go to the Add User tab.
2. Fill in form details to trigger unsaved changes state.
3. Click another tab in the Users window (e.g., Profile or All Users).
4. The native browser modal appears warning that changes won't be saved if you leave the page.
5. Click Cancel on the native browser modal.

## Observed Behavior:

- The browser cancels the iframe navigation and moves to selected tab.
- However, the Users window UI prematurely arms the loading overlay spinner and shifts the active tab highlight to Profile.
- Because navigation was cancelled, os-ready never fires for the target page, leaving the Users window permanently stuck displaying the loading logo over the content.

## Expected Behavior:
  - Cancelling navigation from the browser prompt should restore the window to its active state without context switching.
  - The loading logo overlay should immediately clear, and the active tab highlight should stay on the current tab.

AllTerrainDeveloper:
Fixes #699.

## What was happening

Re-pointing a window's iframe is optimistic. `handleTabStripClick` armed the loading overlay and lit the destination tab, *then* set `iframe.src` — because the alternative is a click that looks like it did nothing for as long as the server takes to answer.

That optimism is wrong exactly once. When the page inside is holding unsaved changes — Add User with a generated password, an editor mid-draft, any plugin with its own `beforeunload` — the browser raises its native "Leave site?" prompt over the navigation. A user who answers **Cancel** produces no `load`, no `os-ready`, and no other event. Nothing corrects the guess: the window keeps a spinner nothing will ever clear, over content that never moved, with the highlight on a tab it never reached.

The reproduction in the issue is Core's own guard — `wp-admin/js/user-profile.js` registers `$( window ).on( 'beforeunload', … )` and arms it as soon as the form diverges from what it loaded with, so `user-new.php` is guarded the moment the user types.

**There is no signal to listen for.** A cancelled navigation is the absence of every event, and no timer can infer one either: the document stays alive and running whichever button was pressed, so from outside a cancelled navigation is indistinguishable from a page waiting on a slow response.

`Window.navigateTo()` and `Window.reload()` had the same flaw. Reload was the worst of the three — the loading modifier is also its own re-entrancy guard, so a cancelled reload left the window locked out of reloading again.

## The fix — ask first, withhold rather than take back

New `src/window/unsaved-guard.ts`. Before painting anything, the shell posts the existing `os-bridge-beforeunload-query` (now carrying a `requestId`) and waits 250 ms:

| Answer | What happens |
|---|---|
| Bridge never announced readiness | No query, no added latency — exactly the pre-guard behaviour |
| `prevent: false` (the overwhelming majority) | Paints and navigates back to back, one task later than the click |
| No answer inside 250 ms | Treated as `prevent: false`, so a hung frame can't block a tab click |
| `prevent: true` | Navigates, paints **nothing**. The prompt decides. |

On the `prevent: true` path, **Cancel** leaves the window exactly as the user left it — which is the whole fix, and better than the issue's "clear the overlay immediately", because nothing wrong is ever shown. **Leave** produces a real unload, both bridges post a new `os-iframe-unloading` from `pagehide`, and the withheld paint runs then.

`pagehide` is the load-bearing choice: it fires once, at commit, and is the only thing that separates a prompt the user accepted from one they cancelled.

### Three supporting pieces, each with a reason

- **`requestId` correlation.** The close flow acts on any *uncorrelated* `os-bridge-beforeunload-response` by destroying the window. Without correlation, a submenu-tab click would close the window it was navigating. The window's message handler now skips every response carrying an id; the guard listens for its own.
- **A 15 s expiry on a withheld paint.** Event-loop time, which the modal prompt blocks, so it effectively starts when the user answers. Without it, a navigation answered "Stay" leaves a callback armed to fire on some unrelated navigation minutes later and light a tab for a page nobody is going to.
- **`os-ready` drops a still-armed paint rather than running it.** Overtaken: arming the overlay for a load that has already finished spends the ready edge that would have cleared it — the original bug, in reverse.

### Why no shell dialog here

`close()` answers the same query with an `<os-confirm-dialog>`. It can, because `destroy()` removes the iframe from the DOM and removal discards a document *without* prompting. A navigation has no such exit — the native prompt fires whatever the shell does, so a dialog in front of it would ask the same question twice.

## Trade-off

On the guarded **Leave** path the spinner now appears at commit rather than at click, so there is roughly one request's worth of no feedback after the prompt closes. That is the price of not being able to un-ring the bell; the alternative was the double prompt above. The unguarded path — every navigation on a page with nothing to lose — is unchanged.

If `os-iframe-unloading` is ever dropped (the parent processing it after the frame has already been replaced), the degradation is graceful rather than a stick: no spinner, and `_wireTabNavSync`'s `load` handler syncs the highlight to wherever the navigation actually landed.

## Tests

- `src/window/unsaved-guard.test.ts` — the query's correlation, its timeout, its throw path, listener cleanup; and the three navigation outcomes including that a guarded page navigates with the paint withheld.
- `tests/vitest/window-unsaved-navigation.test.ts` — the same on the **live `Window` class**: the expiry timer, a newer withheld paint replacing an older one, destroy dropping one, and `reload()` / `navigateTo()` end to end through a stubbed bridge. Includes the reload double-click that the loading modifier can no longer catch on its own.
- `src/window/tabs.test.ts` — a submenu click on a guarded page keeps its highlight and its spinner off, then lands both when the unload is reported.
- `src/window/iframe-bridge.test.ts` — a correlated response does **not** destroy the window; `os-iframe-unloading` releases; `os-ready` drops rather than runs.

## Docs

`docs/bridge-protocol.md` gains a "Pre-navigation unsaved-changes guard" section and the `requestId` contract on the pre-close pair (plus a stale reference to the retired inline PHP bridge, corrected to `src/chromeless-bridge.js`). `docs/javascript-reference.md` notes it under Reload, including that `navigateTo()` returns `true` for a navigation it *issued* — the page inside gets the last word.

## Checks

`npm run build` · `npm run lint` · `npm run typecheck` · `npm run test:js` (5299 passed, 424 files) — all clean. No PHP touched.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-714-26794f5d5c92e1f0ecf0e997d643d5bf6b7891fe.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
@claude can you review?

PR #718, opened 2026-08-30
[Proposal] Workspaces — desktops that know what they're for

AllTerrainDeveloper:
> **This is a proposal.** Nothing here is settled — the shape, the wording, the three shipped templates and the decision to keep everything in Overview are all up for discussion. It is a working implementation so the idea can be tried rather than argued about in the abstract, not a claim that this is how it should ship.

Today a virtual desktop is a box for windows: an id and a name. This proposal adds an optional profile alongside it, so a desk can also carry **what it is for**.

## What a user can do

- **Press the `+` in the Overview top bar** — the one door. It opens a wizard whose first step is a **Blank desktop** card, preselected, with **Create** focused: `+` then Enter is a plain new desk, the same two gestures as before. Template cards sit beside it; **Customize** walks Name → Apps → Widgets → Look → Windows, and Create stays on every step so the wizard can be left at any point.
- **Start from a template.** Three are proposed, **named for the job and built around the products that do it**: **Commerce** (WooCommerce orders / products / analytics in full-height columns), **Learning** (Sensei courses / lessons / learners tiled, with the room's pulse beside them), and **Publishing** (a blank page taking two thirds of the screen, the library in the margin, the dock folded away).
- **Narrow the rails to the apps that desk is about** — a Commerce desk shows the store, a writing desk shows Posts and Media.
- **Give a desk its own widget column** — drafts and a timer where you write, traffic where you sell.
- **Give it its own look** — wallpaper, accent, desktop theme, dock placement.
- **Have it open with the right windows, arranged the right way**, the first time they enter it.
- **Capture what they already have** — "Use the windows I have open now", "Use the look I have now" — instead of filling in a form.
- **Restore a desk** they have since tidied, from a button under its Overview tile — and **Edit** it (hover the tile) in the same wizard, Start step removed.
- **Keep this desk.** `/keep-desk` (also a row in `/workspace`) makes the workspace open the way the desk is right now — these windows *where they are*, these widgets, these apps. Positions are saved as grid cells or fractions of the work area, so they survive a resized browser or another display. Saving a plain desktop is how it becomes a workspace.
- **Reach all of it from `/workspace`** in the command palette.
- **Place a window on a grid.** Hold **Option** (Alt on Windows/Linux) while dragging and the desk becomes a **6×6 grid**: the cell the key went down in is the anchor, the cell under the pointer is the cursor, and the window lands on the span between them — a 2×2 from (1,1) to (2,2), or the same 2×2 dragged backwards. **Shake the pointer** to move the anchor to where your hand is and start the placement over without letting go. The grid is proportional to the work area, so it is the same six columns on any display — and a placed window **stays on its cells**: resize the browser, move the dock, reload on another screen, and a 2×2 at (1,1) is still a 2×2 at (1,1). While the grid is up the held window goes translucent so you can see where it will land.
- **Listen for shakes.** The shake is a new pointer gesture the platform doesn't have — published as `os-pointer-shake` on the dragged window and as the `os.pointer.shake` action, for anything a plugin wants to hang on it.

## Why it might be worth having

- **One WordPress admin becomes several workspaces.** The same site, seen as a shop, a classroom, or a desk to write at — without plugins fighting over one menu.
- **Nothing is lost by trying it.** A workspace is a *view*: the rails are recomputed, the widget column is mounted and unmounted, the appearance is layered over the user's settings and lifted on exit. `navPlacement`, the enabled-widget list and the settings are never written to, so a workspace someone deletes costs them nothing.
- **Templates degrade instead of breaking.** They name what they are about (`post_type=product`, `sensei`) rather than nav ids, so the Commerce desk on a site without WooCommerce is a smaller desk, not four permission errors.
- **A plugin can ship a workspace from one PHP filter** — `openstation_workspace_presets` adds one *or drops a shipped one* a site has no use for.
- **A desktop with no profile behaves exactly as it does today.** Every existing session is in that state; nothing assumes the field is there.

## A note on the names

The templates are named for the **job**, not the plugin. A desk called "Woo" is wrong on a store running something else, and wrong again the day the product is renamed — but the work is commerce either way. The products are still what the templates reach for: the token lists name WooCommerce and Sensei directly, so on a site that has them, Commerce is a WooCommerce desk in everything but its label, down to the accent colour.

## Open questions for reviewers

- Are Commerce / Learning / Publishing the right three jobs to ship?
- Is Overview the right home for all of this, or should the wizard be reachable without opening Overview?
- Is **Restore** the right word for "put this desk back the way its workspace defines it"?
- Should a workspace be allowed to set the *appearance* at all, or is that a step too far into the user's own settings?

## Scope

New: `src/workspaces/`, `includes/workspaces.php`, `assets/css/workspaces.css`, a lazy `workspace-wizard` bundle, `docs/workspaces.md` + an example; `src/window/shake.ts` (the pure gesture detector) and `src/window-manager/grid-snap.ts` (the grid geometry + drag session). Touched: the overview top bar, the layout dispatcher's placement input, the widget layer (one new mount-only method), OS settings (an override layer that is never persisted), the session sanitizer, two new arrangements on the window manager, and the title-bar drag (modifier tracking + shake feed, one `onDragGesture` callback).

Gates: 5445 JS tests, 2668 PHP tests, `phpcs -n`, typecheck, eslint and build all green.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-718-b887a9c9911ef510c7f19c93134a7d546276fc6e.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


nickhamze:
I wouldn't call them by their product names but I would make sure our products are used in the workspaces.

Maybe do Commerce / Learning / Publishing

nickhamze:
It's a great concept though, I can see it being extremely useful.

AllTerrainDeveloper:
> It's a great concept though, I can see it being extremely useful.

@nickhamze What do you think about this? 😆
I believe this concept should be applied in every OS out there.

https://github.com/user-attachments/assets/ab076b6d-1d92-434c-954c-15ed9812037d

AllTerrainDeveloper:
@claude can you review?

epeicher:
I love the tiling window functionality @AllTerrainDeveloper! While reviewing this, I have noticed a small UI regression on the `+` button 
CleanShot 2026-08-31 at 16 11 41@2x

I have also used Claude for review, and it has found the following:

Reviewed on the branch. Nice piece of work, the view-not-write discipline holds up and the degradation story is real. Five things I could reproduce:

---

**1. `includes/workspaces.php:311`: the widgets loop clobbers the preset id**

`foreach ( $raw['widgets'] as $id )` reuses `$id`, set to the preset id at :280 and returned at :358.

```php
add_filter( 'openstation_workspace_presets', function ( $p ) {
    $p[] = [ 'id' => 'support', 'label' => 'Support', 'widgets' => [ 'clock', 'desktop-mode/recent-comments' ] ];
    return $p;
} );
$added = end( openstation_workspace_presets() );
// id => 'desktop-mode/recent-comments'  (expected 'support')
```

Drop the `label` and the label falls back to the widget id too. It also lets a `/` into an id `sanitize_key()` would have stripped, so `server-sync.ts` registers under the wrong key and the retire pass never finds it. The test at `openStationWorkspaces.php:338` asserts the widgets array but not `$added['id']`. Rename the loop var.

---

**2. `src/settings/index.ts:626`: leaving a workspace discards settings edits made on it**

`this.state = base` restores the entry snapshot, so an edit is reverted in memory even though `save()` already sent it to the server, and the next save writes the stale value back.

Repro (unit): wallpaper `galaxy` / accent `pulse`, then `setWorkspaceAppearance({ wallpaper: 'mono' })`, then set `state.wallpaper = 'aurora'` and `state.accent = 'teal'`, then `save()` persists both correctly, then `setWorkspaceAppearance(null)` and state is back to `galaxy` / `pulse`. The next `save()` writes those.

Repro (UI): on a Publishing desk, open Preferences, change the accent, switch desks. The change is gone.

Amplifier: `setWorkspaceProfile` calls `applyWorkspaceView` calls `setWorkspaceAppearance(patch)`, which rebuilds `state` from `base` too, so just adding a widget to the desk reverts an in-flight Preferences edit.

This contradicts `docs/workspaces.md` ("A key they *did* change is theirs, and it is saved"). The restore path needs to merge edits forward, close to what `_persistableState()` already computes.

---

**3. `src/desktop.ts:3193` runs before session restore at `:3651`**

The boot `applyWorkspaceView` fires before `seedDesktops()` attaches the profiles, so on reload the profile is `null` and the look and widget column are never applied.

Repro: create a Commerce desk, stay on it, reload. Rails stay narrowed (the dispatcher reads the profile fresh on every repaint), but wallpaper, accent, dock and widget column are yours again until you switch desks and come back.

---

**4. `resolveLaunches` drops captured windows the navigation can't name, even with a `url`**

`captureWorkspaceWindows` walks `mgr.getAll()`, but restore requires `match` to hit a `getNavItems()` entry, and that list is pre-`computeNav`, so it has no ephemeral tiles for open windows.

Repro: open a post in the editor, run `/keep-desk` (the toast counts it), close the windows, hit **Restore** under the tile. The editor window does not come back.

Unit form: `resolveLaunches( navItems, [ { match: 'post-php-post-42-action-edit', url: 'post.php?post=42&action=edit' } ] )` returns `[]`.

An entry carrying an explicit `url` and not native should open without needing a nav match.

---

**5. Preset server sync is boot-only**

`workspacePresets` is not in `openstation_build_menu_payload()` and `applyServerWorkspacePresets` is not wired into `createApplyPayload()`.

Repro: with the shell open, activate a plugin that adds a template via `openstation_workspace_presets`. It does not appear until F5. Same for removal.

`server-sync.ts` says "the boot payload and every later menu refresh both land here" and the retire branch says "the plugin that added it was deactivated", but neither path exists yet.

---

Smaller notes:

- `src/workspaces/command.ts:21`: `const NEW_PREFIX = 'New: '` is untranslated and concatenated onto a translated label. Wants a `sprintf`.
- `src/window/shake.ts`: `return run.extent >= 0` in `step()` is always true where it's reached.
- Shake needs at least 1000ms sustained with no gap over 320ms, so a natural quick shake takes roughly ten reversals to register. Worth a play test; release and re-press of Alt is an equivalent anchor reset, so not blocking.
- `arrange.ts` `focus()`: `stackHeight` can go negative on a long stack. `columns()` guards this by handing off to `tile()` past four, `focus()` has no floor.
- Nothing pins the two lists the new AGENTS.md row says must stay equal (preset ids, appearance allowlist). Same gap as the desktop-theme slots so it matches the norm, but issue 1 is exactly what that test would have caught.

On the open questions: Commerce / Learning / Publishing feel like the right three. Restore reads fine to me given the `aria-label` carries the full sentence. Appearance-as-a-workspace-setting is the right call, but issue 2 is the reason it currently feels risky.

Issue #716, opened 2026-08-29
Add shortcut key command to close all open windows


AnInfiniteNumberofMonkeys:
Add shortcut key command to close all open windows

<details><summary>Environment</summary>

```
Plugin version:    1.1.4
WordPress version: unknown
User agent:        Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/150.0.0.0 Safari/537.36
Viewport:          2367x1368
Platform:          MacIntel
Current URL:       https://oceansev.com/wp-admin/edit.php?post_type=page
```

</details>

AllTerrainDeveloper:
@AnInfiniteNumberofMonkeys We'd like to interview you, either myself or @epeicher. No need to do a video call or giving real names it can be fully async, and over WordPress (org) slack. We'd like to understand our users, what areas we can improve by listening your use cases.

Would you be up to that?

AnInfiniteNumberofMonkeys:
Yeah, I'd be up for that!

mgiannopoulos24:
@AllTerrainDeveloper Could you close this too since I implemented it. Thanks 👍

AllTerrainDeveloper:
Closed as @mgiannopoulos24 implemented it!

PR #719, opened 2026-08-30
Feat: Add close all windows keybinds.

mgiannopoulos24:
# Close all windows: a shortcut for it, and a dialog that can be told to stop asking

`⌥⌘W` on macOS, `Ctrl+Alt+W` everywhere else, closes every open window on every desktop.

`WindowManager.closeAll()` and its hook chain have been in the framework for a while with no way for a user to reach them — the only callers were plugin code and a docs example. This adds the keyboard route, the confirmation in front of it, and the one preference the confirmation needs to not become a nuisance.

## Related issue
Issue #716.

## The chord

`⌥⌘W` is the chord macOS already uses for "Close All Windows" (Finder's is exactly this), so the shell borrows it rather than inventing one. `Cmd/Ctrl+Shift+W` — the obvious other candidate — is the browser's own close-window shortcut and cannot be taken back from it.

The key is matched on `e.code === 'KeyW'`: Option+W types `∑` on macOS, and the W glyph moves between keyboard layouts. `code` doesn't. There's no text-entry gate — the chord carries two modifiers, so it types nothing into a field, unlike the bare-backtick window switcher that has to yield.

Pressing it inside an admin window works too. Native keydown doesn't cross an iframe boundary and every admin window *is* an iframe, so `src/chromeless-bridge.js` claims the chord in capture phase and forwards it up as `os-window-close-all` — the same shape as the `Cmd+K` and backtick forwarders sitting next to it.

## The confirmation

The chord asks first, naming how many windows are about to go. This isn't redundant with the existing per-window unsaved-changes prompt: that one only fires for a page actually holding unsaved state, and says nothing about the other eleven windows the chord is about to take with it.

The dialog carries a **"Don't ask again"** checkbox. Ticking it writes `confirmCloseAllWindows: false` to the user's OpenStation Preferences — user meta, so the choice follows them to their other browsers — and the chord closes immediately from then on. Pages holding unsaved changes still get to raise their own prompts.

**OpenStation Preferences → Windows → "Ask before closing all windows"** turns it back on. That half is the point: an opt-out with no way back is a trap, not a preference.

The close itself is just `closeAll()` — no reimplementation — so `os.windows.before-close-all`, the `os.windows.close-all` protection filter and `os.windows.after-close-all` all behave exactly as documented, and a plugin protecting a window through the filter still wins.

## `<os-confirm-dialog>` grew a `remember-label`

The checkbox is a general component feature, not a one-off: set `remember-label` and the dialog renders it, reporting its state on the `os-confirm` detail as `remember`. The `osConfirm()` wrapper surfaces it as an `onRemember( remember )` callback so the Promise stays a plain `Promise<boolean>` and every existing caller is untouched.

`onRemember` fires only on **confirm** — a question the user backed out of was never answered, so it can't have meant "stop asking". The checkbox is a native `<input type="checkbox">` rather than `<os-checkbox>`: this component ships in the shell-overlays bundle, which registers only the tags a triggered overlay needs, and a dialog rendering an unregistered tag renders an inert box. It joins the dialog's existing tab trap for free.

## What changed

| Area | Files |
|---|---|
| The shortcut | `src/window-manager/close-all-shortcut.ts` (new), wired in `src/desktop.ts` |
| Iframe forwarder | `src/chromeless-bridge.js` |
| Dialog checkbox | `src/ui/components/os-confirm-dialog/*`, `src/os-confirm.ts` |
| The preference | `src/settings/{types,constants,state,panel}.ts`, `src/settings/sections/close-all.ts` (new), `includes/os-settings.php` |
| In-app shortcut reference | `includes/admin-bar.php` |
| Docs | `docs/javascript-reference.md`, `docs/bridge-protocol.md` (new "Keyboard forwarders" section, which also documents the two forwarders that were never written down), `docs/components-reference.md` |

## Tests

- `tests/vitest/close-all-shortcut.test.ts` (new, 12) — chord matching including the near-misses and the `∑` case, confirm/cancel, empty desktop, no stacked dialogs on a held key, the protection filter surviving, the "don't ask again" persistence and its confirm-only rule, the no-dialog path, and both entry points (keypress + bridge message).
- `src/ui/components/os-confirm-dialog/os-confirm-dialog.test.ts` (+5) — the checkbox renders only when asked for, reports ticked/unticked on confirm, reports nothing on cancel, and joins the tab trap.

Full gates green: `npm run test:js` (5332 tests / 426 files), `lint`, `typecheck`, `lint:php` (`phpcs -n`), `npm run build`, and `test:php --filter=OsSettings` (101 tests).

### The confirmation, with "Don't ask again"

<p align="center">

image

</p>

### Preferences → Windows → Ask before closing all windows

<p align="center">

image

</p>

### The shortcut in the Keyboard shortcuts window

<p align="center">

image

</p>

---

Let me know your thoughts on this PR!

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-719-798e3d9a129bfa2037f607e77a97e5ad19e1e5d7.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


mgiannopoulos24:
Also added the `testdox` option on `phpunit.xml.dist` so we can see the test names while it runs each test.

tests/phpunit/phpunit.xml.dist
```diff
@@ -5,6 +5,7 @@
 	bootstrap="bootstrap.php"
 	backupGlobals="false"
 	colors="true"
+	testdox="true"
```
AllTerrainDeveloper, line 8: ❤️

AllTerrainDeveloper:
Thank you and welcome! We are proud of having you participating in OpenStation!

I'll test and review right now :)

src/window-manager/close-all-shortcut.ts
```diff
@@ -0,0 +1,221 @@
...
+ * batch close is not something to fire on a near miss.
+ */
+export function isCloseAllChord( e: KeyboardEvent ): boolean {
+	if ( e.code !== 'KeyW' ) {
+		return false;
+	}
+	if ( ! ( e.ctrlKey || e.metaKey ) ) {
+		return false;
+	}
+	return e.altKey && ! e.shiftKey;
```
AllTerrainDeveloper, line 85: And If I'm not mistaken and Claude kind of supports me, we should add the guard for non MacOs OS.

Any, I'll ask Claude for it :)

```suggestion
	if ( e.code !== 'KeyW' ) {
		return false;
	}
	if ( e.getModifierState( 'AltGraph' ) ) {
		return false;
	}
	if ( ! ( e.ctrlKey || e.metaKey ) ) {
		return false;
	}
	return e.altKey && ! e.shiftKey;
```

AllTerrainDeveloper (commented):
The change is great, tested and works as you intended. But... I think we should ONLY close the windows in the current selected desktop.

Image

One of the reasons is that it feels (at least to me) not quite intuitive. And other more important reason is the arrival of [WorkSpaces](https://github.com/WordPress/openstation/pull/718) with the next release

I spotted another bug:

Image

Probably you should exclude those windows from the total count of closed windows.

AllTerrainDeveloper:
@claude would you mind a quick review and assert my comments in my own test/review?

AllTerrainDeveloper:
Again, thank you very much and looking forward for the followups, I would love to see Aiden's face here!
image

mgiannopoulos24:
Okay I will fix the 2 bugs you mentioned.

mgiannopoulos24:
I hope it's okay now.

src/window-manager/close-all-shortcut.ts
```diff
@@ -156,16 +196,77 @@ export async function closeAllWindows(
...
+			}
+			setTimeout( check, CLOSE_SETTLE_POLL_MS );
+		};
+		check();
+	} );
+}
+
 /** The close itself, once it is going to happen. */
-function runClose( mgr: WindowManager ): number {
+async function runClose( mgr: WindowManager ): Promise< number > {
```
AllTerrainDeveloper, line 248: ❤️

AllTerrainDeveloper (approved):
LGTM! Thank you!

PR #735, opened 2026-09-02
Multisite: the workspace hop — cross-admin clicks switch desktops in place

epeicher:
## What it does

Cross-admin clicks now **switch desktops in the current tab** instead of forcing a new browser tab. Click the Network Admin tile (or one of its flyout rows) on a site's desktop and this tab travels to the network admin's shell; click a site's "Dashboard" link in the network's Sites window and it travels back. Every admin keeps its own desktop under its own session key (#704/#728), so each side restores exactly as it was left — hopping reads as switching workspaces, with a subtle cross-document view transition covering the swap (a gentle crossfade with a hint of zoom; instant under reduced motion; a plain navigation on browsers without the feature).

The browser-tab behavior survives as the side-by-side gesture: **cmd/ctrl/shift or middle click** still opens the other admin in a new tab.

## Rationale

#704 sent every cross-admin click to a new browser tab: one behavior on all three network shapes, and it never disturbed the desktop you were standing on. But with per-admin sessions in place, leaving a desktop costs nothing — it restores exactly as left — and the tab-per-hop model piles up tabs for what is conceptually one workspace switch. The follow-up idea was "open the other admin in a new Workspace with a subtle animation": this is that, built on the machinery that already exists rather than a multi-admin shell (which the doc's iframe-constraint section explains cannot generalize past subdirectory networks anyway).

## Implementation

- `hopToAdmin()` in `src/multisite/hop.ts` owns the gesture split: plain activation → `location.assign( rawAdminUrl )` (the `admin_init` redirect routes it to the matching shell screen exactly as if typed); modifier/middle click → `window.open`. The Network Admin tile and its flyout rows route through it.
- The chromeless bridge's `other-admin` link kind now assigns `window.top.location` instead of opening a tab (falling back to a tab if a host forbids top navigation). Modified clicks never reach that branch — the handler's first guard already yields them to the browser, whose native new-tab handling is the same gesture.
- The view-transition opt-in (`@view-transition { navigation: auto }` plus the two root animations) lives in `assets/css/desktop.css`, which only the **shell** loads — chromeless iframes never opt in, so ordinary in-window navigations never transition, and classic admin pages don't opt in either, so entering/leaving the shell stays a plain cut. Reduced motion sets `navigation: none`: the hop still happens, the animation does not.
- `SystemDockItem.onOpen` and `SubmenuItem.onSelect` now receive the originating `MouseEvent` when there is one (keyboard and programmatic activations pass nothing), threaded through the dock and the constellation so any navigating handler can honour the same gestures. Handlers that ignore the argument are unaffected.
- Cross-origin links (a subdomain or mapped site's admin) were never classed `other-admin` and keep opening tabs — the hop only exists where a same-origin navigation (and its view transition) can.

`docs/multisite.md` gains a "workspace hop" section replacing the tab contract; `tests/vitest/workspace-hop.test.ts` pins the gesture split and the CSS opt-ins (including that no iframe-loaded sheet ever carries one).

## Testing instructions

```bash
npm run test:js   # workspace-hop, multisite-dock-tiles, chromeless-bridge-links
```

On a subdirectory multisite (plugin network-activated, e.g. the `.wp-env.override.json` setup from #704):

1. On a site desktop, arrange a couple of windows, then click the **Network Admin** dock tile: this tab travels to the network desktop (crossfading in Chromium), restoring it as last left.
2. There, open **Sites → All Sites** and click a site's **Dashboard** row: the tab travels to that site's desktop, with your windows from step 1 intact.
3. Cmd/ctrl-click either of those: a new tab opens and the desktop you are on stays put.
4. Toggle "reduce motion" in the OS: hops still work, instantly.
5. Firefox/Safari: hops work as plain navigations.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-735-e707488255c7305f06fff657104353b8bc29ef77.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/assets/css/desktop.css
+++ b/assets/css/desktop.css
@@ -1525,3 +1525,52 @@ os-notice a:focus,
 os-notice a:active {
 	color: var(--os-ui-notice-link-hover, #fffbff);
 }
+
+/* --------------------------------------------------------------------
+ * The workspace hop — cross-document view transitions between shells.
+ *
+ * Loaded only on the SHELL screen (this sheet never reaches chromeless
+ * iframes), so only shell→shell navigations transition: hopping between
+ * a site's desktop and the network admin's crossfades the two desktops
+ * instead of hard-cutting, which is what makes the per-admin sessions
+ * read as workspaces. Classic admin pages never opt in, so entering or
+ * leaving the shell stays a plain navigation, and reloads are excluded
+ * by the spec. Browsers without cross-document view transitions ignore
+ * all of this and navigate plainly. See docs/multisite.md.
+ */
+@view-transition {
+	navigation: auto;
+}
+
+/* Reduced motion swaps instantly — the hop still happens, the
+ * animation does not. */
+@media ( prefers-reduced-motion: reduce ) {
+	@view-transition {
+		navigation: none;
+	}
+}
+
+/* A gentle zoom-through: the desktop being left recedes as the one
+ * arriving settles in. Subtle on purpose — this runs on a full page
+ * swap, and anything louder reads as a reload effect. */
+::view-transition-old(root) {
+	animation: os-workspace-hop-out 220ms cubic-bezier( 0.4, 0, 1, 1 ) both;
+}
+
+::view-transition-new(root) {
+	animation: os-workspace-hop-in 280ms cubic-bezier( 0, 0, 0.2, 1 ) both;
+}
+
+@keyframes os-workspace-hop-out {
+	to {
+		opacity: 0;
+		transform: scale( 0.985 );
+	}
+}
+
+@keyframes os-workspace-hop-in {
+	from {
+		opacity: 0;
+		transform: scale( 1.015 );
+	}
+}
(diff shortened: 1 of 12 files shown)


epeicher:
Superseded by #736: the clarified requirement was Desktops (Spaces), not tab navigation — the new PR opens another admin on its own scoped desktop and folds this branch's machinery in as the cross-origin and modifier-click fallback.

PR #760, opened 2026-09-04
feat(window): add minimize and restore dock animations.

### LINKED ISSUE #759: Introducing animations: genie effect for minimizing windows
## The Current Behavior 
- Right now, when a window is minimized, it simply fades out in place.
- There is no visual connection between the window disappearing and its icon in the dock.
- This can make restoring windows feel a bit mysterious, as the dock feels like a static list of launchers rather than an active part of the workspace.

## The Proposed Solution
- Introduce a smooth "genie" animation (similar to macOS) when minimizing windows.
- The window will visually shrink and fly directly into its corresponding dock icon.
- When restoring, the animation will play in reverse, growing from the dock icon back to its original size and position.

> You can check visual-demo of animations shared in PR description.

## Why this can be better
- **Intuitive UX:** The visual flight path clearly shows the user exactly where their window went, making it obvious how to get it back.
- **Dynamic feel:** It makes the interface feel much more alive and purposeful.

KarunyaChavan:
This PR introduces a "genie" animation for minimizing and restoring windows, replacing the old fade effect. Windows now visually shrink and fly directly into their corresponding dock icon, and grow back out from the dock when restored.

Closes #759 

https://github.com/user-attachments/assets/c8e2e6a2-8806-452e-8855-aa464a3fc76c

### Why?

- The previous fade transition lacked a visual connection between a minimized window and its dock tile, making the action feel disconnected. 
- The genie flight path gives users a clear, intuitive spatial cue ("this window went into that icon"), making it obvious how to restore it. 
- It also makes the workspace feel more dynamic and polished, elevating the overall feel of the OS to be much more premium.

### What Changed?
- **`src/window/index.ts`**: 
  - Replaced the CSS opacity/fade logic with the Web Animations API (WAAPI) to calculate and play the genie trajectory between the window's bounds and its dock tile icon. 
  - Added structural helpers (`_buildGenieTransform`, `_cancelGenieAnimation`, `_clearGenieStyles`) to keep the implementation DRY across the `minimize`, `restore`, and `destroy` lifecycles.

- **`assets/css/window-states.css`**: 
  - Removed the old CSS fade transitions that were previously managing the minimize state, handing visual control over to the JS animation.

- **`src/dock.ts`**: **Existing Bug Fix:** 
  - Fixed an issue where apps opened from shortcuts or plugin-registered launchers (which aren't permanently pinned to the dock) could not be restored by clicking their dock tile. 
  - The click handler for `windowId`-only tiles was only calling `focus()`, which reordered the z-stack but never unminimized the window. We now correctly check the state and call `restore()` if the window was minimized.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-760-2a89c7f14805f5b672215b561e44733c17738734.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
Wow! Is looking great! I would personally add a more vanish towards the end of the minimize animation, what do you think?

AllTerrainDeveloper:
@KarunyaChavan I've asked the same to another top contributor:

would you like to create a guest-post in our blog as one of the tops contributors? :)

You can write about yourself, who are you, why you like OpenStation, why you contribute, the tools you use, etc.

KarunyaChavan:
I’ve added a fade-out at the end of the flight animation when the window is minimized, creating a vanishing effect. I’ve also adjusted the animation duration so it doesn’t feel too quick.

> would you like to create a guest-post in our blog as one of the tops contributors? :)

Sure, it would be my pleasure. Thanks !

AllTerrainDeveloper:
Sorry! We have plenty of work to do. Reviewing it now

AllTerrainDeveloper:
It works awesomely! I'll merge it, and as a follow up PR we could add options here to enable or even modify the animation.

image

The idea is to add hooks so other developer can extend OpenStation.

PR #775, opened 2026-09-07
Claude workflow: a fork pull request gets neither a write-capable token nor an executable skill

AllTerrainDeveloper:
## Why

The `@claude` workflow's `author_association` gate vouches for the person who mentions Claude, never for the code Claude checks out. Every event it listens to is raised on the base repository, so a maintainer can invoke it on a pull request from a public fork. In that run the action runs Claude Code inside a working tree the fork's author controls.

Two security reports reached us about that run.

The first described a chain that exfiltrated the GitHub App token (`contents: write`) through the action's `.claude-pr/` config snapshot. It was written against action v1.0.183; the pinned v1.0.210 (#726) already scopes the snapshot to files inside the working tree, so that exact chain no longer works here. What remained was the underlying exposure: a write-capable credential living inside an untrusted working tree.

The second found a different, independently rooted chain that the action's own defences do not cover. `.claude/skills` on trunk was a symlink to `../.agents/skills`. The action protects `.claude/` by deleting the checked-out copy and restoring it from the base branch, but it restores the link *as a link*, and the link's target is whatever the pull request says it is. A fork could therefore change `.agents/skills/pixijs/SKILL.md` alone, leaving `.claude/` and the workflow untouched, and own a "trusted" project skill. A skill is executable: a `!` expansion in its body runs while Claude Code builds the prompt, before the model sees anything, and the skill's own `allowed-tools` frontmatter widens the workflow's tool list (`Bash(node *)` where the action's outer list has no Bash at all). One ordinary `@claude /pixijs review this` from a collaborator ran attacker shell with `GITHUB_TOKEN`, `GH_TOKEN` and Claude's own OAuth token in the environment. The reporter's replay reproduced it deterministically against the pinned Claude Code release.

## What

**Token exposure (first report)**

- **New step "Resolve pull request origin"** asks the API whether the event's PR head lives in this repository. Plain issues and same-repo PRs are trusted. A fork, or a deleted head repository, is not. Any API failure fails the job, so an unknown origin never reaches Claude with an App token.
- **Fork PRs run with the workflow's own token** via the action's `github_token` input, so no App token is minted. That token is scoped to the job's `permissions`, where `contents` stays read-only, and expires with the job. Nothing running against a fork's files can push to this repository. Claude can still review and comment on the PR; it could never push to a fork's branch anyway, so no capability is lost.
- **Job permissions** gain `pull-requests: write` and `issues: write` so the workflow token can post Claude's tracking comment on the fork path. The App token already carried both in every run.

**Skill execution (second report)**

- **The symlink is gone.** The pixijs skill now lives as regular tracked files under `.claude/skills/pixijs/`, and `.agents/` is removed. Trusted configuration is now made of files the action actually restores. The `.gitignore` comment says why nothing under `.claude/` may be a symlink again.
- **Skill shell expansion is disabled by policy** for every skill source: the action's `settings` input carries `{ "disableSkillShellExecution": true }`, which merges into the runner's user settings. Every `!` expansion becomes a "disabled by policy" marker.
- **Skills and slash commands are off altogether on a fork PR** (`claude_args: --disable-slash-commands`, gated on the origin step). Nothing in that checkout is ours to trust, and Claude's OAuth token is in the subprocess environment whatever GitHub token the run holds. Same-repo branches and issues keep `@claude /pixijs …`.
- **No project memory or settings from a fork checkout.** The action restores `CLAUDE.md` from the base branch, but ours is an `@AGENTS.md` include and the included file is not on the action's restore list, so the memory Claude loaded was the fork's. On the fork path `claude_args` now adds `--setting-sources user` (the Agent SDK loads `CLAUDE.md` only with the `project` source on), and a new step "Collect the trusted agent guide" builds the guide from the default branch with every `@path` include resolved from the same ref, handed to Claude with `--append-system-prompt-file`. `origin` is this repository, never the fork, so the guide is ours whatever the checkout landed on.

Not done on purpose: `GIT_LITERAL_PATHSPECS=1` would neutralise the pathspec-magic filename from the first report, but the action's own auto-commit relies on `:(exclude)` pathspecs and would break. A token-free analysis phase (the second report's third suggestion) needs changes in the action itself, not in this workflow.

## What this does not cover

- **Claude's own OAuth token is in the CLI's environment on every path**, because the CLI needs it. The reported chain read it from a pre-model shell, which no longer runs. A model-mediated read (a prompt-injected `Read` of the process environment, posted back as a comment) is an action-level concern that applies to every repository using it on fork pull requests, and is outside what this workflow can change.
- On the fork path the tracking comment is posted by `github-actions[bot]` rather than the Claude app, since the job token authors it. Cosmetic.

## Verification

- `actionlint` passes on the first commit; the later ones are validated by a YAML parse that resolves the three action inputs as expected.
- Read against the pinned action source: an empty `github_token` maps to an empty `OVERRIDE_GITHUB_TOKEN` and falls through to the OIDC path; tag mode appends user `claude_args` after its own `--allowedTools`, so the bare flag parses as a boolean and the Agent SDK emits it as one; the write-permission check only relaxes when `allowed_non_write_users` is set, which it never is here.
- The origin script was run against a stubbed `gh` for five cases: plain issue, same-repo PR, fork PR, deleted head repository, API failure. Outcomes: `fork=false`, `fork=false`, `fork=true`, `fork=true`, job fails.
- New `tests/vitest/claude-workflow-hardening.test.ts` asks git for the tracked modes under `.claude/` and asserts no symlink and no `.agents/`, parses the `settings` JSON, and checks the fork gating on `claude_args` and `github_token` plus `contents: read`. All five cases pass here and all five fail on trunk, checked in a trunk worktree.
- `disableSkillShellExecution` and `--disable-slash-commands` confirmed against the Claude Code binary and the pinned action's `setup-claude-code-settings.ts`. Full `npm run test:js` (475 files), `npm run typecheck` and `npm run build` are green.
- The memory fix was probed with the pinned Agent SDK 0.3.251 driving the local CLI inside this repository, three ways: with the fork flags the guide and the step's marker sentence are both present; with `user` alone neither is; with the default sources the guide arrives through the include and the marker does not. The guide step's body was also executed as written, in a throwaway clone, and produced the same file as a dry run.
- Not yet exercised on a live pull request.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-775-68e9146a218fb8d81561a3944c6f3e5a64260efa.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
@claude can you review this security patch?

PR #776, opened 2026-09-08
fix: timeout stalled plugin updates

### LINKED ISSUE #765: A stalled plugin update never times out, and wedges the entire update queue until reload — not just its own row
> **Update, 2026-09-10 — root cause found, and the blast radius is larger than this report says.**
>
> The reasoning below is right that the awaited promise never settles. The reason: **there is no timeout anywhere in the chain.** `runUpdate()` → `enqueueUpdateJob()` → `updateInstalledPlugin()` → `ajaxRequest()` → `await fetcher( … )`, and `AbortController` / `AbortSignal` / `signal:` appear **zero** times under `apps/plugins/`.
>
> `fetch()` rejects on a network failure — which is why a 403 or a non-JSON body toasts correctly — but not when the server accepts the request and then holds the socket: a stalled upgrader, a PHP worker hitting `max_execution_time` without flushing, a proxy keeping the connection alive.
>
> **The row lockout is a consequence, not a second bug.** `runUpdate()`'s `finally` correctly deletes the busy flag, but a `finally` only runs when the awaited promise settles.
>
> **What this report missed: one stall wedges the whole window, not one row.** `drain()` sets `inFlight = true` before awaiting and clears it in a `finally` waiting on that same promise, so `inFlight` stays `true` forever. Every later `enqueueUpdateJob()` queues behind a `drain()` that returns at its `if ( inFlight ) return;` guard. After a single stalled update **no plugin in that window can be updated again**, silently, until reload.
>
> Retitled to say that, since "the row then locks itself out" reads as one row's problem.
>
> The mechanism is settled; the fix shape is not, and that is why there is no PR attached. A timeout has to survive a genuinely slow upgrade over a slow link, Core's own `wp.updates` has no client timeout to copy, and aborting the fetch does not abort the upgrader — the server may still commit. Detail in the comment below.
>
> Original report kept below.

---

> **Edited after re-testing.** The first version of this report said there was no busy state and framed the bug as phone-only. Both were wrong: a busy indicator **does** appear beside the button, and the behaviour reproduces on a desk as well as on a phone. That narrows the diagnosis rather than weakening it — corrected throughout below.

## What happens

Clicking a plugin's **Update** button shows a small busy indicator beside the button — and then nothing, indefinitely. No success toast, no failure toast, no completion. The row stays busy. Further clicks on that row do nothing at all.

Reproduced on **both** the desk and the phone layer, including the busy indicator: it appears on a desk too, and never resolves there either.

## Why the busy indicator matters

It proves the click handler ran. `runUpdate()` marks the row busy and repaints *before* awaiting anything:

```js
// apps/plugins/parts/actions.ts — runUpdate()
if ( host.busy.updating.has( row.plugin ) ) { return false; }
host.busy.updating.add( row.plugin );
host.repaint();
…
const result = await enqueueUpdateJob( () => host.rest.updateInstalledPlugin( row ) );
```

So the wiring is fine and `repaint()` works. And every outcome path raises a toast — success, "already up to date", and failure:

```js
} catch ( err ) {
    if ( isUpToDateError( err ) ) { host.toast( … ); }
    else { host.toast( sprintf( __( 'Update of %1$s failed: %2$s' … ) ), 6000 ); }
}
```

A request that failed, 403'd, timed out at the server or returned non-JSON would all reject and toast. **Never toasting at all means the awaited promise never settles.**

## What that leaves, and why it cannot recover

**1. The single-flight queue has no timeout and no recovery.** `apps/plugins/parts/update-queue.ts`:

```js
const queue = []; let inFlight = false;
async function drain() { if ( inFlight ) return; … }
```

`inFlight` is module-scoped. A job whose `run()` never settles leaves it `true` for the life of the window; every later job is pushed onto a queue that never drains, and every caller's promise stays pending. There is no `AbortSignal`, no timeout and no wedge detection anywhere in the request layer — `parts/rest.ts`, `parts/actions.ts`, `parts/update-queue.ts` and `plugins.os.ts` contain zero occurrences of `AbortController`, `AbortSignal` or a request timeout.

**2. The row then locks itself out, silently.** `host.busy.updating` is cleared only in the `finally` of a settled promise, so the guard at the top of `runUpdate()` turns that plugin's button into a permanent no-op for the rest of the session — with no toast and nothing logged. That is the "further clicks do nothing" half.

So whatever stalls the first request, the window has no path back except being closed and reopened.

## Steps to reproduce

1. Open the Plugins window and filter to **Update available**.
2. Click **Update** on a plugin.

**Expected:** busy state, then a success or failure toast.
**Actual:** busy state that never resolves; no toast; the row's button is dead from then on.

## Ruled out while narrowing this

Recorded so the search is not repeated:

- **Not the service worker.** Its fetch handler begins `if ( n.method !== 'GET' ) return;`, so the admin-ajax POST is never intercepted.
- **Not a stale nonce.** Reproduces on a freshly reloaded shell where `updatesNonce` is minted by `wp_create_nonce( 'updates' )` — and a rejected nonce would toast a failure.
- **Not `stackOnPhone()`,** and not phone rendering generally: it only toggles attributes, never clones nodes, and the bug also occurs on a desk.
- **Not the click wiring or `repaint()`** — the busy indicator appearing proves both work.

## Note on the environment

I cannot yet say what stalls the underlying POST, and it may not be this project's doing — the site involved has an unrelated open question about requests being held at its CDN edge, which I am investigating separately. I have deliberately not asserted a cause here.

The two defects above stand on their own either way: without a timeout, *any* stalled request — whatever its origin — degrades the window to "the button does nothing, forever", with no feedback and no recovery.

## A suggestion, offered lightly

A timeout around the queued job, rejecting rather than hanging, would route every variant of this into the failure toast that already exists and clear `busy.updating` through the existing `finally` — turning a permanent silent wedge into a visible, retryable error, without needing to find the original stall first.

Pranjal1423:
Fixes #765

When a plugin update request stalls indefinitely (e.g. server-side hang, network freeze, or unresponsive proxy), the promise never resolves or rejects. Because the plugins app uses a serial update queue (`updateQueue`), a hanging promise causes the queue to remain locked forever, leaving the plugin row permanently in a busy state and preventing any subsequent updates (both single and bulk) from running.

This PR adds an execution timeout (`DEFAULT_TIMEOUT_MS = 60_000`) to queued update operations via `Promise.race()`. If an update promise fails to resolve within the timeout period, it is rejected with an `Update timeout` error. This allows the existing `try/catch/finally` error handling in `runUpdate()` and `runBulkUpdate()` to catch the error, show a failure notice, release the busy state on the row, and allow subsequent queued updates to proceed.

---

### Changes

- **`apps/plugins/parts/update-queue.ts`**: Wrapped the task invocation in `drain()` with `Promise.race()` against a configurable timeout timer (defaulting to 60s), ensuring the timeout timer is always cleaned up on completion.
- **`tests/vitest/plugins-update-queue.test.ts`**: Added comprehensive unit test coverage verifying:
  - Updates completing before the timeout resolve normally.
  - Stalled updates reject with a timeout error after the configured timeout.
  - Subsequent updates in the queue continue executing after a preceding task times out.
  - Direct task rejections before timeout propagate cleanly without delay.

---

### How to test

1. Open the Plugins app in OpenStation.
2. Trigger a plugin update under conditions where the request stalls or times out.
3. Verify that after the timeout period:
   - A failure notice/toast is displayed indicating the update failed.
   - The busy indicator on the plugin row clears and the row is unlocked.
   - Subsequent plugin updates can be clicked and executed without needing to reload the app.
4. Run automated tests: `npm run test:js tests/vitest/plugins-update-queue.test.ts`

---

### Before the Fix 

```
$ npm run test:js tests/vitest/plugins-update-queue.test.ts

> openstation@1.1.6 test:js
> vitest run tests/vitest/plugins-update-queue.test.ts

 RUN  v4.1.5 /Users/pranjalwadhwani/Desktop/desktop-mode
 tests/vitest/plugins-update-queue.test.ts (7 tests | 7 failed) 6ms
     × has a default timeout of 60 seconds 3ms
     × resolves normally when the job completes before timeout 0ms
     × rejects with custom error when the job fails normally 0ms
     × rejects with a timeout error if a job stalls and does not settle 0ms
     × unblocks subsequent queued jobs after a timed-out job 0ms
     × executes jobs in strict FIFO order 0ms
     × recovers from a stalled update: shows failure toast, clears busy state, and re-enables button 0ms

⎯⎯⎯⎯⎯⎯⎯ Failed Tests 7 ⎯⎯⎯⎯⎯⎯⎯

 Test Files  1 failed (1)
      Tests  7 failed (7)
   Duration  1.15s
```
---

### After the Fix 

```text

$ npm run test:js tests/vitest/plugins-update-queue.test.ts

> openstation@1.1.6 test:js
> vitest run tests/vitest/plugins-update-queue.test.ts

 RUN  v4.1.5 /Users/pranjalwadhwani/Desktop/desktop-mode

 ✓ tests/vitest/plugins-update-queue.test.ts (7 tests) 13ms
   ✓ Plugins update queue (enqueueUpdateJob)
     ✓ has a default timeout of 60 seconds
     ✓ resolves normally when the job completes before timeout
     ✓ rejects with custom error when the job fails normally
     ✓ rejects with a timeout error if a job stalls and does not settle
     ✓ unblocks subsequent queued jobs after a timed-out job
     ✓ executes jobs in strict FIFO order
   ✓ Plugins update action with timeout integration
     ✓ recovers from a stalled update: shows failure toast, clears busy state, and re-enables button

 Test Files  1 passed (1)
      Tests  7 passed (7)
   Duration  1.17s
```

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-776-0e1888355182c8b659576a907b7d7ee002768d7a.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper (approved):
Thanks!

AllTerrainDeveloper:
Sorry for the delay getting back to this—we’ve had a lot of work on our plate. Thanks for the contribution! We’re picking this up now and will work through the failing checks.

Pranjal1423:
No worries at all
Thankyou so much !!

PR #777, opened 2026-09-08
Files: add uploaded files to the Media Library, and start a post or page from an image

AllTerrainDeveloper:
## What

A file dragged onto the desktop lives in desktop storage, outside the Media Library on purpose. This PR is the bridge back, from the tile's right-click menu:

- **Add to Media Library** — on any stored file the site would accept as an upload. The "is this media" check is WordPress's own upload allow-list, decided server-side, filterable via `openstation_stored_file_is_media`, and shipped as `file.isMedia` on the tile. Multi-select aware (one call per file, one toast for the batch). Requires `upload_files`.
- **Start a post with this image** / **Start a page with this image** — on image tiles, each gated by the post type's `create_posts` capability. The server copies the image into the Media Library, inserts an `auto-draft` (exactly what `post-new.php` creates, so an abandoned start is swept with every other auto-draft) with the image block as content and the image as featured image, and the block editor opens in a window.

## How

- The copy is **idempotent per stored file**: the attachment records its source in `_openstation_stored_file_id` and `_openstation_stored_file_key` (the row's disk-name UUID) post meta. "Add" twice, or "Add" then "Start a post", returns the existing attachment. Deleting the attachment lets the file be added again. The stored bytes are never moved; tile and attachment stay independent afterwards.
- Server: `includes/desktop-files/media-library.php` — policy, copy (`media_handle_sideload` on a scratch copy), start-post, `POST /files/uploads/<id>/media` and `POST /files/uploads/<id>/post`, and `desktopStorage.canAddToMedia` / `canStartPost` / `canStartPage` shell-config flags so the tile never offers what the server would 403. Non-readable files answer the same masked 404 as downloads.
- Client: `src/desktop-files/media-menu-items.ts` contributes the entries through the public `os.files.tile-menu` filter; `openUrlWindow()` on the open dispatcher opens the resulting edit screen the way a `url` opener would.
- The server helper also handles video, audio and generic files (video / audio / file blocks); the tile only offers the post entries for images.

## Hooks

Filters: `openstation_stored_file_is_media`, `openstation_stored_file_media_post_data`, `openstation_stored_file_start_post_content`, `openstation_stored_file_start_post_args`. Actions: `openstation_stored_file_added_to_media`, `openstation_stored_file_post_started`. All documented in `hooks-reference.md`, `javascript-reference.md`, `files-on-desktop.md` (new "Into the Media Library" section) and the stored-files example.

## Tests

- PHPUnit: new `Tests_OpenStation_MediaLibrary` (policy, filename derivation, copy + source intact, idempotency incl. the reissued-id case, masking, non-media rejection, auto-draft content/thumbnail, page type, filters, capability failures, route registration, both REST callbacks, permission gate, config flags); `isMedia` assertion added to the upload-shape test.
- Vitest: new `desktop-files-media-menu.test.ts` (entry set per kind/flags, legacy payloads, click paths, bulk toast, error toasts).
- `npm run build`, `lint`, `typecheck`, `test:js` (5910), `lint:php`, `test:php` (2948) all green.

## Manual QA

Drag a photo onto the desktop, right-click → Add twice (second toast says "Already in the Media Library"), then Start a post and confirm the editor opens with the image and featured image set. A `.txt` gets Add only; select several images for the bulk Add.

## Also: Spanish translation guesses removed

The workspace wizard's primary button read "Ese archivo no es una imagen." in Spanish. `msgmerge --update` in `bin/extract-i18n.sh` ran with fuzzy matching on, so every POT refresh paired new source strings with the closest old translation and flagged them `#, fuzzy` (527 of them, "Create workspace" included), and `wp i18n make-json` compiles fuzzy entries like real ones. Second commit: the fuzzy guesses are blanked in the PO (`msgattrib --clear-fuzzy --empty`; the 62 genuine translations stay), the openstation JSON and the MO are rebuilt from it, and the extract script now passes `--no-fuzzy-matching` so this cannot recur. Blanked strings fall back to English exactly as untranslated ones always did.

## Also: drag an upload into the block editor, or onto a post tile

Third commit. Two ways to put a stored desktop file into a post by dragging it:

- **Into the block editor in a window.** A media upload tile lifts with a new `upload` bridge payload. Since a stored file has no attachment id or URL, the bridge gains a resolver registry (`registerBridgePayloadResolver` / `resolveBridgePayload`) and the iframe drop target resolves the payload when the drop lands: copy into the Media Library (idempotent), then the editor receives the `attachment` payload it already knows. Resolution runs at drop, never at lift. While the drag is over the editor the shell streams the pointer into the iframe (`os-drag-move`, once per frame) and the receiver draws Gutenberg's own insertion line at the spot the drop would land (`showInsertionPoint`, computed by `src/gutenberg-insertion-point.ts`: innermost block under the pointer, midpoint split, nested lists, empty space); `os-drop` inserts at that index. The lookup avoids `instanceof HTMLElement` on purpose: the editor canvas is an iframe in another realm.
- **Onto a post or page tile** (wallpaper or folder window). The `desktop-file` payload is opted in through the tile-payload seam when every dragged tile is a media upload ("Add to post" / "Add to page"). `POST /files/posts/<id>/uploads` copies each file, appends it as a block in drop order, attaches it to the post, and makes the first image the featured image when the post has none; the post then opens in a window. Filter `openstation_stored_file_attach_content`, action `openstation_stored_file_attached_to_post`.

Tests: PHPUnit for the attach route (append order, featured image and parent rules, filter, refusals, REST shape); vitest for the upload payload and resolver, the iframe target waiting on resolution, the tile handlers, the `os-drag-move` stream on both paths, and the insertion-point lookup including a canvas-iframe realm case.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-777-04537f41dda8289b72144b69087bfc3902a98788.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
@claude can you review?

PR #779, opened 2026-09-09
Redesign native Plugins, Posts, Pages, and Users workspaces

AllTerrainDeveloper:
Native Plugins, Posts, Pages, and Users now have purpose-built workspaces instead of tables as their default view. Actions stay beside the item, selection never shifts the cards, and content continues loading as you scroll.

## Changes

- **Plugins:** grouped library cards, persistent selection checkboxes, a contextual bulk-action tray, a details inspector, and concise Update buttons.
- **Posts and Pages:** notebook and page-sheet cards, content metrics, copyable IDs, shared details, compact responsive layouts, and continuous scrolling. Categories and Tags gain a searchable topic directory and theme-aware canvas labels.
- **Page Atlas:** a separate spatial view of page hierarchy and content links, with routed connections and real frontend previews. Preview iframes retain a fixed 1440 × 900 viewport while CSS transforms scale them; at most six previews are mounted and the map reports its 500-page limit.
- **Users:** member ID cards with an accessible Actions dropdown, role groups, and an activity dashboard. Roles use a capability-gated, prepared SQL summary with complete counts and at most eight examples per role. Activity reads complete site-wide totals and bounded profile samples in one response. Role counts share one conditional aggregate scan, with bounded per-role sample queries.
- **Interaction and themes:** pinch-to-zoom for Page Atlas, Categories, and Tags; gesture cancellation prevents accidental edits. Native surfaces follow the current desktop palette, including live canvas recoloring on theme and accent changes. Atlas keeps visible preview frames mounted across pans, uses zoom hysteresis, and keeps connection routes stable when filters change.

The role and activity summary endpoints/filters, silent background REST requests, and the reusable `os-action-menu` component are documented alongside their tests.

## Validation

- Full `npm run build`, TypeScript checking and PHP lint pass. JavaScript lint has no errors (27 existing file-length warnings).
- Full JavaScript suite: 497 files, 6,057 tests passed.
- Full PHP suite: 2,964 tests / 22,602 assertions, 12 environment-specific skips. Focused multisite Posts, Activity and Roles: 29 tests / 236 assertions passed, including site isolation and SQL metacharacters.
- Regression coverage includes stale-query responses, refresh preserving the loaded prefix, additive plugin selection, bounded metric retries and silent tracking, palette alpha and live updates, stable preview residency, and tag drag cancellation.
- Browser verification on an isolated WordPress site: Activity displays all 46 seeded users and all 45 published posts; Posts Refresh preserves scroll position and the loaded range; visible-card metrics show real counts. Existing dark-mode screenshots are below. A physical multitouch-device check remains outstanding.

Activity returns a bounded response, but its server-side reduction and temporary memory still grow with site membership. Role samples retain per-role lookups. Frontend preview iframes are passive same-origin site content, not a security isolation boundary.

## Desktop screenshots — default dark theme

All captures use the default OpenStation dark theme at 1600 × 1000. Page Atlas previews preserve the actual frontend theme, which may be light. Unchanged creation/profile/marketplace tabs are outside the redesign.

<details>
<summary>Plugins · Installed</summary>

Plugins · Installed — dark desktop

</details>

<details>
<summary>Posts · All posts</summary>

Posts · All posts — dark desktop

</details>

<details>
<summary>Posts · Categories</summary>

Posts · Categories — dark desktop

</details>

<details>
<summary>Posts · Tags</summary>

Posts · Tags — dark desktop

</details>

<details>
<summary>Pages · All pages</summary>

Pages · All pages — dark desktop

</details>

<details>
<summary>Pages · Page Atlas</summary>

Pages · Page Atlas — dark desktop

</details>

<details>
<summary>Pages · Page Atlas, focused connections</summary>

Pages · Page Atlas, focused connections — dark desktop

</details>

<details>
<summary>Users · People and Actions</summary>

Users · People and Actions — dark desktop

</details>

<details>
<summary>Users · Roles</summary>

Users · Roles — dark desktop

</details>

<details>
<summary>Users · Activity</summary>

Users · Activity — dark desktop

</details>

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-779-f3c8b71c575497954215a4e0e8c33bb1c5c79edd.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


epeicher:
Reviewed the branch and exercised it on a seeded wp-env site (30 users across roles, 31 posts, a linked page hierarchy) plus a SQLite round trip. Three bugs reproduced; I pushed fixes for those and the doc nits as four commits on top of aa66c1c. The rest of what I found is listed below for the author to weigh.

## Pushed to this branch

**ddfc9bd fix(users): keep the role summary query portable to SQLite.** The roles summary built `(SELECT … ORDER BY … LIMIT 8) UNION ALL (…)`. MySQL accepts that; SQLite does not, so on Playground, Studio and any site running the SQLite integration plugin (checked with 3.0.1 on both its translator and its AST driver) the route answered `SQLSTATE[HY000]: General error: 1 near "(": syntax error` and the Roles tab showed "Role groups are unavailable". This also breaks the Playground preview button on this PR. Each sample now rides in a derived table, `SELECT * FROM (… LIMIT 8) AS sample_n`, which both engines parse and which keeps the bound on MySQL. Still one statement; the PHPUnit test additionally pins that every UNION member starts with `SELECT`. After the change the same SQLite setup returns 200 with correct totals, including a `shop_manager` role (the `esc_like` underscore escape survives SQLite too).

**53d3fb6 fix(plugins): keep the library search field in step with its state.** "Show all plugins" cleared `state.search` and repainted, but the search field had no `value` binding, so every card came back while the field still read the stale query and the next keystroke filtered on that text plus the new character. Bound `value` to the state, as the users app already does; the test checks the field before and after a reset.

**faeb6da feat(users): make the role filter server state.** The People tab's role filter was a client-side slice over whatever pages the feed had loaded, and the Roles tab's "Meet this group" handed off to it. Picking the No role group on the seeded site landed on "0 match filters" and "No people match this view", because its one member sat on page 2, and only appeared after the feed auto-paged. On a large site with a small role that sorts late that is many sequential round trips before anyone shows. `role` is now a declared state key the server answers through the Core users collection: a slug becomes `roles`, and `none` (the way `users.php?role=none` spells it) becomes an `include` of `wp_get_users_with_no_role()`, or nobody. The select binds through `os-bind` + `filter`, "Meet this group" dispatches it, the feed treats a role change as a fresh collection, and Activity clears the role scope the way it already cleared the search. Verified in the browser: No role now shows "1 of 1 people loaded" immediately, Editor loads "12 of 12", Activity drops the role and completes. New PHPUnit test covers a slug, `none` with and without role-less accounts, and multisite membership; hooks-reference updated.

**506d603 docs: tidy the workspace notes and the action-menu row.** The `<os-action-menu>` row in the components reference had three cells in a four-column table. The Users role summary and content-card notes were appended under the roadmap's North Star section at the end of architecture.md; they now have their own "Native workspaces" section beside the other shipped surfaces. The pages test that asserts two tabs exist no longer claims to have none.

Verification on the final tree: `npm run build`, `lint` (0 errors, the 27 pre-existing file-length warnings), `typecheck`, `test:js` (6,051 tests), `lint:php` clean, PHPUnit `Tests_OpenStation_Users(App|RolesSummary)` on single site and multisite.

## Other findings, left for the author

Worth a follow-up:

- **Activity gathers every user sequentially at 20 per page** (`apps/users/users.os.php` `perPage`) and runs the grouped stats queries per page. A few thousand users means hundreds of round trips before the tab paints anything. An aggregate endpoint like the roles summary, or a cap with an explicit "partial" state, would scale better.
- **The roles summary is 2N+1 scans of `wp_users`**, each with an EXISTS plus LIKE over usermeta; Core's `count_users()` does the counts in one pass. Fine at this scale, worth knowing for large user tables.
- **Posts desk metrics drive the window spinner and the activity bus.** `desk-stats.ts` fetches through `ctx.fetch`, whose `restFetch` wrapper (`session.ts`) never passes `silent`. Two requests per card refill as you scroll, so the spinner flickers and the activity widget logs traffic the user never initiated. AGENTS.md reserves `silent: true` for exactly this.
- **A refresh or `post` watch after scrolling collapses the feed to page 1 with two round trips** (`content-feed.ts`). Reproduced: at 31 cards, Refresh returned page 2, the feed re-requested page 1, the desk dropped to 20 cards and the scroll position was lost. Any post edit in another window triggers the same reset. If that restart dispatch rejects, `error` is never set, so "Load more" stays offered from stale paging.
- **Atlas threads do not recolor on theme change.** `atlas-scene.ts` reads the accent once and has no `os-desktop-theme-changed` listener, unlike the term canvases. The term canvases also ignore the accent picker and read `themeHue` once. The PR body's "live canvas recoloring" only holds for a desktop theme switch on Categories and Tags.
- **Atlas previews mount six full frontend documents at any zoom above 0.12.** At 35% I saw six live iframes while two sheets still read "Zoom closer for a live preview". Eviction re-ranks by distance on every pan pause (160 ms debounce), so nudging across a large map reloads the same pages. Hysteresis (keep mounted frames that are still on screen, evict only over the cap or off-screen) would stop the churn.
- **Tags canvas drops a drag without restoring position** when a pinch cancels it (`tags-cloud.ts` `cancelGesture`); the mind map's version restores the pinned position.
- **"Select updates" replaces the current selection** instead of adding to it (`installed-library.ts`), so ticked deletions vanish silently.
- **Suspected race in `content-feed.ts`:** typing a filter while a page load is in flight can merge the old query's page under the new key; the users feed guards this with `page !== requested`, the posts one does not. Not reproduced.
- **A failed metric never retries** in `desk-stats.ts` (`pending` keeps the key), so "—" persists until the post's modified date changes and looks identical to a 403.

Smaller:

- `sandbox="allow-scripts allow-same-origin"` on atlas previews is not a boundary (same-origin scripts can drop their own sandbox); acceptable since it is the site's own frontend, but the test asserting no `allow-top-navigation` gives false confidence. `inert`, `tabindex=-1` and `pointer-events: none` do the real work, and those are present.
- `palette.ts` discards alpha when reading tokens; harmless today because every token it reads is an opaque hex, fragile for a future theme.
- `desk.ts` `inspector()` runs on every repaint even when closed, serialising the first row each time.
- `atlas-scene.ts` lane offsets use the edge's index in the filtered list, so routes shift when the connection filter changes. Cosmetic.

Everything else checked out: roles payload exact for multi-role, no-role and empty-role cases with no emails in samples; People feed pagination, dedupe, selection tray and details table; the Actions popover escapes window clipping and restores focus; Posts infinite scroll and the ID copy button; atlas iframes fixed at 1440×900, inert, aria-hidden, no admin bar inside; plugin capability gating tighter than before; every new colour resolves through the palette tokens; no raw `fetch` or `window.confirm`.

PR #785, opened 2026-09-09
Harden stored-file cleanup and make presence updates atomic

AllTerrainDeveloper:
Stored-file reconciliation could mistake a failed database lookup for an empty result and delete valid upload bytes. Presence heartbeats and pruning also rewrote one shared option, allowing concurrent requests to overwrite other users' updates.

This change hardens those two persistence paths:

- Revalidate cleanup candidates under the upload writer lock, use current reads, and require successful database deletion before removing bytes. Failed database/lock checks stop cleanup. Individual unlink failures are reported while later candidates continue; successful row deletions still emit their action. Trashed placements and the grace period continue to protect files.
- Store presence in one primary-keyed row per user with atomic timestamp merges and indexed pruning. An away-intent fence prevents delayed activity from undoing a newer away request. Public response shapes, timestamp units, thresholds, and throttling remain unchanged.
- Keep authorization, collision recovery and upload creation/placement callbacks outside the storage lock. Failed share placement returns its error, and acceptance can be retried to repair a missing tile.
- Reuse presence reads within each request, invalidate after writes/pruning, avoid migration work on ordinary front-end requests, and attempt failed setup only once per request with a nonblocking migration lock. The tracking veto runs once; recorded transitions use the stored fence after a write.

### Existing installations and compatibility

Presence creates a new per-site table and imports the legacy option before publishing a verified checkpoint. Failed setup retains the legacy path and retries on the next request; that fallback retains the original contention until migration succeeds. The five-minute bridge skips unchanged legacy maps and cannot turn old idle heartbeats into fresh away intent. The legacy option remains available for rollback; recovery instructions are in `docs/migration-presence-storage.md`. Subsite deletion includes the new table.

SQLite's `GET_LOCK` compatibility function returns `1=1` without acquiring a lock. Uploads and presence work with that result. Destructive orphan reconciliation deliberately stops safely on SQLite, retaining candidates rather than claiming mutual exclusion. MySQL/MariaDB retain the coordinated cleanup path.

HyperDB/LudicrousDB use their supported primary-routing APIs for protected operations. Routing-adapter tests cover API selection; an actual replicated or active-active deployment has not been integration-tested.

Future presence column changes still require a versioned schema migration. The current `CREATE TABLE IF NOT EXISTS` path handles this initial schema only; a general `dbDelta` upgrade framework is deferred and documented.

Uninstall policy, sharing columns, existing timestamp formats, file references, and agent discovery are unchanged.

### Validation

- Full PHPUnit suites passed: 3,005 tests each in single-site and multisite modes (22,822 / 22,831 assertions; 13 / 12 skips).
- Failure injection covers cleanup queries/locks and migration creation, import, verification, and checkpoint writes. New regressions cover unlink starvation, deletion callbacks, unlocked upload callbacks, retrying share acceptance, request query count, failed-install memoization, stateful vetoes, legacy idle activity, and fenced transitions.
- Independent PHP/database workers exercise concurrent presence updates, heartbeat/pruning races, and upload/cleanup locking. A transaction-snapshot test verifies valid bytes survive a newly committed registration or a database conflict.
- SQLite driver smoke test verified its no-op lock result, presence schema, monotonic merges, and away fence. PHPUnit covers the shim result at the plugin's intake/cleanup boundaries; a full Playground browser run was not performed.
- JavaScript suite passed: 6,073 tests. Build, PHP lint, JavaScript lint, type checking, and diff whitespace checks passed. JavaScript lint reports 27 existing file-length warnings and no errors.
- REST dispatch smoke checks on an isolated WordPress instance verified online → away → online, status 200 responses, and unchanged snapshot fields. Earlier authenticated HTTP checks covered the original implementation.

Both PHP suites exit successfully with a Core `REQUEST_URI` warning during shutdown. The original development checkout and database were left untouched.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-785-17c71881dc3387c7dea8afcb37e0a51a3d7ed02e.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/desktop-files/reconcile.php
```diff
@@ -0,0 +1,217 @@
...
+ *
+ * @internal
+ * @param callable $callback Operation to protect.
+ * @return mixed|WP_Error Callback result, or a retryable lock error.
+ */
+function openstation_stored_files_locked( $callback ) {
+	global $wpdb;
+	$name = 'os-files-' . md5( $wpdb->dbname . ':' . $wpdb->prefix );
+	$lock = $wpdb->get_var( $wpdb->prepare( 'SELECT GET_LOCK(%s, 5)', $name ) );
+	if ( '1' !== (string) $lock ) {
```
epeicher, line 34: **This check fails on every call under SQLite, which breaks Playground.** I ran the PR's statements through both drivers of sqlite-database-integration: `GET_LOCK` returns the string `1=1` on both, so `'1' !== (string) $lock` is always true. On the wp.org Live Preview blueprint and the PR preview workflow that means every upload and every upload placement returns 503 `openstation_storage_busy`, the daily sweep aborts at its first candidate with stage `delete_row`, and presence never migrates because the same comparison in `openstation_presence_migrate_storage()` fails too (so it never prunes there either).

Everything else in the PR's SQL behaves correctly on SQLite: `FOR UPDATE`, `ON DUPLICATE KEY UPDATE ... GREATEST(...)` and the aliased multi-table `DELETE` all parse and return the right rows. Treating the result as acquired unless it is exactly `'0'` or `NULL` matches MySQL's own return values and fixes both call sites. CI runs on MariaDB, so nothing catches this today.

includes/desktop-files/reconcile.php
```diff
@@ -0,0 +1,217 @@
...
+	);
+	if ( '' !== $wpdb->last_error ) {
+		return false;
+	}
+	clearstatcache( true, $entry );
+	if ( null === $known && is_file( $entry ) && ! is_link( $entry ) ) {
+		$mtime = filemtime( $entry );
+		if ( false !== $mtime && $mtime > 0 && $mtime < $cutoff ) {
+			wp_delete_file( $entry );
+			return ! is_file( $entry );
```
epeicher: **One undeletable file now aborts every daily sweep, forever.** When `wp_delete_file()` fails here this returns false, the caller fires `delete_bytes` and returns. Tomorrow the same glob order reaches the same file at the same position, so every entry after it is never cleaned. The pre-PR sweep skipped it and continued.

The abort policy is right for database errors, where the state cannot be trusted, but a filesystem permission error on one file is not that. Suggest: report the failure through the new action and `continue` on unlink failure; abort only on SQL errors. Worth a test with an unlinkable entry followed by a deletable one.

includes/desktop-files/reconcile.php
```diff
@@ -0,0 +1,217 @@
...
+	);
+	if ( false === $deleted ) {
+		return false;
+	}
+	if ( 1 === $deleted ) {
+		$row  = openstation_stored_files_normalize_row( $row );
+		$path = openstation_stored_file_path( $row );
+		if ( $path && is_file( $path ) ) {
+			wp_delete_file( $path );
+			if ( is_file( $path ) ) {
```
epeicher, line 140: Same class as the bytes phase, with one extra consequence: the row is already deleted by the time this unlink fails, so the sweep aborts and `openstation_stored_file_deleted` never fires for a row that is gone. Since the DELETE succeeded, fire the action regardless of the unlink outcome and let the bytes phase pick the file up on a later run.

includes/migrations.php
```diff
@@ -662,3 +662,6 @@ function openstation_migrate_delete_ai_keys() {
 		openstation_save_os_settings( (int) $user_id, $raw );
 	}
 }
+
+// Presence owns a verified checkpoint so failures never advance unrelated migrations.
+add_action( 'init', 'openstation_presence_migration_tick', 20 );
```
epeicher: **This runs on every request, front-end and anonymous included, and reads a non-autoloaded option.** On a site without a persistent object cache that is one extra `SELECT` per page load, permanently, long after the migration has completed. The sibling migration runner deliberately hooks `admin_init`. The checkpoint is a two-key array, so autoloading it removes the query; alternatively hook the tick where presence already runs (Heartbeat, the REST route).

includes/presence-store.php
```diff
@@ -0,0 +1,244 @@
...
+ */
+function openstation_presence_migrate_storage() {
+	global $wpdb;
+	$state = get_option( OPENSTATION_PRESENCE_STORAGE_OPTION, array() );
+	if ( ! empty( $state['ready'] ) ) {
+		return true;
+	}
+	$table = openstation_presence_table();
+	$name  = 'os-presence-' . md5( $wpdb->dbname . ':' . $table );
+	if ( '1' !== (string) $wpdb->get_var( $wpdb->prepare( 'SELECT GET_LOCK(%s, 5)', $name ) ) ) {
```
epeicher: **Nothing memoizes a failed installation within a request, and every reader waits on the lock.** In the fallback state this PR explicitly supports and tests (CREATE denied, lock unavailable), one heartbeat reaches this function about five times: the init tick, the read in `record()`, the write, `get_all()` and again inside `snapshot()`. Each call does a `GET_LOCK`, two cache deletes including the site-wide `notoptions` key (which on persistent-cache sites evicts the negative-option cache for every option), a failing `CREATE TABLE` that wpdb logs through `print_error()`, and a `RELEASE_LOCK`. A per-request static that remembers the failure, plus `suppress_errors()` around the DDL, fixes it.

Two smaller points on the same line. During the one-time import every concurrent request, anonymous ones included, waits up to 5 s here; readers could use a zero timeout and fall back immediately, leaving the wait to the writer. And this block is a second copy of `openstation_stored_files_locked()` with already-diverging busy semantics (`false` here, `WP_Error` there); a shared helper in a file both load would keep the protocol in one place.

epeicher (commented):
Review of 2cf94c25. The design is sound: row before bytes, current-snapshot `FOR UPDATE` checks, monotonic timestamp merges, a verified checkpoint, and a real multi-process concurrency test. CI is green on both PHP versions.

Two things block merging as-is: the lock check fails on every call under SQLite, which breaks uploads and presence on Playground, and the daily sweep now aborts permanently on one undeletable file. The rest are per-request cost regressions on hot paths, a bridge edge case, and a few design notes. Details inline.

One question I could not verify locally: under HyperDB / LudicrousDB, `SELECT GET_LOCK(...)` and `SELECT ... FOR UPDATE` are `SELECT`s and may be routed to a replica connection, where the lock would not be shared with the writes on the primary. Worth confirming they are classified as write queries there.

includes/presence.php
```diff
@@ -60,46 +59,28 @@
...
-		if ( $uid <= 0 || ! is_array( $record ) ) {
-			continue;
-		}
-		$out[ $uid ] = array(
-			'last_seen_ms'   => isset( $record['last_seen_ms'] ) ? (int) $record['last_seen_ms'] : 0,
-			'last_active_ms' => isset( $record['last_active_ms'] ) ? (int) $record['last_active_ms'] : 0,
-		);
-	}
-	return $out;
+	$records = openstation_presence_read_records();
```
epeicher, line 70: **Every read is now a table query; the heartbeat and the Users app regress in query count.** This used to be a cached `get_option()`, so reads after the first were free and, with a persistent object cache, zero. Now each heartbeat issues three `SELECT`s, two of them full scans of this table, because `openstation_presence_heartbeat_received()` calls `get_all()` and `snapshot()` calls it again. The Users REST field calls `openstation_presence_status_for_user()` per row, so a page of 20 users is 20 queries. A request-level cache in `openstation_presence_read_records()`, invalidated by `write_record()`, restores the old cost; passing the already-read map into `snapshot()` removes the duplicate scan.

includes/desktop-files/store.php
```diff
@@ -36,6 +36,40 @@
  * @return int|WP_Error Placement id on success, `WP_Error` otherwise.
  */
 function openstation_files_place( $user_id, $parent_id, $type, $ref, $args = array() ) {
+	if ( 'upload' === (string) $type ) {
+		return openstation_stored_files_locked(
```
epeicher: **Third-party hooks now run under a site-global 5 s lock.** `openstation_files_place_locked()` runs the `openstation_files_can_place` filter, the share-capability lookups, collision recovery and the `openstation_file_placed` action inside the lock. One slow listener, say a synchronous thumbnail job on `openstation_file_placed`, makes every other upload and placement site-wide return 503 `openstation_storage_busy` until it finishes.

Two downstream effects: `openstation_stored_file_share_accept()` ignores the placement result, so a busy error leaves the share `accepted` with no tile and no way to re-accept it. And the upload route takes the lock twice per upload (create, then place), so a timeout at placement discards bytes already transferred. Suggest running `can_place` before taking the lock, holding the lock only around the `FOR UPDATE` check and the INSERT, and firing the action after release.

Also: this pushes `store.php` to 1,007 lines, past the file-length warning; the wrapper could live in `reconcile.php` next to the lock.

includes/desktop-files/stored-files-store.php
```diff
@@ -164,6 +164,22 @@ function openstation_stored_file_path( $row ) {
  * @return int|WP_Error Row id.
  */
 function openstation_stored_files_create( $owner_id, $args ) {
+	return openstation_stored_files_locked(
```
epeicher: `openstation_stored_file_created` now fires inside the lock too (see the note on `openstation_files_place()`). The lock around create only matters for bytes registered with an mtime older than the one-day grace, which the standard intake never produces; if that path is wanted for third-party integrations, keep the lock but fire the action after the callback returns.

includes/presence-store.php
```diff
@@ -0,0 +1,244 @@
...
+	if ( $cut <= 0 || $now - $cut > 5 * MINUTE_IN_SECONDS * 1000 ) {
+		return;
+	}
+	$records = openstation_presence_legacy_records();
+	if ( is_wp_error( $records ) ) {
+		return;
+	}
+	foreach ( $records as $uid => $record ) {
+		if ( $record['last_seen_ms'] > $cut ) {
+			openstation_presence_upsert( $uid, $record, 0 === $record['last_active_ms'] );
```
epeicher: **The bridge can hide real activity, and re-upserts every late record on every request.** A legacy record with zero `last_active_ms` is imported as an explicit away, and old-code idle heartbeats keep the stored zero. So: a user at zero in the legacy option becomes active on new code at T1 (online), then one old-code idle heartbeat lands at T2 > T1; this tick sets `inactive_at_ms = T2` and the read fence hides the T1 activity until the user's next active tick. There is also no high-water mark, so the same records are upserted again on every request for five minutes.

Two options: treat zero as away only in the initial import and pass `false` here, storing the max imported `last_seen_ms` in the checkpoint to skip already-imported records; or drop the bridge. A plugin update replaces files in place, so old workers living for minutes is a fleet-deploy shape, and the migration doc now imposes that five-minute window on every install.

includes/presence-store.php
```diff
@@ -0,0 +1,244 @@
...
+		// A worker that waited for the lock must see the winner's checkpoint.
+		wp_cache_delete( OPENSTATION_PRESENCE_STORAGE_OPTION, 'options' );
+		wp_cache_delete( 'notoptions', 'options' );
+		$state = get_option( OPENSTATION_PRESENCE_STORAGE_OPTION, array() );
+		if ( ! empty( $state['ready'] ) ) {
+			return true;
+		}
+		$collate = $wpdb->get_charset_collate();
+		$created = $wpdb->query(
+			"CREATE TABLE IF NOT EXISTS $table (
```
epeicher, line 156: **No upgrade path for this schema.** `CREATE TABLE IF NOT EXISTS` never alters an existing table, and the verification query below names all four columns. The first column added later leaves existing installs on the old shape, the verification fails, and `migrate_storage()` returns false forever: presence falls back to the legacy option, pruning stops, no notice, and the migration doc says not to drop the table. The repo already has the versioned `dbDelta` pattern in `includes/desktop-files/schema.php`; adopting it now, while the table is new, is cheap.

includes/presence.php
```diff
@@ -538,7 +520,9 @@ function openstation_presence_rest_post( WP_REST_Request $request ) {
 		}
 	} else {
 		$flag = ( null === $active ) ? true : (bool) $active;
-		openstation_presence_record( $user_id, $flag );
+		if ( ! openstation_presence_record( $user_id, $flag ) && apply_filters( 'openstation_presence_can_track', true, $user_id ) ) {
```
epeicher: **The veto filter runs twice per POST, and the second answer decides the status code.** This re-applies `openstation_presence_can_track` to tell a veto from a storage failure. A filter whose answer depends on state that can change between the two calls (an appear-invisible toggle read from user meta or a transient) turns a veto into a 503, or a storage failure into `ok: true`, the exact outcome this PR set out to remove. Side-effecting filters also fire twice on the veto path. Suggest an internal record variant returning `true|WP_Error` with distinct codes for veto, read failure and write failure, and keep the public boolean wrapper on top.

includes/presence.php
```diff
@@ -146,8 +130,11 @@ function openstation_presence_record( $user_id, $active = true ) {
 	$next_status = openstation_presence_status_from_record( $next );
 
 	if ( openstation_presence_should_persist( $all, $user_id, $prev, $prev_status, $next_status, $active, $now_ms ) ) {
-		$all[ $user_id ] = $next;
-		update_option( OPENSTATION_PRESENCE_OPTION, $all, false );
+		$write                   = $next;
+		$write['last_active_ms'] = $active ? $now_ms : 0;
```
epeicher, line 146: Minor: `$next_status` is computed from `$now_ms` without a readback, while the stored fence uses strict `>`. A set-away and an active heartbeat in the same millisecond, or an active heartbeat from a node whose clock is behind the one that recorded the away, fires `openstation_presence_changed` with `online` while the stored record reads `inactive` until the clock passes `inactive_at_ms`. Low probability; worth a comment, or fire the change from the fenced readback.

docs/files-on-desktop.md
```diff
@@ -970,7 +970,7 @@ Uploaded files are **owner-locked**: only the stored file's owner may move, rena
...
+Reconciliation runs on the existing daily prune: placement-less rows and row-less bytes older than a day are removed in both directions. Trashed placements still protect their bytes. Every required lookup must succeed; a failed lookup aborts the remaining sweep. Candidates are revalidated under the same short, site-scoped database lock used by upload registration and placement creation. Final reference checks use current locking reads, so an older transaction snapshot cannot justify deleting newly registered bytes; a database conflict stops cleanup. Database-row deletion must succeed before bytes are removed. Failed unlinks leave row-less bytes for a later sweep. A lock failure returns a retryable `openstation_storage_busy` error (503) to writers; cleanup stops safely. Third-party upload integrations must use the store APIs rather than inserting rows directly to participate in this coordination. See [cleanup diagnostics](./examples/storage-cleanup-diagnostics.md). `deleted_user` purges the user's entire storage. Zip temp files are cleaned on stream end, shutdown, and by the daily sweep.
```
epeicher: The PHP surface list two lines below (Actions: `openstation_stored_file_{created,...}`) is not extended with `openstation_stored_files_reconcile_failed`; the hooks reference and the new example both have it.

tests/phpunit/tests/persistenceConcurrency.php
```diff
@@ -0,0 +1,187 @@
...
+			if ( is_resource( $worker[0] ) ) {
+				proc_terminate( $worker[0] );
+				foreach ( $worker[1] as $pipe ) { if ( is_resource( $pipe ) ) { fclose( $pipe ); } }
+				proc_close( $worker[0] );
+			}
+		}
+		mysqli_query( $this->db->dbh, 'DROP TABLE IF EXISTS ' . $this->prefix . 'openstation_presence' );
+		mysqli_query( $this->db->dbh, 'DROP TABLE IF EXISTS ' . $this->prefix . 'desktop_mode_stored_files' );
+		$this->db->close();
+		foreach ( glob( $this->dir . '/*' ) as $path ) { unlink( $path ); }
```
epeicher: `$this->dir` is assigned after the two `assertTrue( mysqli_query( ... ) )` calls in `set_up()`. If either fails, `tear_down()` runs `glob( '/*' )` and `unlink()` over the filesystem root. Harmless as the non-root user wp-env runs as, but a guard on `$this->dir` before the glob is cheap.

AllTerrainDeveloper:
@claude is all addressed?

PR #788, opened 2026-09-10
fix: separate hidden columns for posts and pages

### LINKED ISSUE #596: Posts and Pages silently share the Author and Date column choices

Pranjal1423:
### Description
Fixes #596.

Previously, Native Posts and Native Pages shared the single `nativePostsHiddenColumns` settings key. Hiding a column (such as Author or Date) in Posts would silently hide it in Pages, and vice versa.

This PR introduces an independent `nativePagesHiddenColumns` setting for Pages so that column visibility choices in Posts and Pages remain isolated and survive reloads independently.

### Changes
- **Backend**: Added `nativePagesHiddenColumns` default and sanitization in `includes/os-settings.php`.
- **Frontend Store**: Added `nativePagesHiddenColumns` to `OsSettingsState`, `DEFAULT_OS_SETTINGS`, sanitizers, and state clone helper (`src/settings/types.ts`, `src/settings/constants.ts`, `src/settings/state.ts`).
- **Client App**: Updated `apps/posts/parts/columns.ts` and `apps/posts/parts/app.ts` so `desktop-mode-pages` targets `nativePagesHiddenColumns` and `desktop-mode-posts` targets `nativePostsHiddenColumns` for initial read, menu toggles, and live subscriptions.
- **Tests**:
  - Added unit tests in `apps/pages/pages.test.ts` for Pages column toggle actions and external subscription changes.
  - Added backend tests in `tests/phpunit/tests/postsWindowSettings.php` for default value, sanitization, and DB round-trip.
  - Updated store partial save tests in `tests/vitest/os-settings-partial-save.test.ts`.
- **Docs**: Documented `nativePagesHiddenColumns` under `getOsSettings()` in `docs/javascript-reference.md`.

### Testing Instructions
1. Open both **Posts** and **Pages** windows.
2. In **Posts**, open `⋯` -> **Show columns** and uncheck **Author**.
3. Verify **Author** is hidden in Posts, but remains visible and checked in Pages.
4. In **Pages**, uncheck **Author** (or **Parent**).
5. Verify it is hidden in Pages without affecting Posts.
6. Reload the page and verify both windows retain their independent column visibility choices.

## Before Fix

https://github.com/user-attachments/assets/ac7d20a5-63c3-4a40-bf80-8e0fdf41b692

## After Fix

https://github.com/user-attachments/assets/af8387ee-c9f8-4067-996d-e2efd7257b3a

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-788-c5dd7536cc6391a19dc60d411009db4c35ce9183.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


epeicher:
Thanks for contributing to the repo @Pranjal1423!

The settings plumbing all looks right to me: the PHP default and sanitizer, the store types and clone helper, and both doc spots are covered. Full JS and PHP suites stay green too.

I gave it a spin locally with `nativePostsHiddenColumns: ['author']` and an empty `nativePagesHiddenColumns`, and the Pages window still came up without its Author column. I think one call site may have slipped through: `hiddenOf( ui )` inside `listPanel` (around line 386 of `apps/posts/parts/app.ts`) still falls back to the default key. Since `hiddenOf` memoises into `ui.hidden`, and `view` runs before `mounted`, that first read is the one that sticks for the life of the window. Toggling any column in Pages afterwards then writes the Posts set into `nativePagesHiddenColumns`, so it ends up persisted.

Passing `hiddenColumnsKey` there should do it. You might also consider dropping the default value on the `settingKey` parameter, so TypeScript points at any call site that forgets to pass one.

Three cases that would pin this down, if you fancy adding them to `apps/pages/pages.test.ts`:

- Pages paints Author when only Posts hid it
- The Pages column menu shows Author checked in that same state
- Toggling a Pages column writes only that column, not the Posts set

The new tests currently seed just the Pages key, so they never reach the path that breaks.

One last thing: the branch conflicts with trunk in `docs/javascript-reference.md`, where #795 rewrote the same `getOsSettings()` table row. Merging trunk in and keeping its version of that line, then adding `nativePagesHiddenColumns` after `nativePagesEnabled`, should clear it.

Thanks again for picking this one up!

AllTerrainDeveloper:
By the way, good catch! 😆 ❤️

Pranjal1423:
Thanks for the detailed review! I’ll fix the missed `hiddenOf(ui)` call site by passing `hiddenColumnsKey`, add the additional Pages test cases, and merge trunk to resolve the `docs/javascript-reference.md` conflict while keeping the updated version from #795.

I’ll push the changes shortly. Thanks  for catching this! 🙌

epeicher (approved):
Thanks @Pranjal1423! I have reviewed the changes, and they LGTM!

epeicher:
Heads up on the red CI here: it has nothing to do with your changes.

WordPress 7.1.1 was released a few hours ago, so the version-check API started advertising it before the `WordPress/WordPress` git mirror had the matching tag. Our wp-env configs track the latest release, so every job that boots wp-env died on `fatal: couldn't find remote ref 7.1.1`. That is exactly the five failures: PHPUnit and PHPUnit multisite on both PHP versions, plus Plugin Check. Lint / Types / Vitest and build need no wp-env, and they stayed green.

The tag has just landed upstream, so I have re-run the failed jobs and they should go green on their own. Nothing for you to do.

Your branch is in good shape otherwise, I re-reviewed it and everything from the earlier round is addressed.

Pranjal1423:
Thankyou so much :) !!

PR #792, opened 2026-09-10
os-text-field: add hide-label so a compact field can have a name without a visible label

### LINKED ISSUE #789: os-text-field: no way to expose an accessible name without rendering a visible label — four first-party search fields ship unnamed
> **Update, 2026-09-10 — the count I first reported was wrong, and a PR is open.** (Title corrected from seven to four.)
>
> **It is four on current `trunk`, not seven.** Two separate reasons:
>
> 1. `trunk` moved between my scan and this report; some of those call sites are already named or gone.
> 2. One of my seven was never real. My scanner matched the literal text `<os-text-field>` inside `os-number-field`'s help `summary` string and read it as a call site. `os-number-field` has its own `label` prop and was never affected — comment-stripping is not enough when the corpus contains code samples in strings.
>
> The four that remain are the search fields in **comments**, **plugins**, **trash** and **my-wordpress**.
>
> **The mechanism below is unchanged and still accurate** — the coupling between the visible label and the accessible name is exactly as described, and it is what the fix addresses.
>
> PR: #792 — adds `hide-label`, which keeps the `<label>` rendered, paired by `for=` and supplying the accessible name, while taking it out of the visual flow. The four call sites adopt it.
>
> Original report follows, kept for the record.

---

## Summary

`<os-text-field>` derives its visible label and its accessible name from the same `label` attribute. There is no third state, so a compact search field — the pattern used across OpenStation's own toolbars — has to choose between a visible label it has no room for and no accessible name at all. Seven first-party call sites currently choose the latter.

## Mechanism

In `src/ui/components/os-text-field/os-text-field.ts`:

- The visible label element is rendered only when `label` is set (L244–248).
- The accessible name is taken from the same value (L265): `aria-label=${ label || '' }`.
- `observedAttributes` (L42) does not include `aria-label`, and setting `aria-label` on the host is not forwarded to the inner `<input>`.

The component's own API description states the coupling: `{ name: 'label', type: 'string', description: 'Visible label above the input.' }` (L66).

So the two reachable states are:

- `label` omitted → the inner `<input>` gets `aria-label=""`, i.e. no accessible name.
- `label` set → the accessible name is correct, and a visible label is rendered above the field.

`placeholder` does not close the gap: a placeholder is not an accessible name, and it is removed from view on the first keystroke.

## Reproduce

1. Render `<os-text-field type="text" placeholder="Search notes"></os-text-field>`.
2. Inspect the inner `<input>` in the accessibility tree — it has no accessible name.
3. Add `label="Search notes"` to the same element. The accessible name appears, and the element's rendered height grows from 34px to 53px as the visible label is inserted.

Measured in Chrome against a shipped build; step 3 height delta observed live on the element.

## Scope in first-party apps

Scan of the tree with comments stripped and test files excluded: 29 shipping `<os-text-field>` call sites, 21 with a name, 7 without. All seven are placeholder-only search inputs:

- `apps/comments/parts/rail.ts:245`
- `apps/my-wordpress/my-wordpress.os.ts:435`
- `apps/plugins/plugins.os.ts:213`
- `apps/plugins/plugins.os.ts:276`
- `apps/posts/parts/app.ts:366`
- `apps/trash/trash.os.ts:360`
- `apps/users/users.os.ts:291`

The consistency is the signal. Every unnamed instance is a toolbar search; the named instances are form fields with room for a label. The omission tracks a layout constraint rather than an oversight, which is why adding `label` at each call site would regress seven toolbars rather than correct seven mistakes.

Both patterns already exist in-tree: `apps/os-settings/parts/components.ts` and `apps/code-blue/code-blue.os.ts` pass `label` alongside a `placeholder`.

## Related

#559 item 3 addressed the same class of problem for `wpd-window-button`, where a host-level label was never forwarded to the focusable inner `<button>`. This is the inverse case: forwarding works correctly, but there is no way to request the accessible name without also requesting the visual one.

## Context

Found while auditing Signal & Noise (https://github.com/juanlentino/signal-and-noise-tools), which registers an OpenStation app whose toolbar search hit the same choice.

juanlentino:
Fixes #789.

## The problem

`label` drove both the visible label and the accessible name from one value, so a compact field had two reachable states and neither was right:

- `label` omitted → the inner `<input>` gets `aria-label=""`, i.e. **no accessible name**
- `label` set → the name is correct, and a label renders above a toolbar search with no room for one

`placeholder` does not close the gap: it is not an accessible name, and it disappears on the first keystroke.

## The change

`hide-label` separates presentation from naming. The `<label>` is **still rendered, still paired with the input by `for=`, and still the accessible name** — it is only taken out of the visual flow, using the same idiom as `.os-constellation__row-note` in `openstation-layout.css`.

```html
<os-text-field label="Search notes" hide-label placeholder="Search notes…"></os-text-field>
```

Hiding it with `display: none`, or dropping the element and relying on `aria-label` alone, would have been easier and worse: the first removes it from the accessibility tree, and the second gives up the `<label for>` association, which is the more robust of the two bindings. Keeping both is the point.

Declared as `hideLabel` in `static props`, so `Component.observedAttributes` kebabs it to `hide-label` — matching `autoGrow`, `labelPosition` and the other multi-word props in the kit.

## Call sites

Four first-party search fields adopt it: **comments**, **plugins**, **trash**, **my-wordpress**. After this, no `<os-text-field>` under `apps/` is unnamed.

**A correction to the issue:** #789 says seven. That number was measured against a checkout that has since moved; on current `trunk` it is four. One of the seven I counted was also a false positive — `os-number-field`'s help `summary` string contains the literal text `<os-text-field>`, which my scanner read as a call site. It has its own `label` prop and was never affected. I have not edited the issue body; happy to if you'd prefer the record corrected there.

## Tests

Three cases in `os-text-field.test.ts`:

1. a labelled field pairs `<label for>` with the input **and** names it
2. `hide-label` keeps the label, the pairing and the name, and only adds the modifier class
3. **negative control** — `hide-label` *without* `label` names nothing, because it is a presentation switch and not a source of names

**One honest caveat on verification.** I could not run your vitest suite locally: `devEngines` pins Node to `>=24 <25` and this machine is on v26, so `npm` refuses before install. I validated what I could without the harness — the prop is declared, read as a boolean the same way `disabled` is, the modifier is applied, `aria-label` is untouched, the style rule exists and is *not* `display: none`, and `kebab( 'hideLabel' ) === 'hide-label'` against your own `kebab()` in `src/ui/core/component.ts`. The vitest cases themselves are unexercised, so please let CI be the judge of those.

## Not included

`os-textarea` and `os-number-field` have the same coupling. I left them alone rather than widen an accessibility fix into a kit-wide API change without your read on the attribute name first — if `hide-label` is the shape you want, extending it is mechanical.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-792-d570f39fe34051e9e602b2a5e415704c87537b7d.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/apps/comments/parts/rail.ts
+++ b/apps/comments/parts/rail.ts
@@ -243,6 +243,8 @@ export function rail( ctx: Ctx, ui: UiState, rows: CommentRow[], error: string )
 	return html`<aside class="${ NS }__rail" aria-label=${ __( 'Conversations' ) }>
 		<div class="${ NS }__search">
 			<os-text-field
+				label=${ __( 'Search comments' ) }
+				hide-label
 				placeholder=${ __( 'Search comments…' ) }
 				os-bind="search"
 				os-action="filter"
--- a/apps/my-wordpress/my-wordpress.os.ts
+++ b/apps/my-wordpress/my-wordpress.os.ts
@@ -450,6 +450,8 @@ export default defineApp< AppState, AppData >( 'my-wordpress', {
 				${ section && ! inFolder && ! isAgents && ! inFootprint
 					? html`<div class="os-mywp__search">
 						<os-text-field
+							label=${ __( 'Search this section' ) }
+							hide-label
 							value=${ state.query }
 							placeholder=${ sprintf(
 								/* translators: %s: section label, lowercased. */
--- a/apps/plugins/plugins.os.ts
+++ b/apps/plugins/plugins.os.ts
@@ -174,6 +174,8 @@ function browsePanel( ctx: Ctx, ui: UiState, phone: boolean ): TemplateResult {
 				} ) }
 				<os-text-field
 					class="os-app-list__search"
+					label=${ __( 'Search WordPress.org', 'desktop-mode' ) }
+					hide-label
 					os-bind="query"
 					os-debounce="300"
 					placeholder=${ __( 'Search WordPress.org…', 'desktop-mode' ) }
--- a/apps/trash/trash.os.ts
+++ b/apps/trash/trash.os.ts
@@ -400,6 +400,8 @@ export default defineApp< AppState, AppData >( APP_ID, {
 						</os-segmented>
 						<os-text-field
 							type="search"
+							label=${ __( 'Search trash' ) }
+							hide-label
 							os-bind="search"
 							os-action="refresh"
 							placeholder=${ __( 'Search trash…' ) }
--- a/src/ui/components/os-text-field/os-text-field.styles.ts
+++ b/src/ui/components/os-text-field/os-text-field.styles.ts
@@ -53,6 +53,25 @@ export const textFieldStyles = css`
 		color: var( --os-ui-fg-muted, #646970 );
 	}
 
+	/*
+	 * `hide-label`: the label is still rendered and still paired with the
+	 * input by `for=`, it is only taken out of the visual flow. Hiding it
+	 * with `display: none` would remove it from the accessibility tree too,
+	 * which is the opposite of the point. Same idiom as
+	 * `.os-constellation__row-note` in openstation-layout.css.
+	 */
+	.os-text-field__label--hidden {
+		position: absolute;
+		width: 1px;
+		height: 1px;
+		margin: -1px;
+		padding: 0;
+		border: 0;
+		overflow: hidden;
+		white-space: nowrap;
+		clip-path: inset( 50% );
+	}
+
 	.os-text-field__row {
 		position: relative;
 		display: flex;
--- a/src/ui/components/os-text-field/os-text-field.styles.ts
+++ b/src/ui/components/os-text-field/os-text-field.styles.ts
@@ -54,11 +54,11 @@ export const textFieldStyles = css`
 	}
 
 	/*
-	 * `hide-label`: the label is still rendered and still paired with the
-	 * input by `for=`, it is only taken out of the visual flow. Hiding it
-	 * with `display: none` would remove it from the accessibility tree too,
+	 * hide-label: the label is still rendered and still paired with the
+	 * input by for=, it is only taken out of the visual flow. Hiding it
+	 * with display: none would remove it from the accessibility tree too,
 	 * which is the opposite of the point. Same idiom as
-	 * `.os-constellation__row-note` in openstation-layout.css.
+	 * .os-constellation__row-note in openstation-layout.css.
 	 */
 	.os-text-field__label--hidden {
 		position: absolute;
--- a/src/ui/components/os-text-field/os-text-field.test.ts
+++ b/src/ui/components/os-text-field/os-text-field.test.ts
@@ -169,7 +169,9 @@ describe( '<os-text-field>', () => {
 		const el = host.querySelector( 'os-text-field' )!;
 		expect( el.shadowRoot!.querySelector( 'label' ) ).toBeNull();
 		const input = el.shadowRoot!.querySelector( 'input' ) as HTMLInputElement;
-		expect( input.getAttribute( 'aria-label' ) ).toBe( '' );
+		// The renderer drops an empty aria-label entirely, so the absence
+		// reads as null — no attribute — rather than an empty string.
+		expect( input.getAttribute( 'aria-label' ) ).toBeNull();
 	} );
 
 	test( 'invalid attribute surfaces aria-invalid on the native input', async () => {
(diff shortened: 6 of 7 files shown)


epeicher:
Thanks for contributing to the repo @juanlentino!

All five checks are green on `ec8e244d`. I had independently reached the same two fixes before you pushed, including `toBeNull()` on the negative control, so the reasoning in your commit message matches what I found reading the renderer: `applyValues()` in `src/ui/core/html.ts` calls `removeAttribute` whenever an attribute composes to `''`, so an unlabelled field carries no `aria-label` at all. Nice catch verifying it red by making the renderer emit a stray one.

The approach itself is the right one, and I think this is ready bar one small thing.

## Verified behaviour

I ran the branch on a local instance and measured the Comments rail, which is one of the four adopted call sites:

| | value |
|---|---|
| label text | `Search comments` |
| label rendered size | 1x1, `clip-path: inset(50%)` |
| `for=` paired with input id | yes |
| accessible name | `Search comments` |
| host height vs input height | 34px vs 34px |

That last row is the part the unit tests cannot show, and it is the one I most wanted to check. Because the label is `position: absolute` it leaves the flex flow, so neither label space nor the `:host` `gap: 4px` is reserved, and the toolbar geometry is unchanged. The same field with a visible label measures 56px. I also confirmed the label keeps `display: block` and `visibility: visible` under the modifier, so it stays in the accessibility tree, which was the whole point of not reaching for `display: none`.

Full local gate on your head: build clean, lint 0 errors, typecheck clean, Vitest 6108/6108, PHPUnit 3036 tests on both single-site and multisite, Plugin Check 0 errors on both plugins.

## One thing to change

`static help` declares the prop as `hideLabel`:

```ts
{
    name: 'hideLabel',
    type: 'boolean attribute',
```

It is the only camelCase entry in any help block across the whole kit. Every other multi-word prop is documented by its attribute spelling, and there is already a `hide-cancel` precedent alongside `auto-grow` and `label-position`. Since Preferences renders these descriptors live and that page is what tells an author what to type in markup, `hideLabel` there will send people to write an attribute that does not exist.

`static props` should of course keep `hideLabel`, because `observedAttributes` kebabs it. It is only the help name.

## Two notes, no action needed here

**On the "no `<os-text-field>` under `apps/` is unnamed" claim.** There is one left, `apps/users/profile/options.ts:96`. It is correct as it stands though: that field is `display: none` and exists only to carry a value into the form's collection, so it is out of the accessibility tree and has no name to give. Worth knowing the exception exists rather than changing it.

**`syncLibraryControls()` is the hack this prop should eventually retire.** In `apps/plugins/parts/installed-library.ts` it mirrors `aria-label` and `aria-pressed` from the host onto the kit component's inner `button`/`input` inside a `queueMicrotask`, scoped to `.os-plugins__workspace`. I confirmed a bare host `aria-label` does not reach the inner input on its own, so that function is the only reason the Installed tab's search has a name today. The upshot is that after this PR the Plugins app carries two naming idioms side by side: the Add Plugin tab uses `hide-label`, the Installed tab still uses the mirror. That is a good follow-up rather than something to widen this PR with, and your instinct to leave `os-textarea` and `os-number-field` alone for the same reason was the right call.

On that last point, if `hide-label` is the shape we settle on, extending it to those two and then unwinding the mirror would be a tidy second PR.

AllTerrainDeveloper (approved):
Thanks!

PR #793, opened 2026-09-10
Post Stats: draw the chart chrome in tokens instead of hardcoded black

### LINKED ISSUE #762: Post Stats: the chart's axis labels and grid are drawn in hardcoded black, invisible on the widget card's dark glass
### What happens

In the Post Stats widget, the bars render but everything drawn as text or rule
on the canvas is invisible: the y-axis tick numbers, the month labels under the
bars, and the horizontal grid lines. The widget's own DOM text — the title, the
`39 posts in 6 mo` metadata, the legend — is correct and legible.

Observed on v1.1.6, default palette (no custom desktop theme installed).

### Why

`src/plugins/post-stats-widget/index.ts` paints the canvas chrome with
hardcoded black:

| line | draws | value |
|---|---|---|
| 131 | horizontal grid lines | `ctx.strokeStyle = 'rgba(0,0,0,0.07)'` |
| 139 | y-axis tick numbers | `ctx.fillStyle = 'rgba(0,0,0,0.35)'` |
| 166 | month labels under the bars | `ctx.fillStyle = 'rgba(0,0,0,0.5)'` |

Line 160 (`ctx.fillStyle = seg.color`) uses the explicit `COLORS` map, which is
why the bars are the only part that survives.

`.os-widgets__card` is the fixed dark glass — `--os-ui-color-surface` is
`rgba(20, 20, 22, 0.55)` — so black ink on it measures, composited over the
wallpaper:

| element | contrast on the card glass |
|---|---|
| grid lines | **1.01 : 1** |
| y-axis numbers | **1.02 : 1** |
| month labels | **1.03 : 1** |
| `--os-ui-color-text-subtle`, for reference | 9.21 : 1 |

(That last figure reproduces the `9.2:1` stated in `variables.css`, as a check
that this is measured the same way you measure it.)

### Why the existing guard cannot see it

`tests/vitest/widget-card-token-contract.test.ts` already fixed exactly this
legibility class for this exact widget — its docblock records *"Post Stats'
metadata measured 3.81:1"* — and requires `> 4.5` on the glass for the text
names, under both the palette and Legacy.

That guard resolves **CSS custom properties**. A canvas `fillStyle` is neither a
custom property nor CSS, so the same bug in a different medium was never in
scope. The CSS half of Post Stats is fixed; the canvas half was never reachable
by the test.

### Scope — I checked whether this is a class, and it is not

Of the nine bundled widgets, Post Stats is the only one affected:

| widget | draws with | hardcoded dark |
|---|---|---|
| post-stats | canvas | **3** (lines 131, 139, 166) |
| site-views | **SVG** — styleable by CSS, inherits the card's tokens | 0 |
| heartbeat | canvas, but paints a gradient | 0 |
| drafts, focus-timer, jazz-quote, notes, recent-comments, starter | no colour assignments | 0 |

Outside the widgets I also looked at the App Framework surfaces that reference a
black literal (`apps/users/profile/insights.ts`); those are CSS `var()` with a
fallback, and the names they read — `--os-ui-holo-track`,
`--os-ui-badge-neutral-bg` — are both declared in `variables.css`, so they
resolve. The wallpaper scenes and `src/games/share-card.ts` also paint canvas
literals, but onto the desktop background and an offscreen share image
respectively, not onto the card, so they are out of scope here.

So this looks like a single-file fix rather than a sweep.

### Suggested direction

The values presumably predate the dark card, from when the widget sat on a light
surface. Two options, and the second looks closer to the file's existing intent:

1. Read the declared names off the card at draw time —
   `getComputedStyle( canvas ).getPropertyValue( '--os-ui-color-text-subtle' )`
   for the labels and the hairline name for the grid — so the chart follows the
   same contract the CSS now follows, including for a theme that moves the card.
2. Literals chosen for the dark glass, matching how `variables.css` argues these
   names should be literals rather than aliases of the text ramp.

Either way the guard gap is worth closing separately, since a canvas-drawn
widget is invisible to the token contract test by construction. A cheap version
is asserting that this renderer contains no `rgba(0,0,0` / `#000` literal.

juanlentino:
Fixes #762.

## The change

Three canvas paints in `src/plugins/post-stats-widget/index.ts` were literal black:

| line | draws | was | now |
|---|---|---|---|
| 131 | horizontal grid lines | `rgba(0,0,0,0.07)` | `--os-ui-color-border` |
| 139 | y-axis tick numbers | `rgba(0,0,0,0.35)` | `--os-ui-color-text-subtle` |
| 166 | month labels | `rgba(0,0,0,0.5)` | `--os-ui-color-text-subtle` |

These are the same two custom properties the widget's **own stylesheet already uses** for its DOM text — which is precisely why the title, the `39 posts in 6 mo` metadata and the legend were legible while everything on the canvas was not. The bars survived because they use the explicit `COLORS` map.

## Two decisions worth flagging

**The ad-hoc alphas go with the black.** `0.07 / 0.35 / 0.5` were compensating for black ink on a light assumption; the tokens already encode the rule-vs-label hierarchy, and a canvas cannot cheaply apply an alpha to an arbitrary colour string without parsing it. If you'd rather keep a weight difference between the ticks and the month labels, `globalAlpha` around the tick pass is the clean way and I'm happy to take direction on it.

**The fallbacks are deliberately not black.** `getPropertyValue()` returns `''` when a token has not resolved, so the fallback is what ships in exactly the case the token exists to cover. Falling back to black would restore the bug there. Both fallbacks are the dark-glass values from `variables.css`.

## Tests

Two cases added to `tests/vitest/widget-card-token-contract.test.ts`, which already exists for this class of bug. One asserts the source carries no literal black and reads both tokens; the other asserts the *fallbacks* are legible on the glass — without that, a future edit could satisfy the first pin and still ship black as the fallback.

A canvas cannot be asserted in jsdom without a 2D context, so these are source-level pins rather than rendered-pixel ones. That is a real limitation of the check and worth knowing when reading it.

**Verification caveat, same as my other PRs:** I could not run your vitest suite — `devEngines` pins Node to `>=24 <25` and this machine is on v26, so npm refuses before install. I ran each pin's logic in plain node against the real file in both directions: it passes on the patched source and **trips on the pre-patch source**, so the guard is failable rather than vacuous. The vitest harness itself is unexercised.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-793-07596f1dc1aeb83c8ba88f3d79d7b1c3a59aee4f.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/src/plugins/post-stats-widget/index.ts
+++ b/src/plugins/post-stats-widget/index.ts
@@ -113,6 +113,21 @@ function drawChart( canvas: HTMLCanvasElement, buckets: Bucket[] ): void {
 	}
 	ctx.scale( dpr, dpr );
 
+	// Canvas ink, read from the same custom properties this widget's own
+	// stylesheet already uses for its DOM text. The chrome was hardcoded
+	// black, which is why the bars survived and everything drawn as text or
+	// rule did not: `.os-widgets__card` is a fixed dark glass in every
+	// desktop theme, so black measured 1.01:1 on it.
+	//
+	// The fallbacks are the dark-glass values from variables.css, NOT the
+	// old black — a widget that renders before the tokens resolve should
+	// fail toward legible rather than back into the bug.
+	const ink = getComputedStyle( canvas );
+	const inkOf = ( token: string, fallback: string ): string =>
+		ink.getPropertyValue( token ).trim() || fallback;
+	const RULE = inkOf( '--os-ui-color-border', 'rgba( 255, 251, 255, 0.12 )' );
+	const LABEL = inkOf( '--os-ui-color-text-subtle', 'rgba( 255, 251, 255, 0.7 )' );
+
 	const W = rect.width;
 	const H = rect.height;
 	const PAD = { top: 18, right: 10, bottom: 22, left: 28 };
@@ -128,7 +143,7 @@ function drawChart( canvas: HTMLCanvasElement, buckets: Bucket[] ): void {
 	const barPad = barGroupW * 0.2;
 	const barW = Math.max( 1, barGroupW - barPad * 2 );
 
-	ctx.strokeStyle = 'rgba(0,0,0,0.07)';
+	ctx.strokeStyle = RULE;
 	ctx.lineWidth = 1;
 	for ( let i = 0; i <= 3; i++ ) {
 		const y = PAD.top + chartH - ( chartH * i / 3 );
@@ -136,7 +151,7 @@ function drawChart( canvas: HTMLCanvasElement, buckets: Bucket[] ): void {
 		ctx.moveTo( PAD.left, y );
 		ctx.lineTo( PAD.left + chartW, y );
 		ctx.stroke();
-		ctx.fillStyle = 'rgba(0,0,0,0.35)';
+		ctx.fillStyle = LABEL;
 		ctx.font = '9px -apple-system, sans-serif';
 		ctx.textAlign = 'right';
 		ctx.fillText( String( Math.round( maxVal * i / 3 ) ), PAD.left - 4, y + 3 );
@@ -163,7 +178,7 @@ function drawChart( canvas: HTMLCanvasElement, buckets: Bucket[] ): void {
 			ctx.fill();
 		}
 
-		ctx.fillStyle = 'rgba(0,0,0,0.5)';
+		ctx.fillStyle = LABEL;
 		ctx.font = '9px -apple-system, sans-serif';
 		ctx.textAlign = 'center';
 		ctx.fillText( b.label, x + barW / 2, PAD.top + chartH + 13 );
--- a/src/plugins/post-stats-widget/index.ts
+++ b/src/plugins/post-stats-widget/index.ts
@@ -113,6 +113,16 @@ function drawChart( canvas: HTMLCanvasElement, buckets: Bucket[] ): void {
 	}
 	ctx.scale( dpr, dpr );
 
+	const W = rect.width;
+	const H = rect.height;
+	const PAD = { top: 18, right: 10, bottom: 22, left: 28 };
+	const chartW = W - PAD.left - PAD.right;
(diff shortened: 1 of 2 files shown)


epeicher:
Thanks for your contribution @juanlentino! Two small things:

- CI's ESLint step fails on `src/plugins/post-stats-widget/index.ts` (lines 128–129): `@wordpress/no-unused-vars-before-return`, the variables are assigned before an early return. Moving those assignments below the return should clear it.
- Trunk now admits Node 26 (#794), so a merge with trunk will let you run `npm ci` and the vitest suite locally without the `devEngines` refusal.

epeicher (approved):
Thanks @juanlentino! It looks much better now. Changes also LGTM!

|Before | After|
|-------|------|
|Image|Image

PR #809, opened 2026-09-12
OS Settings: let a registered settings tab name its sidebar glyph

### LINKED ISSUE #808: OS Settings: a plugin-registered settings tab cannot have a sidebar icon
## Summary

A settings tab registered by a plugin renders in the OS Settings sidebar without a glyph. The label is aligned with the core tabs, but the icon column is an empty spacer. There is no way for the plugin to supply one: the settings-tab registry has no `icon` field, on either the PHP or the JS side.

## Where

`apps/os-settings/parts/nav-icons.ts` documents the limitation:

> A tab with no entry renders without a glyph and keeps its label aligned with the rest, which is the case every third-party tab is in: the settings-tab registry has no icon field, so a plugin cannot supply one even if it wanted to.

`apps/os-settings/parts/pages.ts` resolves the glyph for registry-delivered tabs by raw id only (`icon: NAV_ICONS[ tab.id ]`), which is how the shell's own File Associations tab gets one while any other registered id gets nothing.

Neither `DesktopSettingsTab` (`src/settings/registry.ts`) nor `openstation_register_settings_tab()` (`includes/settings-tabs.php`) accepts an icon.

## Reproduce

1. Register a tab from a plugin:
   ```js
   wp.os.registerSettingsTab( { id: 'my-tab', label: 'My Tab', order: 32, render } );
   ```
2. Open OpenStation Preferences.
3. The tab appears in the sidebar, under Features, with an empty icon slot. Adding an `icon` key to the registration changes nothing.

Observed on 1.1.8 with [Signal & Noise Tools](https://github.com/juanlentino/signal-and-noise-tools) 14.1.0, which registers a `signal-noise` tab via both `openstation_register_settings_tab()` and `wp.os.registerSettingsTab()`.

<details><summary>Screenshot</summary>

Image

</details>

## Proposed shape

An optional `icon` on the registration, in the vocabulary `osIcon()` already uses for the core tabs (an icon-set name, not a URL — the column renders 17px monochrome glyphs), with `pages.ts` preferring `tab.icon` and falling back to `NAV_ICONS[ tab.id ]` as today. Unknown keys are already ignored, so plugins can set it ahead of the release that reads it.

A PR follows.

juanlentino:
Fixes #808

## What changes

A plugin-registered Preferences tab rendered with an empty icon column: the settings-tab registry had no `icon` field, so the sidebar could only draw glyphs for ids the shell knew by hand (File Associations).

- `DesktopSettingsTab` gains an optional `icon` — a name from the OS icon set (`settings`, `bell`, `apps`, …), the same vocabulary the core tabs already use. Names rather than URLs: the column draws 17px monoline glyphs from `currentColor`, so a bitmap or hard-coloured SVG would not match the theme (or, once Brand Studio lands, the brand).
- `registryNavIcon()` in `nav-icons.ts` resolves a row's glyph: the registered name if it is in the set, else the shell's own table by raw id, else none. `pages.ts` calls it in place of the bare `NAV_ICONS[ tab.id ]` lookup.
- An unknown or absent name still draws the spacer, so a plugin can set the field ahead of the release that reads it. `osIconDef()` already refuses inherited keys, so a plugin-supplied string cannot reach the prototype chain.
- Docs: one row in the `registerSettingsTab()` table in `docs/javascript-reference.md`.

The PHP `openstation_register_settings_tab()` is deliberately unchanged: the sidebar is painted from the JS registry, and the server entry is only consulted to unregister tabs on deactivation, so a PHP `icon` would be data nothing reads.

## Test

`apps/os-settings/os-settings.test.ts` — one new case: a tab registered with `icon: 'bell'` renders an `<svg>` and no spacer; re-registered with `icon: 'not-an-icon'` it renders the spacer and no `<svg>`. Run against trunk before the change it fails on the first assertion; with the change the file is 13/13.

`npm run typecheck` and `eslint` on the touched files pass.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-809-35d499fb7df01d1bc677c78f4f73817415e33d18.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/apps/os-settings/os-settings.os.ts
+++ b/apps/os-settings/os-settings.os.ts
@@ -57,9 +57,8 @@ const DEFAULT_TAB = 'appearance';
 /** A row's sidebar glyph — one node per row, made once. */
 function glyph( ui: UiState, row: PageRow ): SVGSVGElement | unknown {
 	if ( ! row.icon ) {
-		// Third-party tabs have no glyph to render: the registry has
-		// no icon field. The spacer keeps their label on the same
-		// line as every other label.
+		// A registry tab that named no icon (or an unknown one). The
+		// spacer keeps its label on the same line as every other label.
 		return html`<span class="os-settings__nav-glyph-blank" aria-hidden="true"></span>`;
 	}
 	let node = ui.glyphs.get( row.id );
--- a/apps/os-settings/os-settings.test.ts
+++ b/apps/os-settings/os-settings.test.ts
@@ -122,6 +122,20 @@ describe( 'OpenStation Preferences — the frame', () => {
 		expect( tabCtx.getOsSettings() ).toMatchObject( { accent: 'pulse' } );
 	} );
 
+	test( 'a registry tab draws the icon-set glyph it names, and a spacer otherwise', () => {
+		registerSettingsTab( { id: 'acme', label: 'Acme', icon: 'bell', render: () => undefined } );
+		paint();
+		const named = root.querySelector( '#os-settings-nav > os-tab[value="ext-acme"]' );
+		expect( named?.querySelector( 'svg' ) ).not.toBeNull();
+		expect( named?.querySelector( '.os-settings__nav-glyph-blank' ) ).toBeNull();
+
+		registerSettingsTab( { id: 'acme', label: 'Acme', icon: 'not-an-icon', render: () => undefined } );
+		paint();
+		const unknown = root.querySelector( '#os-settings-nav > os-tab[value="ext-acme"]' );
+		expect( unknown?.querySelector( 'svg' ) ).toBeNull();
+		expect( unknown?.querySelector( '.os-settings__nav-glyph-blank' ) ).not.toBeNull();
+	} );
+
 	test( 'an admin-only registry tab is hidden from an editor', () => {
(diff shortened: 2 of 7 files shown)


AllTerrainDeveloper (approved):
This is great Juan! Thank you!

PR #816, opened 2026-09-14
Add window-scoped MIO assistance and private Settings actions

AllTerrainDeveloper:
MIO can now help inside an explicitly registered window, using that app’s live prompt, linked Markdown help and private actions. OpenStation Preferences is the worked example: users can ask for changes such as rounded corners plus unified, dynamic docks, with sequential execution and verified before/after results.

https://github.com/user-attachments/assets/36e7fccc-3174-460b-9fa1-5321fa7d8e30
https://github.com/user-attachments/assets/e901ff8f-8d64-4977-aae5-0303022b1a2e

- Add app-opted-in window residency, focus handoffs, per-window title-bar toggles and caller-directed speech bubbles. Idle window mascots stay hidden until chat or a callout needs them.
- Add a themed, memory-only floating chat, bounded help retrieval and a cancellable action loop. Tools remain outside WordPress’s public Abilities registry; existing authenticated app/REST endpoints retain authority for writes. Destructive Settings operations are excluded.
- Supply nine linked Preferences help documents, live theme/wallpaper/control catalogs, private Settings actions and an About journal callout as integration examples.
- Blend mascot transitions, chat anchoring, drag return and thinking animation; preserve the start of long replies and repaint synchronously during resizing.
- Synchronize the dock button with Features → MIO API. Add independent wallpaper visibility in Features, Make it yours and the dock context menu. Ask MIO requires enabled AI and a compatible connector; offline callouts do not.
- Add app-owned response buttons for read/navigation actions, such as opening a saved form preview without another model call or save. Callbacks remain private to the window lease; message history carries only opaque references. Include a receipt-bound Forms integration recipe and an isolated browser fixture.
- Make the chat composer wrap and grow to two rows, with Enter to send, Shift+Enter for a newline and IME-safe submission.
- Restore keyboard focus to the native title-bar control, dispose detached window hosts, and preserve Unicode characters across help continuation boundaries.
- Document the experimental API, hooks, theme tokens, lifecycle and integration examples, including the distinction between app registration, the global master switch and the per-window toggle.

The consumer API also supports bounded argument repair, effect-aware read/write accounting, compact document history with UTF-8 request-budget checks, section-aware help with continuation, and stable turn/call identities with lifecycle observers and read-only operation-status reconciliation. Tests cover a complete form larger than 40 KB with two repairs and help reads, repeated reads after writes, duplicate receipts, retry exhaustion without writes, and cancellation during a save. A form-editing integration recipe documents app-owned revisions, idempotency and status endpoints.

Conversation history is not persisted by OpenStation. Model requests send the bounded context to the configured provider. Action chains are not transactions: cancellation prevents subsequent actions but does not undo completed writes.

Validation:

- Full JavaScript suite: 6,235 tests passed across 515 files, including response-action lifecycle, duplicate activation, focus restoration, detached hosts, two-line composition, Unicode help boundaries, exact UTF-8 budgets and terminal non-retryable failures.
- Full PHP suite: 3,099 tests, 23,284 assertions, zero failures and 13 environment-dependent skips. The Legacy guard covers compound window-root selectors while distinguishing shared palette defaults from descendant/component overrides; the frozen manifest is unchanged.
- Full build, TypeScript check and PHP lint passed. JS lint passed with existing file-length warnings.
- Built and synced to the local WordPress site. Browser-verified dock/checkbox synchronization, wallpaper visibility and reload persistence, dock-menu recovery, chat with desktop MIO hidden, the About callout, and a real read-only wallpaper query through the configured provider. Also browser-verified the isolated response-action fixture: saved-ID preview targeting, duplicate-click suppression, local failures, cancellation/reopening, permission revocation, keyboard focus and narrow/RTL layouts, plus composer wrapping and its two-row limit. The fixture uses simulated saves and provider responses; the external Forms plugin is represented by the documented integration recipe. Manual regression checklist included under `tests/e2e/mio/`.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-816-392d2ae65694128b835d33c219a74efaf54e684c.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


AllTerrainDeveloper:
@claude can you review this?

src/mio/residency.ts
```diff
@@ -0,0 +1,378 @@
...
+
+	public closeChat( focus = false ): void {
+		this.chatPlacement?.close( focus );
+		this.chatPlacement = null;
+		this.chat?.destroy();
+		this.chat = null;
+		if ( this.owner ) {
+			this.owner.button.hidden = ! this.canChat();
+			if ( focus ) {
+				this.owner.button.focus();
```
epeicher: `closeChat( true )` is meant to return keyboard focus to the **Ask MIO** launcher after Escape or the chat's × button, but `button` is an `<os-button>` host: it has no `tabindex`, and `src/ui/core/component.ts:88` attaches the shadow root without `delegatesFocus`. `HTMLElement.focus()` on it is a no-op. The panel is removed a line earlier, so focus falls back to `<body>` and a keyboard user loses their place in the window.

Elsewhere the repo reaches the focusable shadow `<button>` explicitly — `src/workspaces/wizard.ts:1113` and `src/ui/components/os-action-menu/os-action-menu.ts:24` — and `mountMioChat` itself already uses that pattern for the text field (`input.shadowRoot?.querySelector( 'input' )?.focus()`), which is what makes this look like an oversight rather than a deliberate choice.

```suggestion
				( this.owner.button.shadowRoot?.querySelector( 'button' ) ?? this.owner.button ).focus();
```

tests/phpunit/tests/desktopThemesLegacy.php
```diff
@@ -212,6 +212,61 @@ public function test_the_snapshot_is_frozen() {
...
+	 * @param string $css Palette stylesheet.
+	 * @return array<string, bool> Literal token names.
+	 */
+	private function palette_literals( $css ) {
+		$css = preg_replace( '~/\*.*?\*/~s', '', $css );
+		preg_match_all( '/([^{}]+)\{([^{}]*)\}/', $css, $rules, PREG_SET_ORDER );
+		$literals = array();
+		foreach ( $rules as $rule ) {
+			$selectors = array_map( 'trim', explode( ',', $rule[1] ) );
+			if ( ! in_array( 'body.os-active', $selectors, true ) ) {
```
epeicher: This narrowing goes one step further than the new `.os-mio-*` overrides require, and it silently drops a rule that is not a component override.

`variables.css:1793` has a top-level rule outside `body.os-active`:

```css
.os-window:not( .os-window--native ) {
	--os-window-reveal-surface: #fff;
}
```

The previous whole-file scan required Legacy to answer that literal, and it does (`assets/desktop-themes/legacy/theme.json:171`). Requiring the selector list to contain exactly `body.os-active` removes it from the guard, so nothing fails today but the test would no longer notice if that entry were deleted, or if a new global literal landed in another top-level rule.

The earlier review read this hunk as narrowing "to top-level `body.os-active` rules"; the check is stricter than that, and `.os-window:not( .os-window--native )` is the rule that falls outside it.

If the intent is only to exclude component-scoped aliases, match on the absence of a descendant combinator rather than on an exact `body.os-active` selector.

epeicher (commented):
Second pass over this branch, focused on what the earlier automated review did not cover. I agree with its conclusions on the server route, the tool loop, the budget arithmetic and the settings alias — verified independently. Two findings to add, one of which contradicts a point in that review.

Verification run locally at `8eadeb9`: `npm run typecheck` clean, `npm run test:js` 6,196 passed / 7 skipped, `npm run lint` 0 errors (only pre-existing file-length warnings). PHPCS and PHPUnit were not run here — they need a live wp-env.

Both findings are inline below.

PR #825, opened 2026-09-15
Lazy scripts: replay a bundle's declared dependencies on live activation

epeicher:
## What it does

A plugin activated from inside the shell now opens its windows configured, without an F5. Reported with AllTerrain Forms: the tile appeared live, but the builder opened with "AllTerrain Forms is not configured on this page" until the page was reloaded.

https://github.com/user-attachments/assets/24746672-1017-42c5-83b0-99aa34f8fb69

## Rationale

AllTerrain Forms ships its config on a src-less alias handle (`wp_register_script( 'allterrain-forms-config', false )` plus `wp_add_inline_script()` setting `window.allTerrainForms`), declared as a dependency of every bundle so it always runs first. On a normal boot WordPress prints that dependency. On a live activation the shell fetches the bundle lazily, and two things were missing:

- No lazy path except widgets resolved a handle's dependency closure at all.
- `openstation_resolve_script_payload()` returned an empty payload for any handle without a `src`, so the alias could not have been delivered even by a path that walked dependencies.

A second layer showed up while verifying: the plugin registers the same bundle as a command script too, and `loadVendorScript` memoizes by URL, so **whichever path fetches a bundle first decides what ran before it**. Fixing native windows alone left the command sync winning that race with no deps. That is why every lazy payload builder ships the closure now, not just native windows.

## Implementation

- `openstation_resolve_script_payload()` keeps an alias's `before` / `after` / `l10n` with an empty `url`, mirroring what `WP_Scripts::do_item()` prints for it (translations excluded, as Core only prints those for a handle it printed a tag for). The command-palette manifest's local copy of that harvest is gone.
- Native windows: each loadable handle's entry in `nativeWindowScriptData` carries `deps`, an ordered handle list whose members land in the same map, so a shared package is serialized once. `hydrateServerEntries()` joins them into `scriptDeps` on the entry, its companions and its tabs; `loadOnce()` hands them to the loader.
- Commands, settings tabs, dock-rail renderers, title-bar buttons, window actions, unfocus effects, window links, window chrome (themes/controls/slots/chromes), wallpapers, games and desktop-file openers ship `scriptDeps` via `openstation_resolve_script_dependencies()`, and each `server-sync` passes `deps` through. Widgets already did; the type now lives once as `LazyScriptDependency`.
- `loadVendorScript`: a dependency with an empty `url` is replayed inline in print order, once per document (`replayedAliases`), and skipped when Core already printed it. `isScriptInDocument()` gains that third signal: a tag under Core's printed ids (`<handle>-js`, `<handle>-js-before` and siblings), the only trace an alias leaves.

Docs: `docs/migration-wp-package-globals.md` (the "no other lazy path does this" paragraph is now the opposite), the `script` note in `docs/hooks-reference.md`, `docs/architecture.md`, `docs/javascript-reference.md`.

**Not fixed here:** AllTerrain Photo Editor still opens blank after a live activation. Its `lienzo` bundle is only enqueued on `openstation_mode_init` at boot and never appears in any payload, so the shell cannot know about it. That is plugin-side: declare it through the `openstation_app_window_args` filter as a companion `scripts` handle, and attach `window.lienzoConfig` at registration rather than at enqueue time. With this PR merged, both then ride the live payload. Follow-up on the plugin: AllTerrainDeveloper/allterrain-photo-editor#10.

## Cost of resolving the closure per request

Every sub-builder now calls `openstation_resolve_script_dependencies()` for each handle it names, with no memo across builders (a handle named by two builders is resolved twice). Measured on a member site as admin, after `admin_enqueue_scripts`, with 271 registered handles and 827 dependency edges:

| Measurement | Result |
|---|---|
| One `openstation_build_menu_payload()` call, cold / warm | 0.70 ms / 0.38 ms |
| Resolving one handle with a 5-package closure | 0.04 ms |
| Resolving every handle the payload names (16 handles, 31 dependency payloads) | 0.2 ms per request |

The repeated work is about a fifth of a millisecond on a request that costs hundreds, so it stays as is. A per-request memo is not free either: inline data can be attached to a handle between resolutions (the refresh probe replays `admin_enqueue_scripts`, and modules attach config at priority 5 while the payload builds at 10), so a memo that outlives that window would ship stale data, which `test_probe_payload_carries_data_attached_on_admin_enqueue_scripts` exists to catch.

## Testing instructions

On an install with AllTerrain Forms, deactivate it, reload the shell, activate it from the Plugins window, then double-click the Forms desktop icon: the builder opens configured. Before this change the same steps reproduce the report.

```bash
npm run test:js -- vendor-loader script-presence native-windows-lazy-script commands-server-sync
npm run test:php -- --filter='Tests_OpenStation_ResolveScriptDependencies|Tests_OpenStation_NativeWindowLazyScript|Tests_OpenStation_Commands'
```

New tests pin: an alias dependency keeps its inline data and an empty one is dropped; a window ships its closure and a dependency later named as a script resolves its own; the command payload ships `scriptDeps`; alias replay order, once-per-document, and skip-when-printed; id-based presence. Full `test:js`, `test:php`, `lint`, `lint:php` and `typecheck` are green.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-825-75bcf4895cd8870e4d8c7b087e530f9c8b218dec.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


epeicher:
@claude please review

epeicher:
On the "repeated work per request" observation: measured rather than estimated, and it is negligible. On a member site as admin, after `admin_enqueue_scripts`, with 271 registered handles and 827 dependency edges:

| Measurement | Result |
|---|---|
| One `openstation_build_menu_payload()` call, cold / warm | 0.70 ms / 0.38 ms |
| Resolving one handle with a 5-package closure | 0.04 ms |
| Resolving every handle the payload names (16 handles, 31 dependency payloads) | 0.2 ms per request |

About a fifth of a millisecond per request, so no memo. A per-request cache would also need scoping to a single build: inline data can be attached to a handle between resolutions (the refresh probe replays `admin_enqueue_scripts`; modules attach config at priority 5, the payload builds at 10), and a memo that outlives that window ships stale data. Numbers are recorded in the PR description as well.

PR #832, opened 2026-09-16
Add WooCommerce order details to window titles

### LINKED ISSUE #800: Use "Order #" for title of order window.
When opening an order, it would be nice to use the title of the order as the window's title, for example "Order #5632" rather than just "Woocommerce". 

Ideally it would be awesome to have the customer's name appended to the window title, for example "Order #5632 - John Smith"

```
Plugin version:    1.1.7
WordPress version: unknown
User agent:        Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/152.0.0.0 Safari/537.36
Viewport:          2369x1393
Platform:          MacIntel
Current URL:       https://oceansev.com/wp-admin/admin.php?page=openstation
```

</details>

AshishJii:
**Description**

Updates WooCommerce order windows to display titles such as `Order #5632 · John Smith`, with a fallback to `Order #5632` when no customer name is available.

Adds PHPUnit coverage for title formatting and non-WooCommerce behavior.

**Tests**

- Focused PHPUnit suite: 46 tests passed
- PHP lint passed
- Typecheck passed
- Full build passed

Closes #800


mmtr:
Thanks @AshishJii, and welcome! Very nice first contribution 🙂 

However, the custom title ends up being reverted after the iframe loads, which is when `adoptPageTitle()` runs on the iframe load and replaces the title with the page's own screen name:


https://github.com/user-attachments/assets/e7c01d21-f285-44ac-8c98-f5a6b796b07b

The fix can be something like a flag set when the title is externally handled and checked in `adoptPageTitle()` (in `src/window/`).

Minor: `unset( $hook_suffix )` isn't needed, the callback can just take no parameters.

PR #850, opened 2026-09-18
Add the OpenStation Feedback Intake plugin under tools/

epeicher:
## What it does

Adds **OpenStation Feedback Intake** under `tools/feedback-intake/`: the small WordPress plugin that runs on openstation.blog and receives what an OpenStation site sends when an administrator deactivates the plugin and clicks **Send and deactivate** (#840). It stores each submission in one table and shows them under **Tools → OpenStation feedback**.

It is not part of the release: `bin/package.sh` only zips the paths in its include list, and `tools/` is not one of them. It is also not an extension: everything under `extensions/` runs on OpenStation sites, and this runs on ours. Whether it stays in this repo or moves to its own is still open; this PR lets it be reviewed and tested either way.

## The routes

```
POST /wp-json/openstation-feedback/v1/deactivation     public, throttled
GET  /wp-json/openstation-feedback/v1/deactivation     manage_options (cookie or application password)
```

The POST body is the payload `includes/feedback/deactivation.php` builds in #840. Every key is typed and bounded, the enums are closed, and an unknown key is rejected. A submission is one `INSERT IGNORE` on its uuid, so a retried send is a 200 and not a second row. The GET returns rows newest first with `X-WP-Total`, for the metrics script.

## How the public route is protected against abuse

The POST has to be reachable by anyone: the senders are anonymous WordPress sites. What keeps it cheap under a flood, in the order the checks run:

1. **A gate on `rest_pre_dispatch`**, which Core runs before it matches the route, parses the body or validates anything. A permission callback would be too late: Core validates the schema first and only then asks for permission, so the size and throttle checks would have run after a JSON parse of whatever the request carried.
2. **Wrong content type answers 415. Empty body, or a body over 2 KB, answers 413.** A real submission is about 500 bytes. Neither check reads the body as JSON.
3. **Two throttles, both counters in transients** (the object cache when the host has one, so they hold across web nodes). Per client IP, 5 submissions a minute. Global, 300 a minute across every sender, so a distributed flood is bounded before the database becomes the bottleneck. Over either answers 429 with `Retry-After`. The IP is `REMOTE_ADDR` only, since forwarded headers are spoofable and the host's proxy already rewrites it; a filter covers a host where that is not true. The IP is hashed with a site salt to make the transient key and is stored nowhere.
4. **A closed key set.** Core validates the `args` it knows and lets unknown keys through, so the gate rejects any key not in the schema with a 400. Nothing unexpected can reach the row.
5. **Core's schema validation**: uuid pattern, closed enums for the reasons and the context, bounded integers, 32-character version and locale strings restricted to `[A-Za-z0-9._-]`, details capped at 1000 characters.
6. **One `INSERT IGNORE`** on the uuid primary key. Replaying a captured submission cannot add rows.

Every answer is small and carries nothing to probe with: no row counts, no distinct timing for a duplicate versus a new row beyond the status code. Nothing logs the request. The two counts that could fingerprint a site (`enabled_user_count`, `active_plugins`) arrive exact and are stored as buckets. No IP, user agent, referer or URL is ever written.

## What is stored

`{$wpdb->prefix}openstation_feedback_deactivations`, one row per submission: the uuid, the receive time, the ticked reasons (comma-joined, in the dialog's order), the details, four short version and locale strings, three flags, the two nullable ages, the two buckets, and the context. `tools/feedback-intake/README.md` has the column table and the "what it never holds" statement that mirrors the disclosure in `readme.txt`.

## Packaging

`npm run package:feedback-intake` (`bin/package-feedback-intake.sh`) zips `tools/feedback-intake/` as `openstation-feedback-intake.zip` for a manual upload on openstation.blog. It is a separate script rather than a second output of `bin/package.sh` because that one is the wp.org release contract, run by four CI workflows that consume its single artifact, and the intake never ships there. Nothing in CI calls the new script; the zip lands at the repo root, gitignored.

## Testing instructions

The plugin needs a WordPress site to receive on. On the QA wp-env, copy it into the container and activate it:

```
docker cp tools/feedback-intake <wordpress-container>:/var/www/html/wp-content/plugins/openstation-feedback-intake
npx wp-env run cli wp plugin activate openstation-feedback-intake
```

Then, with `U=http://localhost:8890/wp-json/openstation-feedback/v1/deactivation` and a valid body (the keys in `osfi_submission_schema()`), expect:

| Request | Answer |
|---|---|
| valid submission | 201 `{"received":true}` |
| same uuid again | 200, still one row |
| unknown key, bad enum, malformed uuid | 400 |
| 3 KB body | 413 |
| `text/plain` | 415 |
| 6th from one IP within a minute | 429 with `Retry-After: 60` |
| GET anonymous | 401 |
| GET as an administrator | 200, rows plus `X-WP-Total` |

For the end-to-end path, point the OpenStation forwarder from #840 at the local intake with a mu-plugin (`http://localhost/…` inside the container, since the mapped port is host-side only) and deactivate OpenStation from the dialog:

```php
add_filter( 'openstation_deactivation_feedback_endpoint', fn () => 'http://localhost/wp-json/openstation-feedback/v1/deactivation' );
```

Tools → OpenStation feedback then shows the row with the ticked reasons and the bucketed counts, and Download CSV exports it.

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-850-2ca7497689ae94f7f8cdb10f57b36a100247e511.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


epeicher:
Closed in favor of https://github.com/epeicher/openstation-blog/pull/1

PR #570, opened 2026-08-12
Windows: Fix the accessibility semantics of controls, tabs and loading

AllTerrainDeveloper:
Three invisible accessibility bugs in the window shell.

## What was wrong

- **Control buttons had no accessible name.** The label sat on the `<os-window-button>` host, but focus lands on a `<button>` inside its shadow root whose only content is an `aria-hidden` glyph — so Minimize, Maximize, Fullscreen and Close all announced as unnamed buttons.
- **Empty tab lists.** Every iframe window exposed `role="tablist"` whether or not it had sub-pages, announcing a tab list with no tabs.
- **The loading overlay was `aria-hidden`.** A loading window read as a `role="dialog"` that was simply empty.

## What changed

- `<os-window-button>` forwards `aria-label` onto the shadow `<button>`, and observes the attribute so relabelling the host relabels the button.
- The tab strip carries `role="tablist"` + its label only while it holds tabs, `role="presentation"` while empty. `tabs.ts` re-syncs it as external sub-tabs are added and removed.
- The overlay is a `role="status"` / `aria-live="polite"` region, and the window element carries `aria-busy="true"` for the duration of the load.

No CSS touched — visuals are unchanged.

## Repro steps

Open any window in the desktop shell, then in the console:

1. **Control names** — `[ ...document.querySelectorAll( '.os-window__controls os-window-button' ) ].map( b => b.shadowRoot.querySelector( 'button' ).getAttribute( 'aria-label' ) )`
   Before: `[ null, null, null, null ]`. After: `[ "Minimize", "Maximize", "Enter fullscreen", "Close" ]`.
2. **Tab strip** — open a window with no sub-pages (Media) and one with them (Posts), then `[ ...document.querySelectorAll( '.os-window__tabs' ) ].map( n => [ n.getAttribute( 'role' ), n.getAttribute( 'aria-label' ) ] )`
   Before: both `"tablist"`. After: Media is `"presentation", null`; Posts is `"tablist", "Posts sub-pages"`.
3. **Loading** — reload a window from the ⋯ menu with `document.querySelector( '.os-window' )` under a `MutationObserver` on `aria-busy`: `true` on reload, gone when the content lands. `document.querySelector( '.os-window__loading' ).getAttribute( 'role' )` is `"status"`, and no longer `aria-hidden`.

## Tests

New `tests/vitest/window-a11y-semantics.test.ts` (6 tests) plus 4 cases on `<os-window-button>`. Full suite, lint, typecheck and build green.

[...]

PR #857, opened 2026-09-20
os-button: forward a host aria-label onto the inner button so an icon-only button has a name

### LINKED ISSUE #855: os-button: a host aria-label is not forwarded to the inner button, so an icon-only kit button has no accessible name
### What

`<os-button>` renders `<button part="button" ...><slot></slot></button>` inside its shadow root (`src/ui/components/os-button/os-button.ts`, trunk `6fd8d43a`, 1.1.10) and its `static props` are `variant`, `disabled`, `type`, `busy`, `fill-cell`. An `aria-label` set on the host element is inert: the focusable element is the inner `<button>`, and nothing carries the name across. An icon-only kit button (an arrow, a glyph) therefore announces as its slotted character or as nothing.

This is the same class of defect #789 fixed for `os-text-field` (no way to expose an accessible name without a visible label), on the other first-party control that most often ships icon-only.

### Reproduce

```html
<os-button variant="ghost" aria-label="Move up"><span aria-hidden="true">&uarr;</span></os-button>
```

Inspect the accessibility tree: the button's name is empty (the arrow is hidden), `aria-label` sits on the host and is never read.

### Ask

Either forward a host `aria-label` (and `aria-labelledby` / `aria-describedby`) onto the inner `<button>` as `os-text-field` now does for its input, or document that the accessible name must be slotted as visually hidden text. The workaround we shipped is the latter: a slotted `<span class="sr-only">Move up</span>` beside the hidden glyph.

---

Found while building icon-only reorder controls in an app's editor. Happy to take it: WordPress/openstation#857 forwards `aria-label`.

juanlentino:
Fixes #855.

## The problem

`<os-button>` renders its `<button>` inside a shadow root, and that is the element focus lands on. The host is a custom element with no role, so an `aria-label` written on it was inert: an icon-only kit button (an arrow beside an `aria-hidden` glyph) announced as its slotted character or as nothing. The same defect #570 fixed on `<os-window-button>` for the title-bar controls, and the same class #789 fixed on `<os-text-field>`.

## The change

`aria-label` is observed (`static get observedAttributes()`, the `<os-window-button>` precedent) and forwarded onto the shadow `<button>`; a relabel re-renders; a host without one leaves the inner button without one, since the renderer omits an empty attribute rather than writing `aria-label=""`. It stays out of `static props` so the base class does not install a prop accessor over `HTMLElement.prototype.ariaLabel`.

Only `aria-label` is forwarded. `aria-labelledby` and `aria-describedby` are IDREFs, and an IDREF on a shadow-tree element resolves inside that shadow root only, so copying either onto the inner button would name nothing; the docs say so, and authors who need them keep the visually hidden slotted text.

- Tests: host `aria-label` lands on the inner button; a relabel updates it; no host attribute means none inside; `variant`, `disabled` and the slot are untouched. Red against trunk.
- Help block and `docs/components-reference.md` describe the forward. The comment in `src/plugins/drafts-widget/index.ts` that asserted the old behaviour is corrected.
- `npm run build:components`, `lint`, `test:js` (6320 passed), `typecheck` all green.

Where we hit it: signal-and-noise-tools 17.4.0, the native Resume editor's reorder arrows (juanlentino/signal-and-noise-tools#1567).

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22%24schema%22%3A%22https%3A%2F%2Fplayground.wordpress.net%2Fblueprint-schema.json%22%2C%22landingPage%22%3A%22%2Fopenstation%2F%22%2C%22preferredVersions%22%3A%7B%22php%22%3A%228.3%22%2C%22wp%22%3A%22latest%22%7D%2C%22features%22%3A%7B%22networking%22%3Atrue%7D%2C%22siteOptions%22%3A%7B%22blogname%22%3A%22OpenStation%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22login%22%2C%22username%22%3A%22admin%22%2C%22password%22%3A%22password%22%7D%2C%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2FWordPress%2Fopenstation%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-857-039a261ec61b64ecfba943347c9d06771e3467e8.zip%22%7D%2C%22options%22%3A%7B%22activate%22%3Atrue%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


mmtr:
Thanks @juanlentino! This is right and matches what `<os-window-button>` already does.

Noting it also fixes an existing control: the my-wordpress dossier close button sets `aria-label="Close details"` over a slotted `✕`, so it was announcing the glyph.

One small thing I pushed: the drafts-widget comment explained when the forward arrived rather than why the name is slotted, and AGENTS.md rules out history in comments.

mmtr (approved):

